PR #36, opened 2021-06-25
Benny/single nft endpoint

bennycio:
Changes:

- **Changed** many files' formatting with go fmt ./... (I apologize this changed so many files, it happens when I save in vscode)
- **Added** endpoint for retrieving nft by id
- **Added** function for retrieving nft by id from mongo
- **Added** test function for creating an nft then retrieving that nft and finally deleting the nft by id from mongo

Files to review that were not just adjusted by go fmt:
- glry_db/t__nft_test.go
- glry_db/glry_nft.go
- glry_lib/glry_http_handlers.go


fbrcsftnr (approved):

glry_lib/glry_http_handlers.go
```diff
@@ -137,20 +136,60 @@ func HandlersInit(pRuntime *glry_core.Runtime) {
...
+
+				nftIDstr := qMap.Get("id")
+
+				if nftIDstr == "" {
+					// is this the right way to create an error for gf_core?
+					return nil, gf_core.Error__create("nft id not found in query values",
+						"http_client_req_error",
+						map[string]interface{}{
+							"uri": "/glry/v1/nfts/get",
+						}, nil, "glry_core", pRuntime.RuntimeSys)
```
kaitoo1, line 171:
```suggestion
						}, nil, "glry_lib", pRuntime.RuntimeSys)
```

kaitoo1: super small comment but I think `glry_core` should be `glry_lib` since this file lives in `glry_lib`

kaitoo1 (approved):

kaitoo1:
Nice work!! that was fast  🔥

Looks great overall - I have one comment about the NOT FOUND case, and left a smaller nit comment above.

I see that for the NOT FOUND case where the requested nft ID doesn't exist, we return `{"data":{"nfts":[]},"status":"OK"}`
Could we return an error instead of status OK and an empty array of nfts? Similar to how you return an error if no id param is provided?

That would make it easier to handle NOT FOUND cases on the client, and also makes it easier to classify as an error case for logs and other uses

glry_lib/glry_http_handlers.go
fbrcsftnr: awesome catch

fbrcsftnr:
@kaitoo1 agreed that we should return an error in this scenario

fbrcsftnr:
example error creation: https://github.com/gallery-so/go-gallery/blob/f50e1b3b846dcac6a6be74203d80c64b212ad6e1/glry_lib/glry_auth_jwt.go#L104

i believe you just need to return this error out of the top-level HTTP handler and it'll send down a 200 with an error message attached

glry_lib/glry_http_handlers.go
bennycio: Thank you! Didn't even notice!

bennycio:
Thank you guys! Will handle the not found case as soon as I can with a similar error as the id not found error.

PR #37, opened 2021-06-26
Benny/edit nft endpoint

bennycio:
Changes:

- **Added** error handle for single nft get empty result possibility
- **Added** endpoint for updating nft by id
- **Added** mongo function for updating nft by id
- **Changed** nft mongo test function to also run an update after creation

Considerations:

- In the update http handler, I chose to loop over the results of find nft and update each nft rather than grabbing the first nft from the find results (assuming there should only ever be one nft per id anyway) and just updating that one. Either way it will only update one nft and that loop should only ever run once but is it a better idea to instead grab `nfts[0]` and just update that one? We could also add a check to ensure that the length of `nfts` is always one. 
- Is `http.MethodPatch` the right method we would like to use for updates? Or is `http.MethodPost` better and the endpoint itself ending in update will be what differentiates it from e.g. a create nft request. 
- The mongo function for updating an nft by id utilizes the mongo go library instead of `gfcore` because `gfcore` did not have a function that I could find for updating. I did see a function for upserting, however, and I was wondering if it would be worth it to use `gfcore.Mongo__upsert()`, use the mongo lib like is currently implemented, wait for a possible update to `gfcore` to add an update func, or another reason why we are not using a mongo update altogether?

### DIFF
--- a/glry_db/glry_nft.go
+++ b/glry_db/glry_nft.go
@@ -155,10 +155,35 @@ func NFTgetByID(pIDstr string, pCtx context.Context, pRuntime *glry_core.Runtime
 	if err := cur.All(pCtx, &result); err != nil {
 		return nil, gfcore.Error__create("nft id not found in query values",
 			"mongodb_cursor_all",
-			map[string]interface{}{}, err, "glry_core", pRuntime.RuntimeSys)
+			map[string]interface{}{}, err, "glry_db", pRuntime.RuntimeSys)
 	}
 
 	return result, nil
+
+}
+
+//-------------------------------------------------------------
+
+// NOTE: there is no gfcore mongo func for update... using default mongo lib for now
+func NFTupdateById(pIDstr string, updatedNft *GLRYnft, pCtx context.Context, pRuntime *glry_core.Runtime) *gfcore.Gf_error {
+
+	opts := &options.FindOptions{}
+	if deadline, ok := pCtx.Deadline(); ok {
+		dur := time.Until(deadline)
+		opts.MaxTime = &dur
+	}
+
+	col := pRuntime.RuntimeSys.Mongo_db.Collection("glry_nfts")
+
+	updateResult, err := col.UpdateOne(pCtx, bson.D{{"_id", pIDstr}}, bson.D{{"$set", updatedNft}})
+
+	if err != nil || updateResult.ModifiedCount == 0 {
+		return gfcore.Error__create("unable to update nft",
+			"mongodb_update_error",
+			map[string]interface{}{}, err, "glry_db", pRuntime.RuntimeSys)
+	}
+
+	return nil
 
 }
 
--- a/glry_db/glry_nft.go
+++ b/glry_db/glry_nft.go
@@ -141,7 +141,7 @@ func NFTgetByID(pIDstr string, pCtx context.Context, pRuntime *glry_core.Runtime
 
 	col := pRuntime.RuntimeSys.Mongo_db.Collection("glry_nfts")
 
-	cur, gErr := gfcore.Mongo__find(bson.M{"_id": pIDstr},
+	cur, gErr := gfcore.Mongo__find(bson.M{"_id": pIDstr, "deleted": false},
 		opts,
 		map[string]interface{}{},
 		col,
--- a/glry_db/glry_nft.go
+++ b/glry_db/glry_nft.go
@@ -167,10 +167,11 @@ func NFTgetByID(pIDstr string, pCtx context.Context, pRuntime *glry_core.Runtime
 // NOTE: there is no gfcore mongo func for update... using default mongo lib for now
 func NFTupdateById(pIDstr string, updatedNft *GLRYnft, pCtx context.Context, pRuntime *glry_core.Runtime) *gfcore.Gf_error {
 
-	opts := &options.FindOptions{}
-	if deadline, ok := pCtx.Deadline(); ok {
-		dur := time.Until(deadline)
-		opts.MaxTime = &dur
+	//------------------
+	// VALIDATE
+	gErr := glry_core.Validate(updatedNft, pRuntime)
+	if gErr != nil {
+		return gErr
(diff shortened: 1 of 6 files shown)


fbrcsftnr:
> Either way it will only update one nft and that loop should only ever run once but is it a better idea to instead grab nfts[0] and just update that one? We could also add a check to ensure that the length of nfts is always one.

we should just grab the first one and update that; no need to loop since there won't be multiple IDs. we can also ensure that there's only one NFT by logging an error (that doesn't surface to the user) and storing it in Sentry. 

> Is http.MethodPatch the right method we would like to use for updates?

let's just keep things simple and use `MethodPost`! assuming we can differentiate them by the URL

> gfcore.Mongo__upsert()

i believe upsert will give us everything we need (update OR insert), so we can probably just continue to use gf_core. what do you think?

glry_db/glry_nft.go
```diff
@@ -155,13 +155,38 @@ func NFTgetByID(pIDstr string, pCtx context.Context, pRuntime *glry_core.Runtime
...
 
 }
 
+//-------------------------------------------------------------
+
+// NOTE: there is no gfcore mongo func for update... using default mongo lib for now
+func NFTupdateById(pIDstr string, updatedNft *GLRYnft, pCtx context.Context, pRuntime *glry_core.Runtime) *gfcore.Gf_error {
+
+	opts := &options.FindOptions{}
+	if deadline, ok := pCtx.Deadline(); ok {
```
fbrcsftnr, line 171: what does this deadline do?

glry_db/t__nft_test.go
```diff
@@ -58,6 +58,30 @@ func TestCreateAndGetNFT(pTest *testing.T) {
 		pTest.Fail()
 	}
 
+	singleNft := results[0]
```
fbrcsftnr, line 61: lovely test add

glry_lib/glry_http_handlers.go
```diff
@@ -151,7 +151,7 @@ func HandlersInit(pRuntime *glry_core.Runtime) {
 	gf_rpc_lib.Create_handler__http("/glry/v1/nfts/get",
 		func(pCtx context.Context, pResp http.ResponseWriter, pReq *http.Request) (map[string]interface{}, *gf_core.Gf_error) {
 
-			if pReq.Method == "GET" {
+			if pReq.Method == http.MethodGet {
```
fbrcsftnr, line 154: love it :)

glry_lib/glry_http_handlers.go
```diff
@@ -190,6 +198,67 @@ func HandlersInit(pRuntime *glry_core.Runtime) {
...
+		func(pCtx context.Context, pResp http.ResponseWriter, pReq *http.Request) (map[string]interface{}, *gf_core.Gf_error) {
+
+			if pReq.Method == http.MethodPatch {
+
+				//------------------
+				// INPUT
+
+				//------------------
+
+				pReq.ParseForm()
```
fbrcsftnr, line 212: parseform is able to accept any json body right?

fbrcsftnr: if so, super duper cool

glry_lib/glry_http_handlers.go
```diff
@@ -190,6 +198,67 @@ func HandlersInit(pRuntime *glry_core.Runtime) {
...
+
+				nftIDstr := pReq.FormValue("id")
+
+				// TODO any other values that could be potentially changed here
+
+				nftCollectorsNote := pReq.FormValue("collectors_note")
+
+				if nftIDstr == "" {
+					// is this the right way to create an error for gf_core?
```
fbrcsftnr, line 221: looks right to me!

glry_db/glry_nft.go
bennycio: I wrote this to ensure that should a context have a deadline, the deadline is actually reflected in the mongo query. I'm not sure if we plan on using contexts with deadlines in any part of the app, but if there was a 3 second deadline on a query, this would prevent a mongo query from stalling us.

glry_db/t__nft_test.go
bennycio: Thank you!

glry_lib/glry_http_handlers.go
bennycio: `req.ParseForm()` actually will only parse query values from the url or form values from a multipart form or xxx-url-form-encoded. I was under the assumption that the request would be a post request from a form on the web page which is why I chose to use this method over a json unmarshal. Was I correct in assuming that this endpoint will be hit by a form or will there be a post request sending a json NFT object instead? Or will both be possibilities? If both we could check the form values and if not try to parse the body as json and handle it separately.

bennycio:
> i believe upsert will give us everything we need (update OR insert), so we can probably just continue to use gf_core. what do you think?

Is there a possible case where we might only want to update an nft if it exists? If so, upserting would not allow us to catch this case. If not, I will switch it to upsert as soon as I can!

glry_lib/glry_http_handlers.go
fbrcsftnr:
```
/glry/v1/nfts/update?id=12345

JSON body: { collectors_note: 'this is cool' }
```

glry_lib/glry_http_handlers.go
```diff
@@ -190,6 +198,67 @@ func HandlersInit(pRuntime *glry_core.Runtime) {
 		},
 		pRuntime.RuntimeSys)
 
+	// SINGLE UPDATE
+	gf_rpc_lib.Create_handler__http("/glry/v1/nfts/update",
+		func(pCtx context.Context, pResp http.ResponseWriter, pReq *http.Request) (map[string]interface{}, *gf_core.Gf_error) {
+
+			if pReq.Method == http.MethodPatch {
```
fbrcsftnr, line 205: let's use MethodPut over MethodPatch

Issue #38, opened 2021-06-29
Schema Updates for v1


fbrcsftnr:
Questions for Ivan
1. What is `mapstructure` within a struct?
2. can gf_core support the standard Update query (not just Upsert)?

Meeting Notes
1. Should we save all data from opensea (including various nested data)? **No, delete `glry_opensea.go`**
2. Delete [GlryNftLegacy type](https://github.com/gallery-so/go-gallery/blob/main/glry_db/glry_nft.go#L48) after migration

**Updates to [NftStruct](https://github.com/gallery-so/go-gallery/blob/main/glry_db/glry_nft.go#L19)**
- remove CollectionNamesLst: https://github.com/gallery-so/go-gallery/blob/main/glry_db/glry_nft.go#L19
- add CreatorNameStr field. this it obtained from `creator.user.username` in opensea asset. note that the `user` field could be null so we'll want to null check 
![image](https://user-images.githubusercontent.com/84365743/123724803-f8eefc00-d85a-11eb-9be5-3a7531949698.png)
- rename `OpenseaTokenIdStr` to `OpenseaTokenId`: https://github.com/gallery-so/go-gallery/blob/main/glry_db/glry_nft.go#L35
- add comment above [`OpenseaIdStr`](https://github.com/gallery-so/go-gallery/blob/main/glry_db/glry_nft.go#L34) that links to an example opensea api result using asset contract address + token id: https://api.opensea.io/api/v1/asset/0xa7d8d9ef8d8ce8992df33d8b8cf4aebabd5bd270/26000331 (/asset/:contract_address/:token_id)
- add `ImageOriginalUrlStr` from opensea
- add `AnimationUrlStr` and `AnimationOriginalUrlStr` from opensea
- remove [`HiddenStr`](https://github.com/gallery-so/go-gallery/blob/main/glry_db/glry_nft.go#L42) from schema
- add `CollectorsNoteStr`. this is distinct from existing `DescriptionStr`, as the description is provided directly from the NFT source / creator, whereas Collector's Note will be gallery-specific
- add fields from asset_contract:`ContractNameStr`, `ContractNameStr`, `ContractDescriptionStr`, `ContractExternalLinkStr`, `ContractSchemaNameStr`, `ContractSymbolStr`, `ContractTotalSupplyInt`
![image](https://user-images.githubusercontent.com/84365743/123726266-afec7700-d85d-11eb-89f3-3344aee56a05.png)
[...]

PR #42, opened 2021-06-29
add new fields and externalize contract fields

bennycio:
Changes:

- Added fields and removed fields as mentioned in https://github.com/gallery-so/go-gallery/issues/38
- Externalized the contract struct so that each nft has a contract struct that has more details about the contract rather than putting all of those details in the nft struct and cluttering it up

Considerations:

- bson tags can have an optional `,omitempty` such as ``bson:"contract_name,omitempty"`` that will make sure that empty falsy fields are not put into mongo when a struct field is it's 0 value. Which fields do we want to have omitempty? My vote is that most fields should have omitempty and only the required ones don't

### DIFF
--- a/glry_db/glry_nft.go
+++ b/glry_db/glry_nft.go
@@ -22,24 +22,42 @@ type GLRYnft struct {
 	CreationTimeF float64   `bson:"creation_time"        json:"creation_time"`
 	DeletedBool   bool      `bson:"deleted"`
 
-	NameStr            string   `bson:"name"                 json:"name"`
-	DescriptionStr     string   `bson:"description"          json:"description"`
-	CollectionNamesLst []string `bson:"collection_names"     json:"collection_names"`
+	NameStr           string `bson:"name"                 json:"name"`
+	DescriptionStr    string `bson:"description"          json:"description"`
+	CollectorsNoteStr string `bson:"collectors_note" json:"collectors_note"`
 
-	ExternalURLstr     string `bson:"external_url"         json:"external_url"`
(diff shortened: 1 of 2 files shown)


kaitoo1 (approved):

glry_db/glry_nft.go
```diff
@@ -22,24 +22,42 @@ type GLRYnft struct {
...
-	ImageThumbnailURLstr string `bson:"image_thumbnail_url" json:"image_thumbnail_url"`
-	ImagePreviewURLstr   string `bson:"image_preview_url"   json:"image_preview_url"`
+	ImageURLstr             string `bson:"image_url"           json:"image_url"`
+	ImageThumbnailURLstr    string `bson:"image_thumbnail_url" json:"image_thumbnail_url"`
+	ImagePreviewURLstr      string `bson:"image_preview_url"   json:"image_preview_url"`
+	ImageOriginalUrlStr     string `bson:"image_original_url" json:"image_original_url"`
+	AnimationUrlStr         string `bson:"animation_url" json:"animation_url"`
+	AnimationOriginalUrlStr string `bson:"animation_original_url" json:"animation_original_url"`
+
+	AcquisitionDateStr string `bson:"aquisition_data" json:"aquisition_date"`
```
kaitoo1, line 50:
```suggestion
	AcquisitionDateStr string `bson:"aquisition_date" json:"aquisition_date"`
```

kaitoo1 (changes requested):
I think there's a small typo in AcquisitionDateStr field bson/json - otherwise the PR looks great 👍

PR #45, opened 2021-06-29
add glry_gallery.go with structs 

bennycio:
Changes:

- **Added** new gallery struct with some helper functions for getting (literally just the same functions as for collections but with some minor changes)

Considerations:

- bson tags can have an optional ,omitempty such as bson:"contract_name,omitempty" that will make sure that empty falsy fields are not put into mongo when a struct field is it's 0 value. Which fields do we want to have omitempty? My vote is that most fields should have omitempty and only the required ones don't

### DIFF
--- a/glry_db/glry_gallery.go
+++ b/glry_db/glry_gallery.go
@@ -20,8 +20,8 @@ type GLRYgallery struct {
 	CreationTimeF float64    `bson:"creation_time" json:"creation_time"`
 	DeletedBool   bool       `bson:"deleted"`
 
-	OwnerUserIDstr string   `bson:"owner_user_id" json:"owner_user_id"`
-	CollectionsLst []string `bson:"collections"          json:"collections"`
+	OwnerUserIDstr string   `bson:"owner_user_id,omitempty" json:"owner_user_id"`
+	CollectionsLst []string `bson:"collections,omitempty"          json:"collections"`
 }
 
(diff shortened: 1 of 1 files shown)


glry_db/glry_gallery.go
```diff
@@ -0,0 +1,121 @@
...
+
+	OwnerUserIDstr string   `bson:"owner_user_id" json:"owner_user_id"`
+	CollectionsLst []string `bson:"collections"          json:"collections"`
+}
+
+//-------------------------------------------------------------
+func GalleryCreate(pGallery *GLRYgallery,
+	pCtx context.Context,
+	pRuntime *glry_core.Runtime) *gf_core.Gf_error {
+
```
fbrcsftnr, line 31: is it worth running any validations when creating a gallery? does a collectionsLst have to be present? is it always assumed this will be provided?

fbrcsftnr:
> My vote is that most fields should have omitempty and only the required ones don't

tend to agree here, as long as it's still safe to query for fields that may not exist on certain objects. i believe mongo is cool with this

fbrcsftnr (approved):
excellent work!

glry_db/glry_gallery.go
bennycio: Definitely will run validations, thank you!

kaitoo1 (approved):

PR #46, opened 2021-06-29
Benny/jwt middleware

bennycio:
Changes:

- **Added** jwt middleware function that will add a value to context including whether the user is authenticated with a jwt and if so their user address (what is this address that is being stored in the jwt claims?)
- **Added** the middleware to the functions that were verifying jwt's and used context to determine the cases where the authentication information was needed
- **Changed** auth db to no longer store jwts and no longer handle anything related to jwts and http
- **Changed** the verify jwt func to also return the address from the claims it parses

Considerations:

- Middlewares currently require a parameter hell to create because `gf_core.http_handler` is private meaning that instead of `func middleware(gf_core.http_handler, pRuntime) gf_core.http_handler` we have `func middleware(midd func(pCtx context.Context, pResp http.ResponseWriter, pReq *http.Request) (map[string]interface{}, *gf_core.Gf_error), pRuntime *glry_core.Runtime) func(context.Context, http.ResponseWriter, *http.Request) (map[string]interface{}, *gf_core.Gf_error)`. Do we want to adjust the way we handle http, possibly write different http wrappers, use another library, etc.?
- Previously, the secret signing key that was being used to sign the jwts was a randomly generated string that was stored with the jwt in the database. Because we are trying to move to stateless jwt, this is harder to do. **Option one** (currently implemented) is using an env variable as a secret for all jwt's. Less secure but as long as we don't accidentally commit the key to the repo we should be find, and if we do mess up, pushing a new key to production would just force everyone to log out which is not the end of the world. **Option two** is store a random key in the redis with the blacklisted jwts seeing as we are going to have to check the redis every time someone uses a jwt anyway. More secure but almost as stateful as the previous solution; only difference being redis instead of mongo. **Option three** is anything else we can come up with. I vote for option one.
- Currently I am attaching a key value pair of string to struct to the context of the middleware which will be accessible to all further funcs using that context. For example, in the currently implemented jwt middleware it will attach a key value pair of `"auth": {Authenticated: bool, UserAddress: string}`. In this implementation the value struct has very explicit fields. We could also, however, store the value as a map and retrieve items with `pCtx.Value("auth").(map[string]interface{})["authenticated"]`. The final solution I can think of is have the key value pair be `"authenticated" => bool` and attach another key, value pair for `"user_address" => string`. This could get really messy in code, though, because a new context has to be created for every key value pair that basically copies the last context and adds a new key value pair, and if there are a lot of fields to be added in a middleware we want down the line it would require a lot more lines of code and way more memory allocations.

### DIFF
--- a/glry_db/glry_auth.go
+++ b/glry_db/glry_auth.go
@@ -52,17 +52,6 @@ type GLRYuserNonce struct {
 	AddressStr GLRYuserAddress `bson:"address" mapstructure:"address"`
 }
 
-// USER_JWT_KEY - is unique per user, and stored in the DB for now.
-type GLRYuserJWTkey struct {
-	VersionInt    int64            `bson:"version"       mapstructure:"version"`
-	ID            GLRYuserJWTkeyID `bson:"_id"           mapstructure:"_id"`
-	CreationTimeF float64          `bson:"creation_time" mapstructure:"creation_time"`
-	DeletedBool   bool             `bson:"deleted"       mapstructure:"deleted"`
-
-	ValueStr   string          `bson:"value"   mapstructure:"value"`
-	AddressStr GLRYuserAddress `bson:"address" mapstructure:"address"`
-}
-
 // USER_LOGIN_ATTEMPT
 type GLRYuserLoginAttempt struct {
 	VersionInt    int64              `bson:"version"`
@@ -85,70 +74,6 @@ type GLRYuserLoginAttempt struct {
 type GLRYuserUpdate struct {
 	CreationTimeF float64 `bson:"creation_time"`
 	NameNewStr    string  `bson:"name_new"`
-}
-
-//-------------------------------------------------------------
-// JWT
-//-------------------------------------------------------------
-// GET
-func AuthUserJWTkeyGet(pUserAddressStr GLRYuserAddress,
-	pCtx context.Context,
-	pRuntime *glry_core.Runtime) (*GLRYuserJWTkey, *gf_core.Gf_error) {
-
-	record, gErr := gf_core.MongoFindLatest(bson.M{
-		"address": pUserAddressStr,
-		"deleted": false,
-	},
-		"creation_time", // p_time_field_name_str
-		map[string]interface{}{
-			"address":            pUserAddressStr,
-			"caller_err_msg_str": "failed to get JWT key from DB",
-		},
-		pRuntime.DB.MongoDB.Collection("glry_users_jwt_keys"),
-		pCtx,
-		pRuntime.RuntimeSys)
-	if gErr != nil {
-		return nil, gErr
-	}
-
-	var JWTkey GLRYuserJWTkey
-	err := mapstructure.Decode(record, &JWTkey)
-	if err != nil {
-		gErr := gf_core.Error__create("failed to load DB result of GLRYuserJWTkey into its struct",
-			"mapstruct__decode",
-			map[string]interface{}{
-				"address": pUserAddressStr,
-			},
-			err, "glry_db", pRuntime.RuntimeSys)
-		return nil, gErr
-	}
-
-	// spew.Dump(JWTkey)
-
-	return &JWTkey, nil
-}
-
-//-------------------------------------------------------------
-// CREATE
-
-func AuthUserJWTkeyCreate(pJWTkey *GLRYuserJWTkey,
-	pCtx context.Context,
-	pRuntime *glry_core.Runtime) *gf_core.Gf_error {
-
-	collNameStr := "glry_users_jwt_keys"
-	gErr := gf_core.Mongo__insert(pJWTkey,
-		collNameStr,
-		map[string]interface{}{
-			"address":        pJWTkey.AddressStr,
-			"caller_err_msg": "failed to insert a new GLRYuserJWTkey into the DB",
-		},
-		pCtx,
-		pRuntime.RuntimeSys)
-	if gErr != nil {
-		return gErr
-	}
-
-	return nil
 }
 
 //-------------------------------------------------------------
--- a/glry_lib/glry_auth_http_handlers.go
+++ b/glry_lib/glry_auth_http_handlers.go
@@ -3,13 +3,14 @@ package glry_lib
 import (
 	// "fmt"
 	"context"
+	"net/http"
+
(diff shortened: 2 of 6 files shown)


fbrcsftnr:
> what is this address that is being stored in the jwt claims?

this is the user's wallet address. may need to be updated when a user can have multiple addressses

> Do we want to adjust the way we handle http, possibly write different http wrappers, use another library, etc.?

agreed on parameter hell. i would defer to you in terms of migrating to a different wrapper or library: whatever would make you most comfortable, and coding most enjoyable!

> Secret signing key

i also vote for option one, and this is common practice in the industry. in the future we can upgrade this and have regular key rotation. we'll also move over to using cookies over auth headers

> string struct vs. context

thank you so much for doing this thorough analysis. if there will be a lot of overhead using context, let's stick to your solution

glry_db/glry_auth.go
```diff
@@ -87,70 +76,6 @@ type GLRYuserUpdate struct {
...
-		},
-		pCtx,
-		pRuntime.RuntimeSys)
-	if gErr != nil {
-		return gErr
-	}
-
-	return nil
-}
-
```
fbrcsftnr, line 153: glorious deletions

glry_lib/glry_auth_http_handlers.go
```diff
@@ -131,8 +132,14 @@ func AuthHandlersInit(pRuntime *glry_core.Runtime) {
 	// AUTHENTICATED
 
 	gf_rpc_lib.Create_handler__http("/glry/v1/users/update",
-		func(pCtx context.Context, pResp http.ResponseWriter, pReq *http.Request) (map[string]interface{}, *gf_core.Gf_error) {
+		precheckJwt(func(pCtx context.Context, pResp http.ResponseWriter, pReq *http.Request) (map[string]interface{}, *gf_core.Gf_error) {
```
fbrcsftnr, line 135: depends on whether we stay on this long-term, but wonder if it's worth defining the handlers elsewhere, so the routes look less verbose:

```go
gf_rpc_lib.Create_handler__http("/glry/v1/users/update", precheckJwt(updateUser), pRuntime)
gf_rpc_lib.Create_handler__http("/glry/v1/users/get", precheckJwt(getUser), pRuntime)

...

updateUser, getUser defined elsewhere
```

glry_lib/glry_auth_jwt.go
```diff
@@ -2,97 +2,42 @@ package glry_lib
...
+// USER_JWT_KEY - is unique per user, and stored in the DB for now.
+// type GLRYuserJWTkey struct {
+// 	VersionInt    int64            `bson:"version"       mapstructure:"version"`
+// 	ID            GLRYuserJWTkeyID `bson:"_id"           mapstructure:"_id"`
+// 	CreationTimeF float64          `bson:"creation_time" mapstructure:"creation_time"`
+// 	BlackListed   bool             `bson:"blacklisted"       mapstructure:"blacklisted"`
+
+// 	ValueStr   string          `bson:"value"   mapstructure:"value"`
+// 	AddressStr GLRYuserAddress `bson:"address" mapstructure:"address"`
+// }
```
fbrcsftnr, line 25: delete?

glry_lib/glry_auth_jwt.go
```diff
@@ -124,40 +75,19 @@ func AuthJWTgeneratePipeline(pAddressStr glry_db.GLRYuserAddress,
 	pCtx context.Context,
 	pRuntime *glry_core.Runtime) (string, *gf_core.Gf_error) {
 
-	JWTkeyStr := AuthGenerateRandom()
-
+	// previously we would generate a random string and use that as jwt secret and store
+	// the string in the db with the jwt for verifcation. with stateless auth, we might
+	// use an environment variable like so as the secret. worth considering other options
+	// to increase security
```
fbrcsftnr, line 81: agreed with approach, let's not commit this comment :)

glry_lib/glry_http_middlewares.go
```diff
@@ -0,0 +1,58 @@
...
+
+type authContextValue struct {
+	AuthenticatedBool bool
+	UserAddressStr    string
+}
+
+// jwt middleware
+// parameter hell because gf_core http_handler is private :(
+// both funcs (parameter and return funcs) are of type gf_core.http_handler implicitly
+func precheckJwt(midd func(pCtx context.Context, pResp http.ResponseWriter,
```
fbrcsftnr, line 24: oh man thanks for dealing with this regardless! would putting each parameter in a new line at least make things look more palatable?

awesome middleware nonetheless

glry_lib/glry_http_middlewares.go
```diff
@@ -0,0 +1,58 @@
...
+	pReq *http.Request) (map[string]interface{}, *gf_core.Gf_error), pRuntime *glry_core.Runtime) func(context.Context, http.ResponseWriter,
+	*http.Request) (map[string]interface{}, *gf_core.Gf_error) {
+	return func(pCtx context.Context, pResp http.ResponseWriter,
+		pReq *http.Request) (map[string]interface{}, *gf_core.Gf_error) {
+
+		authHeaders := strings.Split(pReq.Header.Get("Authorization"), " ")
+		if len(authHeaders) > 0 {
+			// get string after "Bearer"
+			jwt := authHeaders[1]
```
fbrcsftnr, line 33: instead of splitting, prefer verifying the token is preceded by `"Bearer "` explicitly. there are ways to do this with regex if you google around

just so that people can't send `"garbage ${jwt}"`

glry_lib/glry_auth_http_handlers.go
bennycio: Awesome, that looks so much better! I will definitely be changing it to this once I decide on a way to handle http.

glry_lib/glry_auth_jwt.go
bennycio: will delete 👍🏻

glry_lib/glry_http_middlewares.go
bennycio: True, I will put every param on a new line and that will make it look way better. Thank you!

glry_lib/glry_http_middlewares.go
bennycio: Will do!

PR #49, opened 2021-07-02
mongo wrappers

bennycio:
Changes:

- **Added** `glry_core/glry_persist.go` with helper funcs that will reflectively insert fields
- **Added** missing quote on struct tag
- **Accidentally Added** incoming changes to main because I suck at git

Considerations:

- Reflection is dangerous but in this case prevents us from having to write separate wrappers for each type being inserted into the database. A possible option that would allow us to avoid reflection would be to write an interface that has funcs such as `CreateId() GLRYdbId`, `SetVersion(ver int64)`, `SetLastUpdated(time float64)`, and `SetCreationTime(time float64)` and ensure that each struct we are using as a mongo document implements that interface. This will increase type safety but require us to implement each of those funcs for each of the structs.
- Do we need an individual id type for each structs id field or can we use `GLRYdbId` like I have implemented? I cannot think of a way to generically fill id field if the type of the id differs from struct to struct
- Is there a better package to put `glry_persist.go`?
- Currently the `GLRYpersister` interface is general enough where if we decided we hated mongo and wanted to move somewhere else to a different database provider we could. Because of this, however, each implementation then requires us to check make type checks to ensure fields are of types that correspond to mongo options, etc.. Should `GLRYpersister` be this generic or should we assume that mongo is going to be our provider and switch fields with `interface{}` to `*options.SomeMongoOptions`


glry_core/glry_persist.go
```diff
@@ -0,0 +1,148 @@
...
+	Version    int64
+	Collection *mongo.Collection
+}
+
+func NewMongoPersister(version int64, collName string, runtime Runtime) GLRYpersister {
+	collection := runtime.DB.MongoDB.Collection(collName)
+	return GLRYmongoPersister{Version: version, Collection: collection}
+}
+
+// i must be a pointer to a struct
```
fbrcsftnr, line 36: what is "i" referring to?

fbrcsftnr:
first off, this is tremendous work. wow!!!

so my understanding was that we would write wrappers around the ones we're currently using with `gf_core`, such as `Mongo__insert`, `MongoFind`, etc. these wrappers would receive the `bson.M` queries and simply append things we know we'd like to append, like creation time, filtering out `"deleted": true`, and so on. did not expect such low level wrappers!

happy to discuss in person, but again, whatever makes sense for your setup and would be most helpful to you. if we do take things into our own hands, i'd be curious if we can store and edit deep objects, such as nested objects or slices. 

> This will increase type safety but require us to implement each of those funcs for each of the structs.

yeah that seems like overkill haha

> Do we need an individual id type for each structs id field or can we use GLRYdbId like I have implemented? I cannot think of a way to generically fill id field if the type of the id differs from struct to struct

let's use a general ID. in fact, i'm down to moving away from having it be dependent on any parameters (such as ...Fields) and simply generate an md5 based on timestamp. collision should be minimal if we use nanosecond-level precision

> Is there a better package to put glry_persist.go?

makes sense where it is!

> Should GLRYpersister be this generic or should we assume that mongo is going to be our provider and switch fields with interface{} to *options.SomeMongoOptions

i vote we assume we will be sticking with mongo in the medium to long run, no need to overengineer at this stage

glry_core/glry_persist.go
bennycio: Meant to be "insert", sorry about that.

bennycio:
> so my understanding was that we would write wrappers around the ones we're currently using with gf_core, such as Mongo__insert, MongoFind, etc. these wrappers would receive the bson.M queries and simply append things we know we'd like to append, like creation time, filtering out "deleted": true, and so on. did not expect such low level wrappers!

Newest commit uses `gf_core` instead of mongo lib. My only hesitation here is that gf_core does not allow us to specify any mongo specific options for insertion or updating and inconsistently does allow for mongo options on the find func. It is also inconsistent about using the `mongo.Collection` type versus passing in a string and having the collection be established under the hood. If you look at `gf_core.Mongo__insert`'s parameters versus `gf_core.Mongo__upsert` you can see that one takes a string for collection and one takes the type `*mongo.Collection`. Either way, looking at gf_cores mongo funcs reveals that they are mostly to gracefully handle mongo errors with `gf_core.Gf_error` so we could go either way and it should be fine.

Also forgot to include the `deleted: true` in my initial commit, added it in most recent commit.

> let's use a general ID. in fact, i'm down to moving away from having it be dependent on any parameters (such as ...Fields) and simply generate an md5 based on timestamp. collision should be minimal if we use nanosecond-level precision

Sounds good, reflected in newest commit.

Finally, the reason why the insert and update seem so low level and use reflection is because upon insertion and update, we are using structs with bson tags instead of a `bson.M` or `bson.D`. It's easy to append values to a `bson.M` or `bson.D` but unfortunately requires reflection for hot-insertion of fields on a struct. For `Find()`, however, we are using a bson.M as a filter which can be appended to and therefore does not require reflection, yay!

fbrcsftnr:
@bennycio thank you for the thorough explanations! i remember our discussion now, makes sense to have lower level control over the mongo library especially if we can only upsert from gf_core

bennycio:
For some reason in most recent commit, the update method requires that the bson id field has `,omitempty` or it complains that we can't update an immutable field. The current update being used for the db nft does not have this problem but I cannot see a difference between that update and this more generic update other than the reflection. If anyone can figure out why this might be happening I would be very happy!

bennycio:
PR looking pretty good at this point with aggregation able to solve the problem of filling nested documents. Something to consider now is that most our find operations are being decoded and unmarshalled directly onto a struct. This works because in the db, each document holds its sub documents as an array of strings (array of ids) which corresponds to the go structs which also hold an array of strings for ids. In aggregation, however, the returned set of documents replaces that array of strings to an array of sub documents which does not correspond to the go struct meaning we cannot directly unmarshall this onto any struct. We have a few options plus anything anyone else can think of:

1. We replace the struct field of array of ids to an array of structs and ensure that on insert, rather than adding each struct we loop through the child structs and only add the id. This would make the insertion process a bit more complicated and more manual. 
2. We create a copy struct where the only difference is that instead of array of ids there are arrays of structs. This would cause some code duplication but to make it clear we could name them something like `GLRYnftInsert` and `GLRYnft`/`GLRYnftResponse` seeing as the one with the array of strings will most likely only be used during the insertion process.
3. We do not even unmarshall onto a struct and instead turn it straight into json to be returned by the api. This would our options with regards to handling the data on the backend before it reaches the front end but if all we are doing is finding and returning data without touching it in between, this could work.
4. Two struct fields for the array of string ids and the array of structs. This could cause some inconsistencies where a struct might have both arrays filled (duplicate information) or just one filled.

Let me know what you think!

fbrcsftnr:
@bennycio thank you for the analysis. i think we should definitely go with option 2, which is to create a separate struct that represents the response from the aggregation query, regardless of code duplication. aggregations are super flexible and we can return all kinds of custom data that may look very different than simply an embedded object (such as adding a new "count" field that represents how many documents were returned, etc.).

we should decouple the structs and have two separate representations, i.e., the shape of the _stored_ data vs. the shape of the _exposed/returned_ data

PR #50, opened 2021-07-05
http-framework

bennycio:
We have been talking about looking at different ways of handling http that give us more control over things like middlewares and return values. In the most recent commit I have experimented with gin, a very popular go web framework that is very clear and is especially easy to use with regards to middlewares and keeping our routes organized and clear in code. I wrote out some examples using our current code but using the gin web framework. You can check out gin at https://github.com/gin-gonic/gin.

It would take some adjustments to switch over to gin, nothing major, but I believe it will save us headaches and roadblocks in the future. If we believe it will be too much to change now or anyone has any other suggestions I would love to hear! The standard library also has a lot of the versatility as gin just with less of the built in optimization, API features, logging, error handling, etc. so we could also take a look at how that would look.

If we believe gin is the right way to go, I will get going on converting our current handlers to gin handlers.

### DIFF
--- a/glry_lib/glry_auth_http_handlers.go
+++ b/glry_lib/glry_auth_http_handlers.go
@@ -47,7 +47,7 @@ func AuthHandlersInit(pRuntime *glry_core.Runtime, parent *gin.RouterGroup) {
 
 	usersGroup.POST("/login", func(c *gin.Context) {
 		input := &GLRYauthUserLoginInput{}
-		if err := c.BindJSON(input); err != nil {
+		if err := c.ShouldBindJSON(input); err != nil {
 			c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
 			return
 		}
@@ -85,14 +85,14 @@ func AuthHandlersInit(pRuntime *glry_core.Runtime, parent *gin.RouterGroup) {
 
 	usersGroup.POST("/update", func(c *gin.Context) {
 
-		if auth, ok := getAuthFromCtx(c); !ok || !auth.AuthenticatedBool {
+		if auth := c.GetBool("authenticated"); !auth {
 			c.JSON(http.StatusUnauthorized, gin.H{"error": "authorization required"})
 			return
 		}
 
 		up := &GLRYauthUserUpdateInput{}
 
-		if err := c.BindJSON(up); err != nil {
+		if err := c.ShouldBindJSON(up); err != nil {
 			c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
 			return
 		}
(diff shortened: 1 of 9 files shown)


kaitoo1:
@bennycio  thanks for looking into this, and the detailed write up!

I'm not sure what the discussion has been so far about exploring different ways to handle http requests, but I'm totally down to switch to a more widely used framework. 100% agree about preventing roadblocks and headaches in the future, and should help with future ease of development and maintainability. 

Is there any benefit to using the standard library over Gin? 

In terms of timing for the switch - I think a good way to think about it is what work there is left for v1. We had [that list of backend functionality](https://github.com/gallery-so/go-gallery/issues/33) that we want to build in order to release v1, so the main question IMO is: would it be easier/faster to do the switch now (and migrate existing code) and build the remaining endpoints with the new framework, or build the remaining work with the existing framework, and migrate everything after?
Since it looks like you already managed to get started with migrating a bit of it, and you seem comfortable with Gin, I'm leaning towards the former and switching sooner rather than later.
Gin looks like it has great documentation as well, which should make it easier to build the remaining features as well.

What do you think @fbrcsftnr ?

bennycio:
Thanks for responding so quickly @kaitoo1! The only advantage I see to the standard library is that it is extremely stable and most likely will never change where as gin could potentially change. Gin is also a major web framework though and I doubt will change in a way that will break should we update to a new version. In practically every other way, though, Gin improves upon what the standard library offers. I believe it is a better choice.

I also believe that transitioning our current handlers to gin handlers would be a positive move and not take very much time. It took me very little time to make the examples because most of the inner logic stays the same but things like handling request query params, different request formats, responses, and middlewares are different.

bennycio:
Just spent a very short amount of time rewriting the handlers in the gin way so, after some testing of course, the work is already practically done. It was much easier to do now while we did not have very many handlers rather than later when we have handlers galore.

glry_lib/glry_auth_http_handlers.go
```diff
@@ -2,20 +2,20 @@ package glry_lib
...
-	log "github.com/sirupsen/logrus"
 	// "github.com/davecgh/go-spew/spew"
 )
 
 //-------------------------------------------------------------
-func AuthHandlersInit(pRuntime *glry_core.Runtime) {
+func AuthHandlersInit(pRuntime *glry_core.Runtime, parent *gin.RouterGroup) {
+
+	usersGroup := parent.Group("/users")
+	usersGroup.Use(jwtMiddleware(pRuntime))
```
fbrcsftnr, line 18: love this

fbrcsftnr (approved):
in total agreement that going with a stable, widely accepted library will improve speed of development. thank you for implementing this proof of concept, i'm in favor of gin!

PR #51, opened 2021-07-06
cleanup external services to reflect changes to structs

bennycio:
Changes:

- **Added** `openSeaApiResponse` struct to make json unmarshalling easy into our newly updated nft struct
- **Changed** fetch data from open sea func to return `GLRYnft`'s after unmarshalling them from the request response
- **Changed** persist func to no longer turn the old open sea struct into nft structs manually, because the fetch data func already returns nfts, yay!

Considerations:

- Currently the persisting func does not consider the fact that the user might already have persisted nfts from open sea and will create bulk without checking if the nfts already exist. We will have to handle this in the near future


fbrcsftnr (approved):
nice work!

PR #54, opened 2021-07-14
architecture adjustments, thanks vscode

bennycio:
Changes:

- **Changed** almost everything but did not actually change much. Just moved things around and renamed them with the handy dandy refactoring capabilities of vscode

Considerations:

- A lot of things are renamed and moved around, if anyone has any opinions on where things should be instead of where they are or what they should be named, I am all ears!


persist/collection.go
```diff
@@ -0,0 +1,112 @@
...
+	"context"
+	"time"
+
+	"github.com/mikeydub/go-gallery/runtime"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+const collectionColName = "glry_collections"
```
fbrcsftnr, line 13: should we update the collection names as well?

persist/mongo.go
```diff
@@ -1,4 +1,4 @@
-package glry_core
+package persist
```
fbrcsftnr, line 1: noticed that files within glry_db/ and glry_core/ are given `package persist`. is this intentional? should we rename the directory to be `persist/`?

fbrcsftnr: ^ nevermind!! i see that it's been moved haha

fbrcsftnr (approved):

persist/collection.go
bennycio: Very true, I will adjust those as well! Thank you for pointing this out to me

PR #55, opened 2021-07-15
add get by username and get by wallet addr

bennycio:
Changes:

- **Added** parts of get func to handle getting by address or username
- **Changed** get user input to reflect possible query values
- **Added** db func for getting by username

Considerations:

- Currently there is no validation to ensure that the query only contained one of: username, wallet, user_id -- and instead will only handle once based on their priority. This priority is user_id, username, address (in order of least to most complex/time consuming query). That means the query could be something like /glry/v1/users/get?username=:username&addr=:walletAddress&user_id=:userId and it would still work. Do we want it to function this way, or do we want to ensure that only one query param is given?

### DIFF
--- a/persist/user.go
+++ b/persist/user.go
@@ -104,8 +104,11 @@ func UserGetById(userId DbId,
 		return nil, err
 	}
 
-	if len(result) == 0 || len(result) > 1 {
-		return nil, fmt.Errorf("invalid amount of returned users: %d", len(result))
+	if len(result) == 0 {
+		return nil, fmt.Errorf("no users found")
+	}
+	if len(result) > 1 {
+		return nil, fmt.Errorf("more than one user found when expecting a single result")
 	}
 
 	return result[0], nil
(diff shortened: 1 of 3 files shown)


bennycio:
Somehow I didn't even mention that this is for getting **users** if that was not clear.

persist/user.go
```diff
@@ -109,10 +117,44 @@ func UserGetByAddress(pAddress string,
...
+	mp := NewMongoStorage(0, usersCollName, pRuntime)
+
+	result := []*User{}
+	err := mp.Find(pCtx, bson.M{"addresses": bson.M{"$in": []string{pAddress}}}, result, opts)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if len(result) == 0 || len(result) > 1 {
```
fbrcsftnr, line 135: Can we break this into two separate errors? i.e. "no users found with wallet address" and "greater than 1 user found with wallet address"

server/user.go
```diff
@@ -245,16 +225,38 @@ func userGetDb(pInput *userGetInput,
...
+			return nil, err
+		}
+		break
+	case pInput.Address != "":
+		user, err = persist.UserGetByAddress(pInput.Address, pCtx, pRuntime)
+		if err != nil {
+			return nil, err
+		}
+		break
+	}
```
fbrcsftnr, line 248: Beautiful

fbrcsftnr:
Awesome cleanup!

Re: priority/validation of query params, it's fine if the client sends multiple in my opinion, since it's sort of on them if they decide to include a user ID + wallet pair that's invalid

fbrcsftnr (approved):

persist/user.go
bennycio: sounds good, reflected on most recent commit!

PR #59, opened 2021-07-16
Benny/opensea sync

bennycio:
Changes:

- **Added** upsert funcs for ensuring that we do not create duplicates and only add or update what's necessary when grabbing info from open sea
- **Added** func that will use opensea_id as a baseline to find out which ones have already been added

Considerations:

- Is opensea_id always going to be the most definitive field to see which nfts need to be added versus updated?


fbrcsftnr:
incredible work on this PR!

> Is opensea_id always going to be the most definitive field to see which nfts need to be added versus updated?

i believe so, given that this is a sync pipeline that is specialized for opensea. we can explore longer term solutions after v1

persist/nft.go
```diff
@@ -150,3 +152,83 @@ func NftUpdateById(pIDstr DbId, updatedNft *Nft, pCtx context.Context, pRuntime
...
+
+		now := float64(time.Now().UnixNano()) / 1000000000.0
+
+		// TODO last updated
+
+		upsertModels[i] = mongo.UpdateOneModel{
+			Upsert: &weWantToUpsertHere,
+			Filter: bson.M{"owner_address": walletAddress, "opensea_id": v.OpenSeaIDstr},
+			Update: bson.M{
+				"$setOnInsert": bson.M{"_id": generateId(now), "created_at": now},
```
fbrcsftnr, line 182: will this generate new IDs and creation timestamps for updated NFTs that already exist? or only for new models that don't exist in our DB yet?

persist/nft.go
```diff
@@ -150,3 +152,83 @@ func NftUpdateById(pIDstr DbId, updatedNft *Nft, pCtx context.Context, pRuntime
 	return mp.Update(pCtx, bson.M{"_id": pIDstr}, updatedNft)
 
 }
+
+//-------------------------------------------------------------
+
+func NftBulkUpsertOrRemove(walletAddress string, nfts []*Nft, pCtx context.Context, pRuntime *runtime.Runtime) error {
```
fbrcsftnr, line 158: this function would be really awesome to test. should be possible with some solid mock data, with following cases:
1. no DB NFTs, some opensea NFTs (should append all)
2. some DB NFTs, no opensea NFTs (should delete all)
3. a mix of scenarios

i can help work on this at some point

fbrcsftnr (approved):

persist/nft.go
```diff
@@ -150,3 +152,83 @@ func NftUpdateById(pIDstr DbId, updatedNft *Nft, pCtx context.Context, pRuntime
...
+		opts.MaxTime = &dur
+	}
+
+	allNfts := []*Nft{}
+	if err := mp.Find(pCtx, bson.M{"owner_address": walletAddress}, allNfts, opts); err != nil {
+		return err
+	}
+
+	if len(allNfts) > len(nfts) {
+		currOpenseaIds := map[string]bool{}
```
kaitoo1, line 206: Nit, non-blocking: consistent capitalization of OpenSea
```suggestion
		currOpenSeaIds := map[string]bool{}
```

persist/nft.go
kaitoo1, line 158: The function NftBulkUpsertOrRemove is quite big - would it be possible to abstract out some of the logic into separate functions?

That'll help simplify the scope of each function and improve readability and make them easier to modify in the future.

kaitoo1: Some candidates are the logic contained in the for loop on line 168 and the if statement on line 205

persist/nft.go
bennycio: this will generate id's and creation times for the nft's that are being upserted and therefore do not exist in the current db

persist/nft.go
bennycio: great! i think we can make a separate PR for the tests

bennycio: for sure! I will see where we can extract some logic, thank you!

PR #61, opened 2021-07-16
use authentication for create collection

bennycio:
Changes:

- **Added** separate jwt middleware for requiring jwt on endpoints
- **Changed** create collection endpoint to use user_id with jwt and require it
- **Changed** endpoints middlewares to use the jwt middlewares that best reflect their requirements

Considerations:

- In [The Api Spec](https://github.com/gallery-so/go-gallery/issues/33), the request for creating a collection just has a list of nfts. Currently in code, it also has a name and description. Are we going to split those up into separate endpoints, create and update, or should we do them all at once like it is currently? From a UX perspective, will the user be choosing nfts for collection first and then naming and describing the collection? If so, saving the nfts in the create and updating later would work so long as we can identify the collection when we are updating it. Since the collection has no information other than nfts and owner id, we should be able to find it by using one of the nfts along with the user_id such as `find(bson.M{"_id": userId, "nfts": bson.M{"$in": []DbId{nftIds...}}})`.


fbrcsftnr:
> From a UX perspective, will the user be choosing nfts for collection first and then naming and describing the collection?

yes. the steps should look something like this:
1. user chooses NFTs for collection and sends array of IDs to backend [POST /collections/create]
2. backend successfully stores collection and returns 200, as well as the ID of the collection
3. client uses ID of collection to make subsequent edit request to edit collection name and description [PUT /collections/update?id=:id]

for part 3, since the endpoint will need to be authenticated, we can use the user ID from the JWT as well as the ID provided in query params `{userId: id, _id: queryId}` to execute the DB find

server/collection.go
```diff
@@ -28,7 +31,7 @@ func getAllCollectionsForUser(pRuntime *runtime.Runtime) gin.HandlerFunc {
 		//------------------
 		// INPUT
 
-		userIDstr := c.Query("userid")
+		userIDstr := c.Query("user_id")
```
fbrcsftnr, line 34: we should probably make `"user_id"` a const

server/middleware.go
```diff
@@ -21,14 +26,49 @@ func jwtMiddleware(runtime *runtime.Runtime) gin.HandlerFunc {
...
+			return
+		}
+		c.Next()
+	}
+}
+
+func jwtOptional(runtime *runtime.Runtime) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		header := c.GetHeader("Authorization")
+		if header != "" {
```
fbrcsftnr, line 52: if the jwt is optional, i wonder if we should even bother verifying the header. maybe we should fail silently instead.

example:
1. user requests a public page (not theirs) at gallery.so/cool_user
2. they haven't logged in in a while, so their JWT has expired
3. the request to fetch public gallery data for cool_user fails, and user is hit with error page

in theory, the user should have been able to see cool_user's profile, despite their auth status

fbrcsftnr: it feels unusual to have middleware for something that is optional. if we deprecate `jwtOptional`, handlers should be cleaned up to only be used for the required case, which makes sense:

```js
	nftsGroup.GET("/get", getNftById(pRuntime))
	nftsGroup.GET("/user_get", getNftsForUser(pRuntime))
	nftsGroup.POST("/update", jwtRequired(pRuntime), updateNftById(pRuntime))
```

fbrcsftnr (approved):

server/collection.go
bennycio: with the new struct validators, they are struct tags instead of string literals so this should not longer be an issue 👍🏻

bennycio:
> 3. client uses ID of collection to make subsequent edit request to edit collection name and description [PUT /collections/update?id=:id]

Sounds great! I am wondering what will happen if a user leaves before step three? Will the ID be saved somewhere for the return or will the front end hand a different query to the backend to now find the recently created collection?

server/middleware.go
bennycio: Currently the JWT optional middleware is a middleware for the endpoints that react differently to authenticated vs non-authenticated users. It will insert the auth status and user ID into the context so that future handlers can access them without blocking access to any future handlers.

We could have only one jwt middleware like we used to that just inserts the auth status into the context but this would require each handler to handle it's own authentication protection (e.g. the update collection handler will first have to grab the auth status from the context and check if a user is authed before continuing). I believe this is a bit out of the scope of most handlers. The handlers that react differently to different auth statuses, however, I believe should not have to verify jwt's themselves and should just have that status handed to them from the context. Let me know if this is incorrect or if there is a better solution you have in mind.

> in theory, the user should have been able to see cool_user's profile, despite their auth status

Currently, this middleware will not stop the user from reaching the next handler even if they are not authed/jwt invalid. Therefore, even a user with an expired JWT will continue through the `jwtOptional` middleware. Is this alright?

> if the jwt is optional, i wonder if we should even bother verifying the header. maybe we should fail silently instead.

Will do!

fbrcsftnr: @bennycio this is perfect, thank you for laying it out so clearly!

fbrcsftnr:
> I am wondering what will happen if a user leaves before step three

they would move onto the step that lists all their collections via GET, at which point they would be able to use the collection ID to make further edits!

server/collection.go
```diff
@@ -10,10 +10,16 @@ import (
 	"github.com/mikeydub/go-gallery/runtime"
 )
 
+type collectionGetByIdInput struct {
+	Id persist.DbId `json:"id" binding:"required"       `
```
fbrcsftnr, line 14: why are there so many spaces before the closing `? haha

server/collection.go
```diff
@@ -10,10 +10,16 @@ import (
...
+}
+
+type collectionGetByUserIdInput struct {
+	UserId persist.DbId `json:"user_id" binding:"required"       `
```
fbrcsftnr, line 18: same here

server/collection.go
bennycio: fmt be like 😵‍💫

PR #63, opened 2021-07-16
validators for binding query or json

bennycio:
Changes:

- **Added** some validator funcs for validating things like eth address and signatures
- **Added** some input and output structs for easy binding to and from json or query

### DIFF
--- a/persist/collection.go
+++ b/persist/collection.go
@@ -18,7 +18,7 @@ const (
 //-------------------------------------------------------------
 type CollectionDb struct {
 	VersionInt    int64   `bson:"version,omitempty"       json:"version"` // schema version for this model
-	IDstr         DbId    `bson:"_id,omitempty"           json:"id"`
+	IDstr         DbId    `bson:"_id,omitempty"           json:"id" binding:"required"`
 	CreationTimeF float64 `bson:"creation_time,omitempty" json:"creation_time"`
 	DeletedBool   bool    `bson:"deleted,omitempty"`
 
@@ -33,7 +33,7 @@ type CollectionDb struct {
 
 type Collection struct {
 	VersionInt    int64   `bson:"version"       json:"version"` // schema version for this model
-	IDstr         DbId    `bson:"_id"           json:"id"`
+	IDstr         DbId    `bson:"_id"           json:"id" binding:"required"`
 	CreationTimeF float64 `bson:"creation_time" json:"creation_time"`
 	DeletedBool   bool    `bson:"deleted"`
 
--- a/server/nft.go
+++ b/server/nft.go
@@ -26,13 +26,13 @@ func getNftsById(pRuntime *runtime.Runtime) gin.HandlerFunc {
 		input := &getNftsByIdInput{}
 
 		if err := c.ShouldBindQuery(input); err != nil {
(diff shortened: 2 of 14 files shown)


persist/collection.go
```diff
@@ -18,7 +18,7 @@ const (
 //-------------------------------------------------------------
 type CollectionDb struct {
 	VersionInt    int64   `bson:"version,omitempty"       json:"version"` // schema version for this model
-	IDstr         DbId    `bson:"_id,omitempty"           json:"id"`
+	IDstr         DbId    `bson:"_id,omitempty"           json:"id" binding:"required"`
```
fbrcsftnr, line 21: haven't seen `binding` before. nice!

fbrcsftnr (approved):

server/nft.go
```diff
@@ -9,23 +9,38 @@ import (
...
-			c.JSON(http.StatusBadRequest, gin.H{"error": "nft id not found in query values"})
+		input := &getNftsByIdInput{}
+
+		if err := c.ShouldBindQuery(input); err != nil {
+			c.JSON(http.StatusOK, gin.H{"error": "nft id not found in query values"})
 			return
 		}
-		nfts, err := persist.NeftGetById(persist.DbId(nftIDstr), c, pRuntime)
+
+		nfts, err := persist.NeftGetById(input.NftId, c, pRuntime)
```
kaitoo1, line 33: Looks like this was already here but typo? "Neft" -> "Nft"

server/nft.go
```diff
@@ -9,23 +9,38 @@ import (
...
-			c.JSON(http.StatusNoContent, gin.H{"error": fmt.Sprintf("no nfts found with id: %s", nftIDstr)})
+			c.JSON(http.StatusNoContent, gin.H{"error": fmt.Sprintf("no nfts found with id: %s", input.NftId)})
 			return
 		}
 
-		// TODO: this should just return a single NFT
-		c.JSON(http.StatusOK, gin.H{
-			"nfts": nfts,
-		})
+		if len(nfts) > 1 {
```
kaitoo1, line 39: Are ids a unique key in the db? Would it be possible for nfts to be > 1? 
I think it's ok to be safe here and check for this case, but maybe instead of an error we could return the first nft, and log an error message internally. ( if we don't have logging setup yet via Sentry or something, that could be a todo)

server/server.go
```diff
@@ -2,12 +2,29 @@ package server
...
 	"fmt"
+	"strings"
 
 	"github.com/gin-gonic/gin"
+	"github.com/go-playground/validator"
 	"github.com/mikeydub/go-gallery/runtime"
 	log "github.com/sirupsen/logrus"
 )
 
+var ethValidator validator.Func = func(fl validator.FieldLevel) bool {
```
kaitoo1, line 13: Could we define these validator funcs in their own separate file, to keep this server.go file lean?

server/nft.go
bennycio: oops! thank you!

server/nft.go
bennycio: id's are unique, so yes we could return first and log an error internally, thank you!

server/server.go
bennycio: yes indeed! will do!

PR #64, opened 2021-07-17
return empty arrays when requests are ambiguous

bennycio:
Changes:

- **Changed** endpoints that have more ambiguous requirements such as get nfts or collections by user_id and return an empty array instead of an error

### DIFF
--- a/server/collection.go
+++ b/server/collection.go
@@ -2,7 +2,6 @@ package server
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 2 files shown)


server/collection.go
```diff
@@ -33,8 +32,7 @@ func getAllCollectionsForUser(pRuntime *runtime.Runtime) gin.HandlerFunc {
 
 		colls, err := persist.CollGetByUserID(persist.DbId(userIDstr), c, pRuntime)
 		if len(colls) == 0 || err != nil {
-			c.JSON(http.StatusNoContent, gin.H{"error": fmt.Sprintf("no collections found with user_id: %s", userIDstr)})
-			return
+			colls = []*persist.Collection{}
```
fbrcsftnr, line 35: small note: could we differentiate `err` from an explicit "no collections found" error vs. an actual DB or internal server error?

for example, if our DB is down, i would expect this endpoint to not return an empty array and instead return an actual error.

thinking of something like this:
```go
if len(colls) == 0 || err == SomeMongoCursorNotFoundError {
  colls = []*persist.Collection{}
}
if err != nil {
  c.JSON(more serious error)
}
```

fbrcsftnr: note that this is low priority

fbrcsftnr (approved):

server/collection.go
kaitoo1: Second this ^

kaitoo1 (approved):

PR #65, opened 2021-07-19
Integration tests for Get NFT by ID route

fbrcsftnr:
**Added**
- `ErrorResponse` struct for standardizing error responses
- `server/t__main_test.go` file for triggering setup/teardown once for all package tests
- integration test utils for detecting server error responses
- integration tests for getNftById

**Changed**
- remove `omitempty` from `DeletedBool` to ensure the field gets set as `false` when saving to DB
- minor bugfixes, cleanup


bennycio (approved):
lookin good to merge!

PR #67, opened 2021-07-19
delete collection persist func

bennycio:
Changes:

- **Added** delete coll func
- **Changed** delete endpoint to use delete func


fbrcsftnr (approved):
🚀

PR #68, opened 2021-07-19
update endpoint and persist update

bennycio:
Changes:

- **Added** update endpoint and persist update func

### DIFF
--- a/server/collection.go
+++ b/server/collection.go
@@ -24,9 +24,9 @@ type collectionCreateInput struct {
 
 type collectionUpdateByIdInput struct {
 	Id     persist.DbId   `json:"id" binding:"required"`
-	Name   string         `json:"name"`
-	Nfts   []*persist.Nft `json:"nfts"`
-	Hidden bool           `json:"hidden"`
+	Name   string         `json:"name,omitempty"`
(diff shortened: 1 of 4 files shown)


server/collection.go
```diff
@@ -76,6 +83,30 @@ func createCollection(pRuntime *runtime.Runtime) gin.HandlerFunc {
...
+			c.JSON(http.StatusBadRequest, ErrorResponse{Error: err.Error()})
+			return
+		}
+
+		userId := c.GetString(userIdContextKey)
+
+		err := persist.CollUpdate(input.Id, persist.DbId(userId), &persist.Collection{
+			NameStr:    input.Name,
+			HiddenBool: input.Hidden,
+			NFTsLst:    input.Nfts,
```
fbrcsftnr, line 99: What if the user simply wants to update the collection name, so no NftsLst is provided (empty)? Would this override the DB with an empty list? Or will the field be omitted

fbrcsftnr: TLDR for updates, I think we only want to update fields that are explicitly provided by the client

fbrcsftnr (approved):

server/collection.go
bennycio: Good point!

PR #69, opened 2021-07-19
separate handlers for hidden/name/nfts

bennycio:
Changes:

- **Added** separate endpoints for different update types
- **Changed** gin.H errors to ErrorResponse

Considerations:

- Currently doing /update/name and /update/hidden etc.. Is there a better way we want to do endpoint naming for these? We could also have one handler and do /update/:type and do a switch statement on the type to determine what they are trying to update. Possibly even a required query param. What do you think?


fbrcsftnr:
> Currently doing /update/name and /update/hidden

i prefer this method, where we have explicit routes (instead of a single route with params) !

fbrcsftnr (approved):
thank you for the quick turnaround!

PR #71, opened 2021-07-20
Update nft by id test

fbrcsftnr:
- created global `TestConfig` for fixtures when running tests (user, jwt, etc)
- handle various `TestUpdateNftById` cases
- created `copy` package for stable copy strings
- fixed silent error with `c.GetString` (should be improved later)

### DIFF
--- a/server/collection.go
+++ b/server/collection.go
@@ -92,7 +92,7 @@ func createCollection(pRuntime *runtime.Runtime) gin.HandlerFunc {
 		//------------------
 		// CREATE
 
-		id, err := collectionCreateDb(input, persist.DbId(userId), c, pRuntime)
+		id, err := collectionCreateDb(input, userId, c, pRuntime)
 		if err != nil {
 			c.JSON(http.StatusInternalServerError, ErrorResponse{
 				Error: err.Error(),
@@ -118,7 +118,7 @@ func updateCollectionName(pRuntime *runtime.Runtime) gin.HandlerFunc {
 
 		coll := &persist.Collection{NameStr: input.Name}
 
-		err := persist.CollUpdate(input.Id, persist.DbId(userId), coll, c, pRuntime)
+		err := persist.CollUpdate(input.Id, userId, coll, c, pRuntime)
 		if err != nil {
 			c.JSON(http.StatusInternalServerError, ErrorResponse{Error: err.Error()})
 			return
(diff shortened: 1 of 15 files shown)


bennycio (approved):
looking clean!

server/t__routes_nft_test.go
```diff
@@ -54,4 +56,33 @@ func TestGetNftById_NotFoundError(t *testing.T) {
...
+	// build update request body
+	type Update struct {
+		Id   string `json:"id"`
+		Name string `json:"name"`
+	}
+	update := Update{Name: "new nft name", Id: string(nftId)}
+	data, err := json.Marshal(update)
+	assert.Nil(err)
+
+	resp, err := http.Post(fmt.Sprintf("%s/nfts/update?id=%s", serverUrl, nftId),
```
kaitoo1, line 79: does this request need to be authenticated?

fbrcsftnr: @kaitoo1 it does indeed! we were talking about seeding a test JWT at the beginning of a test suite

kaitoo1 (approved):

kaitoo1:
SO SICK
i love tests

server/collection.go
```diff
@@ -85,12 +85,14 @@ func createCollection(pRuntime *runtime.Runtime) gin.HandlerFunc {
...
-		ownerId := c.GetString(userIdContextKey)
+		// TODO: make this a util, especially since coercion to persist.DbId may break
+		val, _ := c.Get(userIdContextKey)
+		userId := val.(persist.DbId)
 
 		//------------------
 		// CREATE
 
-		id, err := collectionCreateDb(input, persist.DbId(ownerId), c, pRuntime)
+		id, err := collectionCreateDb(input, persist.DbId(userId), c, pRuntime)
```
bennycio, line 95: `userId` is already of type `persist.DbId` so does not need a cast here

server/middleware.go
```diff
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/gin-gonic/gin"
+	"github.com/mikeydub/go-gallery/copy"
 	"github.com/mikeydub/go-gallery/runtime"
 )
```
bennycio, line 12: For getting UserId from context we should be able to do this as a sort of context extension function.
```suggestion
func (c *gin.Context) getUserId() persist.DbId {
val, _ := c.Get(userIdContextKey)
return val.(persist.DbId)
}
```

bennycio (approved):

server/collection.go
fbrcsftnr: great point, addressed!

server/middleware.go
fbrcsftnr: this would be awesome, but i'm not sure how to address this- as far as i know, you can only define methods on a type defined in that same package. 

to implement your strategy we'd probably need to create a new type then pass it all the way down right? either way, i'll merge this for now and maybe you can take a look later!

PR #73, opened 2021-07-20
authentication adjustments and additions

bennycio:
In progress

### DIFF
--- a/server/user.go
+++ b/server/user.go
@@ -57,7 +57,7 @@ type userCreateOutput struct {
 //-------------------------------------------------------------
 // HANDLERS
 
-func updateUserAuth(pRuntime *runtime.Runtime) gin.HandlerFunc {
+func updateUser(pRuntime *runtime.Runtime) gin.HandlerFunc {
 	return func(c *gin.Context) {
 
 		up := &userUpdateInput{}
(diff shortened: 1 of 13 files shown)


bennycio:
Changes:

- **Added** rotating nonces after login or user creation
- **Added** checks to ensure the right information is available for a login versus signup
- **Changed** auth tests to compile (not necessarily functional yet, see considerations)

Considerations:

- server/t__auth_user_test.go requires a signature. How should we sign the nonce from go?

server/user.go
```diff
@@ -301,32 +301,27 @@ func userIsValid(pAddress string,
...
+	if user != nil {
+		userIdStr = user.IDstr
+	} else {
+		return "", "", errors.New("no user found")
 	}
-
 	//------------------
 
-	return userExistsBool, nonceValueStr, userIDstr, nil
+	return
```
kaitoo1, line 326: should this return the nonceValueStr and userIDStr?
 according to the comment `returns nonce value string, user id`

server/user.go
```diff
@@ -281,18 +287,12 @@ func userDeleteDb(pUserIDstr persist.DbId,
 }
 
 //-------------------------------------------------------------
-func userIsValid(pAddress string,
+// returns nonce value string, user id
+// will return empty strings and error if no nonce found
+// will return empty string if no user found
+func userWithNonce(pAddress string,
```
kaitoo1, line 293: maybe this can be renamed to something like `getUserWithNonce` or `getUserAndNonce` - a verb can help clarify the intent/expected outcome of the function which makes it easier to read when it's called in other places

kaitoo1 (approved):

server/user.go
bennycio: Since the return values are named, go will allow an empty return statement to return the current values of the named return values. We could explicitly write what we are returning as well, whichever will do.

server/user.go
bennycio: For sure!

PR #74, opened 2021-07-21
Add Gin CORS middleware

kaitoo1:
This PR adds and configures Gin's CORS middleware so that we can try and get the frontend working with backend on dev
https://github.com/gin-contrib/cors

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/gin-gonic/gin/binding"
+	"github.com/gin-contrib/cors"
 	"github.com/go-playground/validator"
 	"github.com/mikeydub/go-gallery/runtime"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -18,6 +19,10 @@ func Init(pPortInt int,
 
 	pRuntime.Router = gin.Default()
 
+	config := cors.DefaultConfig()
+	config.AllowOrigins = []string{"https://gallery-dev.vercel.app/"}
```
fbrcsftnr, line 23: should we make this an env var?

fbrcsftnr: also i'm wondering if the default will handle pre-flight requests

fbrcsftnr: (i think it will)

kaitoo1: good call on env var - I was thinking just test it out first, and once it works I'll refactor to put things in the right place

bennycio (approved):

kaitoo1:
Closed in favor of #76

PR #75, opened 2021-07-21
Move .env to bin and flexibily point to .env on startup

kaitoo1:
Move .env to bin and flexibily point to .env on startup


fbrcsftnr (approved):
👏

PR #76, opened 2021-07-22
Add CORS middleware to set response headers and handle OPTIONS method

kaitoo1:
Closed #74 in favor of this one. 
Instead of installing and using the official Gin CORS middleware, this pr just adds the headers we need (for now)


fbrcsftnr (approved):
🔥🔥🔥🔥🔥

PR #77, opened 2021-07-23
aggregations handle deleted entities

bennycio:
Changes:

- **Added** new filter expressions in match aggregation cases for filtering out deleted entities

### DIFF
--- a/persist/collection.go
+++ b/persist/collection.go
@@ -78,7 +78,7 @@ func CollGetByUserID(pUserIDstr DbId,
 
 	result := []*Collection{}
 
-	fil := bson.M{"owner_user_id": pUserIDstr}
+	fil := bson.M{"owner_user_id": pUserIDstr, "deleted": false}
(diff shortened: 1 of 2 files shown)


persist/collection.go
```diff
@@ -201,10 +203,25 @@ func newCollectionPipeline(matchFilter bson.M) mongo.Pipeline {
...
+						"$and": []bson.M{
+							{"$in": []string{"$_id", "$$array"}},
+							{"$eq": []interface{}{"$deleted", false}},
+						},
+					},
+				}}},
+			},
+			"as": "nfts",
 		}}},
+		// {{Key: "$lookup", Value: bson.M{
```
kaitoo1, line 220: delete?

kaitoo1 (approved):

persist/collection.go
bennycio: For sure!

PR #78, opened 2021-07-24
Rename ID field on auth schema to IDstr

kaitoo1:
We get a duplicate key error when inserting a record into the db (and 1 record already exists in the db), because the `_id` field is getting set as an empty string. I encountered this when trying to generate a nonce more than once.

It looks like it's just an issue setting the ID field using an incorrect field name. We're using `IDstr` instead of `ID` which is what is set in the schema.

This PR changes `IDstr` to `ID` in our mongo Insert function so that the id gets persisted.

Empty ID before fix:
Screen Shot 2021-07-24 at 10 51 38 AM

 ID after fix:
Screen Shot 2021-07-24 at 11 28 24 AM

@bennycio @fbrcsftnr  Do we prefer `ID` or `IDstr` ?  we can use `IDstr` if we want to be consistent with our other fields. But I don't feel strongly since this field has the type `DbId`, which is also a `string` so both are correct

### DIFF
--- a/persist/mongo.go
+++ b/persist/mongo.go
@@ -33,8 +33,9 @@ func (m *MongoStorage) Insert(ctx context.Context, insert interface{}, opts ...*
 	elem := reflect.TypeOf(insert).Elem()
 	val := reflect.ValueOf(insert).Elem()
 	now := float64(time.Now().UnixNano()) / 1000000000.0
-	if _, ok := elem.FieldByName("IDstr"); ok {
-		idField := val.FieldByName("IDstr")
+
+	if _, ok := elem.FieldByName("ID"); ok {
+		idField := val.FieldByName("ID")
 		if !idField.CanSet() {
 			// panic because this literally cannot happen in prod
 			panic("unable to set id field on struct")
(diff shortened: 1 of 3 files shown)


persist/mongo.go
```diff
@@ -33,8 +33,9 @@ func (m *MongoStorage) Insert(ctx context.Context, insert interface{}, opts ...*
 	elem := reflect.TypeOf(insert).Elem()
 	val := reflect.ValueOf(insert).Elem()
 	now := float64(time.Now().UnixNano()) / 1000000000.0
-	if _, ok := elem.FieldByName("IDstr"); ok {
-		idField := val.FieldByName("IDstr")
+
+	if _, ok := elem.FieldByName("ID"); ok {
+		idField := val.FieldByName("ID")
```
kaitoo1, line 38: Same issue exists below in `InsertMany()`

It would be good to decide whether we use `IDstr` or `ID` and find+replace across our codebase

fbrcsftnr (approved):

bennycio (approved):
Using IDstr for everything is looking good to me!

PR #81, opened 2021-07-25
sync test and fixes

bennycio:
Changes:

- **Added** tests for opensea sync
- **Changed** funcs for upserting, checking difference, etc. to ensure all opensea nfts are reflected accurately on gallery

Considerations:

- Should we mark nfts that are not on opensea for a user but on gallery as 1. deleted, 2. remove their `owner_user_id` and `owner_address`, or 3. remove their `owner_user_id` and set their `owner_address` to what it is on opensea. Currently I have implemented 2 because deleting feels odd given that the nft's themselves can't really be deleted and later on someone else could come along with that same nft.

### DIFF
--- a/persist/nft.go
+++ b/persist/nft.go
@@ -2,7 +2,6 @@ package persist
 
 import (
 	"context"
-	"errors"
 	"time"
 
 	"github.com/mikeydub/go-gallery/runtime"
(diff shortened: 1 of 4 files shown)


fbrcsftnr:
I'm in favor of 3!

bennycio:
> I'm in favor of 3!

I realized as I tried to implement this that this would require more roundtrips to opensea than I thought seeing as opensea does not give us the NFT's that a user used to own with the address of the new owner that we would need for option three. This should not matter because whenever the owner of the NFT signs on it will still sync up to them anyway. 

That being said, I think this PR is complete unless anything else comes to mind.

kaitoo1 (approved):

PR #84, opened 2021-07-27
Bunch of fixes to get onboarding e2e closer to working

kaitoo1:
Made a couple of changes to code that gets hit during onboarding, up until right after Update User

- changed Mongo Update function to return a 200 instead of 500 if no documents were modified
- changed signature length validation to medium_string bc short_string was too short
- changed Login handler to return error instead of 200 if json mapping error (now its 500, but it should probably be 400)
- fix authJwtParse, was returning an error for valid jwt
- add UserId to user Get output


server/user.go
```diff
@@ -29,6 +29,7 @@ type userGetInput struct {
 
 // OUTPUT - USER_GET
 type userGetOutput struct {
+	UserId      string   ` json:"id"`
```
bennycio, line 32: This could be of type `persist.DbId` for consistency because in JSON marshalling it will be converted to a string anyway. Whichever you think is better is fine though!

bennycio (approved):
Thanks Kaito! Looks fantastic!

server/user.go
kaitoo1: Awesome, I wasn't sure if it'd auto-convert. I've changed it to `persist.DbId` for consistency 👍

server/user.go
```diff
@@ -29,9 +29,10 @@ type userGetInput struct {
 
 // OUTPUT - USER_GET
 type userGetOutput struct {
-	UserNameStr string   ` json:"username"`
-	BioStr      string   ` json:"bio"`
-	Addresses   []string ` json:"addresses"`
+	UserId      persist.DbId ` json:"id"`
```
fbrcsftnr, line 32: nit: remove space in front of the backtick `

kaitoo1: tru. 
fixed in follow up PR: https://github.com/gallery-so/go-gallery/pull/86

PR #88, opened 2021-07-29
get user id from context

bennycio:
Changes:

- **Added** helper func for getting userid from gin context
- **Changed** funcs that manually got user id from context to use helper func

Considerations:

- In the funcs that have the jwtRequired middleware, do we need to use the ok bool with the helper function since we can guarantee that the userId will be there?

### DIFF
--- a/server/collection.go
+++ b/server/collection.go
@@ -85,9 +85,11 @@ func createCollection(pRuntime *runtime.Runtime) gin.HandlerFunc {
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


fbrcsftnr:
> In the funcs that have the jwtRequired middleware, do we need to use the ok bool with the helper function since we can guarantee that the userId will be there?

yes- better safe than sorry, in case we move that middleware at some point

fbrcsftnr (approved):

PR #90, opened 2021-07-30
collections get by id

bennycio:
Changes:

- **Changed** collection get func and input struct to also accept getting by id

### DIFF
--- a/server/collection.go
+++ b/server/collection.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 
 	"github.com/gin-gonic/gin"
@@ -10,13 +11,11 @@ import (
 	"github.com/mikeydub/go-gallery/runtime"
 )
 
-type collectionGetByIdInput struct {
-	Id persist.DbId `form:"id" json:"id" binding:"required"`
+type collectionGetInput struct {
+	Id     persist.DbId `form:"id" json:"id"`
+	UserId persist.DbId `form:"user_id" json:"user_id"`
 }
 
-type collectionGetByUserIdInput struct {
-	UserId persist.DbId `form:"user_id" json:"user_id" binding:"required"`
-}
 type collectionGetOutput struct {
 	Collections []*persist.Collection `json:"collections"`
 }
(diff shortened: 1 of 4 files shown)


server/collection.go
```diff
@@ -50,12 +49,12 @@ type collectionDeleteInput struct {
 //-------------------------------------------------------------
 // HANDLERS
 
-func getAllCollectionsForUser(pRuntime *runtime.Runtime) gin.HandlerFunc {
+func getCollections(pRuntime *runtime.Runtime) gin.HandlerFunc {
```
kaitoo1, line 52: Good call adding the ability to get a collection by id! We'll definitely need that.

I think this is a good candidate for splitting into 2 different handlers and endpoints.

I see two separate pieces of functionality: 
- retrieving a single collection for a given collection id
- retrieving all collections for a given user Id

And they don't really overlap, because it doesn't look like you'd pass in **both** a collection id and user id together as a "valid" input.

So instead of one method doing both depending on what the input is, it would be simpler to have separate handlers like `getCollectionById` and `getCollectionsForUser`, as well as separate endpoints for each action. 

That way the expected outcome is more explicit, which generally makes code easier to use (for both consumers of the endpoint as well as devs working on related code) and easier to maintain (less branching logic = less bugs, also easier to remember what's happening when you look at this code 4 months from now)

kaitoo1 (changes requested):
Good call adding the ability to get collections by id! I left a comment in `collection.go` about how we could reduce some of the complexity inside the handler and make it more explicit.

server/collection.go
```diff
@@ -64,16 +65,51 @@ func getAllCollectionsForUser(pRuntime *runtime.Runtime) gin.HandlerFunc {
...
 			colls = []*persist.Collection{}
 		}
 
 		c.JSON(http.StatusOK, collectionGetOutput{Collections: colls})
+
 	}
 }
 
+//-------------------------------------------------------------
+func getCollectionsById(pRuntime *runtime.Runtime) gin.HandlerFunc {
```
kaitoo1, line 79: should this be `getCollectionById` ? (singular) 
The name suggests it returns multiple collections but it looks like it queries for and returns 1

PR #94, opened 2021-08-07
username indexing

bennycio:
Changes:

- **Added** username unique/collation index
- **Changed** update user func to take in generic param and allow for specific fields to be updated without adjusting other fields

Considerations:

- Still needs a good test to ensure collation works as expected


persist/mongo.go
```diff
@@ -198,3 +202,8 @@ func generateId(creationTime float64) DbId {
 	hexStr := hex.EncodeToString(sum)
 	return DbId(hexStr)
 }
+
+// function that returns the pointer to the bool passed in
```
fbrcsftnr, line 206: Lol

fbrcsftnr (approved):

PR #98, opened 2021-08-08
sanitize bio and set username requirements

bennycio:
Changes:

- **Added** username validator
- **Added** sanitization library and policy to sanitize bio on update

Considerations:

- Any other fields need sanitizing on any other struct?

### DIFF
--- a/server/validate.go
+++ b/server/validate.go
@@ -1,10 +1,16 @@
 package server
 
 import (
+	"regexp"
 	"strings"
 
 	"github.com/go-playground/validator/v10"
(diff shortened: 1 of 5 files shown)


server/validate.go
```diff
@@ -1,11 +1,17 @@
 package server
 
 import (
+	"regexp"
 	"strings"
 
 	"github.com/go-playground/validator/v10"
+	"github.com/microcosm-cc/bluemonday"
```
Robinnnnn, line 8: haven't heard of bluemonday- looks awesome!!

Robinnnnn (approved):

PR #121, opened 2021-08-27
pull operation to ensure no duplicate nfts

bennycio:
Changes

- **Added** a pull operation to remove an array of items from an array field of queried documents
- **Added** an update nfts func that will first pull nfts from all other documents before adding it to the new one.
- **Changed** create coll func to also remove nfts from other colls that have the nfts it is being created with

Considerations:

- Needs tests still...

### DIFF
--- a/persist/collection.go
+++ b/persist/collection.go
@@ -76,8 +76,16 @@ func CollCreate(pCtx context.Context, pColl *CollectionDB,
 
 	mp := newStorage(0, collectionColName, pRuntime)
 
+	if pColl.OwnerUserID == "" {
+		return "", errors.New("owner_user_id is required")
+	}
+
 	if pColl.Nfts == nil {
 		pColl.Nfts = []DBID{}
+	} else {
+		if err := mp.pull(pCtx, bson.M{"owner_user_id": pColl.OwnerUserID}, "nfts", pColl.Nfts); err != nil {
+			return "", err
+		}
 	}
 
 	return mp.insert(pCtx, pColl)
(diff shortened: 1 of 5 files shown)


persist/storage.go
```diff
@@ -148,6 +148,21 @@ func (m *storage) Push(ctx context.Context, query bson.M, field string, value in
 	return nil
 }
 
+// pull puls items from an array field for a given queried document(s)
```
Robinnnnn, line 151: pulls* :)

persist/storage.go
```diff
@@ -148,6 +148,21 @@ func (m *storage) Push(ctx context.Context, query bson.M, field string, value in
...
+// value must be an array
+func (m *storage) pull(ctx context.Context, query bson.M, field string, value interface{}) error {
+
+	result, err := m.collection.UpdateOne(ctx, query, bson.D{{Key: "$pull", Value: bson.M{field: bson.M{"$in": value}}}})
```
Robinnnnn, line 155: wow, TIL $pull!

persist/collection.go
```diff
@@ -156,6 +164,24 @@ func CollUpdate(pCtx context.Context, pIDstr DBID,
...
+// being included in the updated collection. This is to ensure that the NFTs are not
+// being shared between collections.
+func CollUpdateNFTs(pCtx context.Context, pIDstr DBID,
+	pUserID DBID,
+	pUpdate *CollectionUpdateNftsInput,
+	pRuntime *runtime.Runtime) error {
+
+	mp := newStorage(0, collectionColName, pRuntime).withRedis(CollectionsUnassignedRDB, pRuntime)
+
+	if err := mp.pull(pCtx, bson.M{"owner_user_id": pUserID}, "nfts", pUpdate.Nfts); err != nil {
```
Robinnnnn, line 178: question: what is `pull` doing in this case where you're providing `owner_user_id: pUserID` as the query? is it in case the NFTs exist in other collections _that belong to the current user?_

i'm wondering if we need to check _all_ NFTs that belong to _any_ user and pull those. for example, if i were trying to update my collection with an NFT that belongs to DCInvestor, would this operation catch this?

bennycio: this will only remove it from your other collections, if we want it to be global we could just make this query every collection in the DB, could be a pretty expensive operation. I believe too that if an NFT moves accounts, it won't show up in the old account's collections even if it is technically still there in the db

PR #122, opened 2021-08-27
Benny/opensea fields

bennycio:
Changes:

- **Added** new structs representing inner JSON objects for unmarshalling from opensea API response 
- **Changed** mapping of opensea NFT to gallery NFT to reflect new openseaNFT fields


Robinnnnn (approved):
looks like there's some stuff here from your other `pull` PR!

PR #134, opened 2021-09-07
date-times, pull-all, fix tests

bennycio:
Changes:

- **Changed** db structs to use `primitive.DateTime` instead of floats with magic numbers
- **Changed** `$pull` to `$pullAll` which accomplishes the same thing we do with pull with less steps
- **Changed** generate ID func to take in an `interface{}` which represents the whole struct not just the time. Since time can overlap (which it did during a test) we don't want to be able to accidentally create the same ID twice. Hashing the entire struct is more reliable and no less efficient than just using time
- **Changed** collections routes tests to use a cleanup func on each test because `TestMain` does not tear down db in between each test leading to unexpected outcomes.


persist/auth.go
```diff
@@ -6,6 +6,7 @@ import (
 
 	"github.com/mikeydub/go-gallery/runtime"
 	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson/primitive"
 	"go.mongodb.org/mongo-driver/mongo/options"
 	// "github.com/davecgh/go-spew/spew"
```
Robinnnnn, line 11: toss?

server/t__routes_collection_test.go
```diff
@@ -120,6 +124,7 @@ func TestCreateCollection_Success(t *testing.T) {
 }
 
 func TestGetUnassignedCollection_Success(t *testing.T) {
+	t.Cleanup(clearDB)
```
Robinnnnn, line 127: i'm surprised this cleanup doesn't break existing tests that require some "global" seeded variables

Robinnnnn (approved):

server/t__routes_collection_test.go
kaitoo1: Instead of calling `t.Cleanup()` from every test, could it be done in a setup or teardown function that gets called before and after each test?
If we want it to happen for every test I think it should be done automatically

kaitoo1 (changes requested):
Date/time change looks good, I left a comment automating `t.Cleanup(clearDB)` so we don't have to call it from each test

server/t__routes_collection_test.go
bennycio: I did some research and there currently is not a way in Go to run something before or after each test, TestMain is the closest thing and that just runs after/before all of the tests

PR #135, opened 2021-09-07
make funcs for requests

bennycio:
Changes:

- **Added** funcs for requests in tests so we aren't cluttering our tests with awkward go http requests


Robinnnnn (approved):
so satisfying!

PR #145, opened 2021-09-10
add collection nft

bennycio:
Changes:

- **Added** struct representing an NFT in a collection that has less return fields

Considerations:

- Did I choose the right fields?


Robinnnnn:
Thanks for doing this! Some additional fields we need:
- nft.asset_contract.name => https://github.com/gallery-so/gallery/blob/main/src/components/NftPreview/NftPreviewLabel.tsx
- nft.asset_contract.asset_image_url => https://github.com/gallery-so/gallery/blob/917f4b845228b1b2a16087d6b767089095a10d18/src/utils/resizeNftImageUrl.ts#L12

PR #146, opened 2021-09-10
banned usernames

bennycio:
Changes:

- **Added** a map of banned usernames and a small check to see if a username is banned before creation


Robinnnnn:
also: 404, nuke

PR #149, opened 2021-09-13
setup potential multiple dbs and alchemy rpc

bennycio:
Changes:

- **Added** infra package for using alchemy rpc to get token information without opensea
- **Changed** runtime package to support connecting to multiple databases (e.g. we could have a gallery db for gallery users and collections and an infra database for storing block chain data)

Todo:

- [ ] Get ERC721 TokenURI

### DIFF
--- a/infra/rpc.go
+++ b/infra/rpc.go
@@ -1,0 +1,171 @@
+package infra
+
+import (
+	"os"
+	"sort"
+
+	"github.com/ethereum/go-ethereum/rpc"
+)
+
+// RPC is a wrapper for the RPC client
+type RPC struct {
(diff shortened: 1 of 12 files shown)


bennycio:
Currently I have extended the current runtime config to support some variable for infra, maybe we should just create a separate infra config struct? Some variables would be shared though between both configs so it might be worth it to just keep it all together.

Robinnnnn:
> Currently I have extended the current runtime config to support some variable for infra, maybe we should just create a separate infra config struct? Some variables would be shared though between both configs so it might be worth it to just keep it all together.

down to make a separate config if this ends up being sufficiently different!

PR #150, opened 2021-09-14
whitelist preview url for cors

kaitoo1:
Add preview url that we're going to use for beta testing to the list of whitelisted urls so that CORS will work


bennycio:
After carefully reading every character, I have determined that this PR is good to go 💯

bennycio (approved):

PR #151, opened 2021-09-15
events and feeds

bennycio:
Changes:

- **Added** structs representing feeds and events for mongo
- **Added** funcs for creating and broadcasting events to feeds
- **Added** funcs for creating individual events relevant to other parts of the application such as creating a new collection and updating a collections info
- **Changed** funcs that should broadcast events to optionally broadcast events should they have not been broadcasted for that collection/user/nft/gallery recently

Considerations:

- Still needs testing
- Currently event related thinks are happening in goroutines because event data does not need to be returned in every case so far where an event is broadcasted. For example, when a collection is created with `persist.CollCreate` the caller does not expect any data on the event created during that process. The event also does not need to block anything else from happening because the feed being accurate by the millisecond isn't necessary it seems. Therefore, event stuff happens asynchronously right now. Is that how it should be?


persist/collection.go
```diff
@@ -278,6 +290,84 @@ func CollGetUnassigned(pCtx context.Context, pUserID DBID, skipCache bool, pRunt
 
 }
 
+func hasRecentCollectionEvent(pCtx context.Context, pEventType int, pCollectionID DBID, pRuntime *runtime.Runtime) (bool, error) {
+	mp := newStorage(0, eventColName, pRuntime)
+
+	query := bson.M{"type": pEventType, "created_at": bson.M{"$gte": time.Now().Add(-eventRecencyThreshold)}}
```
Robinnnnn, line 296: nice!

persist/feed.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func feedCreate(pCtx context.Context, pUserID DBID, pRuntime *runtime.Runtime) (DBID, error) {
+	mp := newStorage(0, feedColName, pRuntime)
+	feed := &Feed{UserID: pUserID}
+	return mp.insert(pCtx, feed)
+}
+
+// FeedGetByUserID gets a feed from the DB by user ID
+func FeedGetByUserID(pCtx context.Context, pUserID DBID, pRuntime *runtime.Runtime) ([]*Feed, error) {
```
Robinnnnn, line 108: interesting to see a feed made per user ID, rather than a global one!

what would happen if a logged out user looks at the homepage/feed?

bennycio: Very true, did not consider this possibility somehow! I remembered hearing you talk about possibly the way twitter did feeds and how it kind of pre-prepared a feed for each user, maybe I misinterpreted but this was my implementation of that. Maybe to match both cases we could keep all events in a separate collection and have a user collection specifically for keeping track of what kind of events we want them to see, how to order it, and how active they are so we know if it needs updating. In the future if we are personalizing the feed in any way that would be useful as well.

PR #153, opened 2021-09-15
make histories work and do some async

bennycio:
Changes:

- **Added** a check to make sure ownership history populates in opensea test
- **Added** a function to sync ownership histories with async for each NFT but blocking for the group of NFTs as a whole. This will find the history for each individual NFT, each in its own goroutine, and once all of the goroutines have completed, the result will be returned. The function will block until all goroutines have been completed. 
- **Added** a func to get nft from DB by opensea ID
- **Changed** sync history function to find the NFT in the DB that it represents the history of and make sure to store that NFTs ID with the history in the database


persist/nft.go
```diff
@@ -175,19 +193,25 @@ func NftUpdateByID(pCtx context.Context, pID DBID, pUserID DBID, pUpdate interfa
...
 	mp := newStorage(0, nftColName, pRuntime)
 
 	upsertModels := make([]mongo.WriteModel, len(pNfts))
 
+	allIDS := []DBID{}
+
 	for i, v := range pNfts {
 
+		if v.ID != "" {
+			allIDS = append(allIDS, v.ID)
```
Robinnnnn, line 207: i'd need to double check this but i'm pretty sure appending dynamically within a for loop is not super performant. an improvement would be to `make` a slice in advance with the length of `pNfts`, and just assign the IDs via index in the for loop.

server/opensea.go
```diff
@@ -98,25 +100,57 @@ func openSeaPipelineAssetsForAcc(pCtx context.Context, pOwnerWalletAddress strin
...
 
-func openseaSyncHistory(pCtx context.Context, pNFTID persist.DBID, pTokenID string, pTokenContractAddress string, pRuntime *runtime.Runtime) (*persist.OwnershipHistory, error) {
+func openseaSyncHistories(pCtx context.Context, pNfts []*persist.Nft, pRuntime *runtime.Runtime) ([]*persist.Nft, error) {
+	resultChan := make(chan *persist.Nft)
+	errorChan := make(chan error)
+	updatedNfts := []*persist.Nft{}
+	// atomic counter to track number of goroutines until all are done
+	left := int64(len(pNfts))
+	// create a goroutine for each nft and sync history
+	for _, nft := range pNfts {
```
Robinnnnn, line 123: beautiful use of channels!!

Robinnnnn (approved):

PR #155, opened 2021-09-16
rate limit, faster sync, auth only

bennycio:
Changes:

- **Added** a rate limiter middleware and used it for opensea sync
- **Changed** opensea sync to be an authed endpoint
- **Changed** bulk upsert func to use async (cut average test time down to 1/3 of what it was)

### DIFF
--- a/server/middleware.go
+++ b/server/middleware.go
@@ -17,6 +17,8 @@ const (
 	authContextKey   = "authenticated"
 )
 
+var rateLimiter = NewIPRateLimiter(1, 5)
+
 func jwtRequired(runtime *runtime.Runtime) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		header := c.GetHeader("Authorization")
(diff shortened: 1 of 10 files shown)


server/middleware.go
```diff
@@ -17,6 +17,8 @@ const (
 	authContextKey   = "authenticated"
 )
 
+var rateLimiter = NewIPRateLimiter(1, 5)
```
kaitoo1, line 20: What limit does this translate to? in terms of # of allowed requests?

server/nft.go
```diff
@@ -154,6 +156,22 @@ func getNftsFromOpensea(pRuntime *runtime.Runtime) gin.HandlerFunc {
 			c.JSON(http.StatusBadRequest, errorResponse{Error: err.Error()})
 			return
 		}
+
+		userID, ok := getUserIDfromCtx(c)
+		if !ok {
+			c.JSON(http.StatusBadRequest, errorResponse{Error: "user id not found in context"})
+			return
+		}
+		ownsWallet, err := doesUserOwnWallet(c, userID, input.WalletAddress, pRuntime)
```
kaitoo1, line 165: Nice!!

server/middleware.go
bennycio: This basically means that any IP can call our endpoint 5 times before they get rate limited. There is basically an imaginary bucket that holds tokens and a new token gets added every second until there is 5. When a user calls the endpoint it takes a token from the bucket. If there are no tokens left in the bucket it says no so will basically limit them to 1 second in between each request after that.

PR #156, opened 2021-09-17
remove sync nfts from other collections

bennycio:
Changes:

- **Added** a func that will remove a set of NFTs from all other user's collections and remove any of your NFTs that are not a part of that set. In other words, ensure that only you and ONLY you can display only the NFTs passed into that function
- **Changed** opensea sync to use this func

### DIFF
--- a/server/t__opensea_test.go
+++ b/server/t__opensea_test.go
@@ -13,36 +13,46 @@ func TestFetchAssertsForAcc(t *testing.T) {
 	setupTest(t)
 	ctx := context.Background()
 
-	user := &persist.User{Addresses: []string{"0x485b8ac36535fae56b2910780245dd69dda270bc"}}
-
-	userID, err := persist.UserCreate(ctx, user, tc.r)
+	robin := &persist.User{Addresses: []string{"0x70d04384b5c3a466ec4d8cfb8213efc31c6a9d15"}}
+	gianna := &persist.User{Addresses: []string{"0xdd33e6fd03983c970ae5e647df07314435d69f6b"}}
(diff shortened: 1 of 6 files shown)


server/t__opensea_test.go
```diff
@@ -13,36 +13,46 @@ func TestFetchAssertsForAcc(t *testing.T) {
...
 
 	_, err = persist.NftCreateBulk(ctx, []*persist.Nft{nft, nft2}, tc.r)
 	assert.Nil(t, err)
 
-	nfts, err := openSeaPipelineAssetsForAcc(ctx, "0x485b8ac36535fae56b2910780245dd69dda270bc", true, tc.r)
+	now, err := persist.NftGetByUserID(ctx, giannaUserID, tc.r)
+	assert.Nil(t, err)
+	assert.Len(t, now, 1)
+
+	nfts, err := openSeaPipelineAssetsForAcc(ctx, robinUserID, "0x70d04384b5c3a466ec4d8cfb8213efc31c6a9d15", true, tc.r)
```
Robinnnnn, line 44: does this actually send a request out to opensea? if so, in the future we should set up some mock infrastructure to mock responses to save on network/time (with certain exceptions where we def wanna call out to opensea)

Robinnnnn: https://github.com/stretchr/testify#mock-package

Robinnnnn (approved):

server/t__opensea_test.go
bennycio: this does send a request to opensea, will definitely look into that package!

PR #158, opened 2021-09-20
aggregate owner username for nft get

bennycio:
Changes:

- **Added** steps to aggregation pipeline for NFTs to go get username from user collection
- **Added** `NftDB` struct and `Nft` struct that represent Nfts with or without aggregation additions
- **Added** check in nft get by ID test to make sure username is filled

### DIFF
--- a/persist/nft.go
+++ b/persist/nft.go
@@ -18,7 +18,45 @@ const (
 	nftCollectionColName = "nft_collections"
 )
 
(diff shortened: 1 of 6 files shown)


kaitoo1 (approved):

kaitoo1:
Sickkk looks great
Separating `NftDB` from `Nft` is clean, and good call with updating the test 🔥

PR #159, opened 2021-09-20
block signup with fake contract check

bennycio:
Changes:

- **Added** check in preflight for owning a non-existent NFT in the meantime while we make contracts
- **Removed** redeemable contarct


server/auth.go
```diff
@@ -304,6 +304,14 @@ func authUserGetPreflightDb(pCtx context.Context, pInput *authUserGetPreflightIn
 	var nonce *persist.UserNonce
 	if !userExistsBool {
 
+		hasNFT, err := hasAnyNFT(pCtx, "0x0", pInput.Address, pRuntime)
+		if err != nil {
+			return nil, err
+		}
+		if !hasNFT {
+			return nil, errors.New("user does not own required nft to signup")
```
kaitoo1, line 312: 🚀

kaitoo1: Nice! functionality looks great - have we been able to test this gating with a valid nft ?

Also related to testing - I guess the fact we can make this change without updating any tests means we need to add some tests for the User Creation/Sign up flow. 
That doesn't have to been part of this PR - just calling it out as something we should add to our todos

Otherwise PR looks good ✅

kaitoo1 (approved):

PR #160, opened 2021-09-21
Map OS asset's external_link to our NFT model

kaitoo1:
Changes:
- Change the field that we get the NFT's external link from during the OpenSea -> Gallery NFT mapping

### DIFF
--- a/server/opensea.go
+++ b/server/opensea.go
@@ -31,7 +31,7 @@ type openseaAsset struct {
 	Name        string `json:"name"`
 	Description string `json:"description"`
 
-	ExternalURL      string           `json:"external_url"`
+	ExternalURL      string           `json:"external_link"`
 	TokenMetadataURL string           `json:"token_metadata_url"`
 	Creator          openseaAccount   `json:"creator"`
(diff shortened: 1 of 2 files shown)


server/opensea.go
```diff
@@ -31,7 +31,7 @@ type openseaAsset struct {
 	Name        string `json:"name"`
 	Description string `json:"description"`
 
-	ExternalURL      string           `json:"external_url"`
+	ExternalURL      string           `json:"external_link"`
```
kaitoo1, line 34: `external_url` doesn't look like it exists on the OS asset, but `external_link` does

I kept our internal name as `ExternalURL` instead of changing it to `ExternalLink` bc that felt more consistent with the rest of our internal naming scheme

kaitoo1:
@bennycio do I need to change anything in the migration script to reflect this change?

bennycio:
> @bennycio do I need to change anything in the migration script to reflect this change?

Yes, line 101 in migrations script uses external_url instead of external_link. Other than that, this PR is fantastic thank you for finding that little bug!

bennycio (approved):

PR #161, opened 2021-09-21
Include user created_at time in Get User response

kaitoo1:
see title

### DIFF
--- a/server/user.go
+++ b/server/user.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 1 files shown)


server/user.go
```diff
@@ -332,6 +334,7 @@ func userGetDb(pCtx context.Context, pInput *userGetInput,
 		UserID:      user.ID,
 		UserNameStr: user.UserName,
 		BioStr:      user.Bio,
+		CreatedAt:   user.CreationTime.Time(),
```
kaitoo1, line 337: @bennycio 
This was the first way I found that works, lmk if there's a better way to do this! or different type we should use

bennycio: This works perfectly!

bennycio (approved):

PR #162, opened 2021-09-21
remove addresses endpoint

bennycio:
Changes:

- **Added** remove addresses endpoint
- **Added** func to ensure that user does not remove all of their addresses
- **Added** persist func to pull addresses from their addresses list

### DIFF
--- a/persist/collection.go
+++ b/persist/collection.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"strings"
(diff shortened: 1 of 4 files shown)


kaitoo1:
Remove address functionality looks good to me - I think we'll also want to consider removing the removed addresses's nfts from the user's collections.

Robinnnnn (approved):

kaitoo1 (approved):

PR #163, opened 2021-09-22
authed preflight no NFT check

bennycio:
Changes:

- **Added** jwtOptional middleware to preflight endpoint
- **Added** check for already authed during preflight to skip the NFT check


kaitoo1 (approved):
💪💪

PR #164, opened 2021-09-23
Support ledger-produced signature

kaitoo1:
Ledger-produced signatures have v = 0 or 1, meaning they fail the existing signature validation which expects v to be 27 or 28.
Manually making an exception to allow v to be 0 or 1 seems to be an accepted solution for now.
The issue is documented in [this Github Issue](https://github.com/MetaMask/metamask-extension/issues/10240)

More info: https://ethereum.stackexchange.com/questions/103307/cannot-verifiy-a-signature-produced-by-ledger-in-solidity-using-ecrecover

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -268,7 +268,12 @@ func authVerifySignature(pSignatureStr string,
 	if err != nil {
 		return false, err
 	}
-	if sig[64] != 27 && sig[64] != 28 {
+	v := sig[64]
+	// Ledger-produced signatures have v = 0 or 1
+	if v == 0 || v == 1 {
+		v = v + 27
+	}
+	if v != 27 && v != 28 {
 		return false, errors.New("invalid signature (V is not 27 or 28)")
 	}
 	sig[64] -= 27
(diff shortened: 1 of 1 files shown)


server/auth.go
```diff
@@ -268,7 +268,12 @@ func authVerifySignature(pSignatureStr string,
 	if err != nil {
 		return false, err
 	}
-	if sig[64] != 27 && sig[64] != 28 {
+	v := sig[64]
+	// Ledger-produced signatures have v = 0 or 1
+	if v == 0 || v == 1 {
+		v = v + 27
```
Robinnnnn, line 274: should be `sig[64] += 27`, to update the actual value at that index (instead of just the variable v)

server/auth.go
```diff
@@ -268,7 +268,12 @@ func authVerifySignature(pSignatureStr string,
...
+	}
+	if v != 27 && v != 28 {
 		return false, errors.New("invalid signature (V is not 27 or 28)")
 	}
 	sig[64] -= 27
```
Robinnnnn, line 279: lol weird that we subtract it later

kaitoo1: so it ends up being 0 or 1 anyway 🙃

server/auth.go
kaitoo1: good catch

PR #168, opened 2021-09-27
loosen cors for vercel

bennycio:
Changes:

- **Added** check in cors middleware for request origin starting with a vercel generated git link

### DIFF
--- a/server/middleware.go
+++ b/server/middleware.go
@@ -89,7 +89,7 @@ func handleCORS(runtimeConfig *runtime.Config) gin.HandlerFunc {
 		requestOrigin := c.Request.Header.Get("Origin")
 		allowedOrigins := strings.Split(runtimeConfig.AllowedOrigins, ",")
 
-		if util.Contains(allowedOrigins, requestOrigin) || strings.HasPrefix(requestOrigin, "https://gallery-git") {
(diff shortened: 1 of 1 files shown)


server/middleware.go
```diff
@@ -89,7 +89,7 @@ func handleCORS(runtimeConfig *runtime.Config) gin.HandlerFunc {
 		requestOrigin := c.Request.Header.Get("Origin")
 		allowedOrigins := strings.Split(runtimeConfig.AllowedOrigins, ",")
 
-		if util.Contains(allowedOrigins, requestOrigin) {
+		if util.Contains(allowedOrigins, requestOrigin) || strings.HasPrefix(requestOrigin, "https://gallery-git") {
```
Robinnnnn, line 92: Thinking about this more, a malicious user could theoretically set up a website on gallery-git.hacker.net. thoughts on this @bennycio @kaitoo1?

kaitoo1: I think we can do two things to prevent that:
- Only allow gallery-git origins on dev. We can use `env_variables` in `app.yaml` to configure this
- Check the suffix of the request origin in addition to the prefix, so we ensure that the origin ends in `vercel.app`. This could be a regex pattern match to check prefix and suffix at once. Basically `https://gallery-git-{ALPHANUMERIC+DASHES ALLOWED}-gallery-so.vercel.app`

That should allow something like the following to hit the backend while preventing gallery-git.hacker.net
https://gallery-git-kaito-video-image-url-gallery-so.vercel.app

Robinnnnn: agreed with above.

very small thing is this would prevent us from having a preview that points to production, but at that point we might as well just be testing locally

PR #171, opened 2021-09-29
many tests, many fixes

bennycio:
Changes:

- **Added** many tests for things like creating users, logging in, verifying signatures, verifying jwts, adding/removing addresses
- **Changed** various functions that were discovered to be malignant during the new tests such as `CollRemoveNFTsOfAddresses` that was not unsetting owner user ID and owner address for the NFTs being removed


bennycio:
Changed opensea pipeline to ensure that every NFT returned has its database ID. Does the front end use the IDs from the opensea response when updating collections or does it use the IDs from get unassigned? If the latter, we can not waste our time finding the ID for every NFT in opensea get.

kaitoo1:
M A S S I V E 👏

server/t__auth_test.go
```diff
@@ -0,0 +1,47 @@
+package server
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/mikeydub/go-gallery/util"
+	"github.com/stretchr/testify/assert"
+)
```
kaitoo1, line 10: are there any failure cases we can test for these? to ensure we're not getting false positives

server/t__routes_auth_test.go
```diff
@@ -27,6 +32,66 @@ func TestAuthPreflightUserExists_Success(t *testing.T) {
 	assert.True(output.UserExists)
 }
 
+func TestUserCreate_Success(t *testing.T) {
```
kaitoo1, line 35: Could we add failure cases for TestUserCreate and TestUserLogin ?

kaitoo1 (approved):
Changes look great! Thanks for adding tests, that's going to be super helpful in catching regressions. 
I think it would be great to add failure cases wherever possible as well, so that we're sure that the success cases aren't returning false positives, and incorrect inputs result in the expected errors.

PR #181, opened 2021-10-06
abstractions and specifications

bennycio:
Changes:

- **Added** database abstractions so that we are not bound to mongo
- **Removed** runtime package in exchange for a more idiomatic go approach that lets us see more specifically what dependencies any given function/endpoint has


memstore/redis.go
```diff
@@ -0,0 +1,81 @@
...
+	"context"
+	"errors"
+	"time"
+
+	"github.com/go-redis/redis/v8"
+)
+
+const (
+	// CollUnassignedRDB is a throttled cache for expensive queries finding unassigned NFTs
+	CollUnassignedRDB DB = iota
```
Robinnnnn, line 13: why is there a space between `DB DB` ?

persist/mongodb/nft.go
```diff
@@ -0,0 +1,275 @@
...
+	return mongo.Pipeline{
+		{{Key: "$match", Value: matchFilter}},
+		{{Key: "$lookup", Value: bson.M{
+			"from":         "history",
+			"localField":   "_id",
+			"foreignField": "nft_id",
+			"as":           "ownership_history",
+		}}},
+		{{Key: "$set", Value: bson.M{"ownership_history": bson.M{"$arrayElemAt": []interface{}{"$ownership_history", 0}}}}},
+		// {{Key: "$lookup", Value: bson.M{
```
Robinnnnn, line 262: saving this for later?

persist/nft.go
```diff
@@ -144,250 +130,17 @@ type UpdateNFTInfoInput struct {
...
-		// 			"$expr": bson.M{
-		// 				"$in": []interface{}{bson.M{"$first": "$addresses"}, "$$owners"},
-		// 			},
-		// 		}}},
-		// 	},
-		// 	"as": "owner_users",
-		// }}},
-	}
+// NFTRepository represents the interface for interacting with persisted NFTs
+type NFTRepository interface {
```
Robinnnnn, line 134: lovely

persist/mongodb/nft.go
bennycio: yes! this is from when we made it not fill out username for an NFT.

PR #191, opened 2021-11-08
set admin pass for admin bypass

bennycio:
Changes:

- **Changed** middleware to accept a different format for auth header (Authorization: [ADMIN PASS] [USER_ID_IMPERSONATING])
- **Changed** secrits to include an ADMIN_PASS env var


server/server.go
```diff
@@ -86,6 +86,7 @@ func setDefaults() {
 	viper.SetDefault("CONTRACT_ADDRESS", "0xe01569ca9b39e55bc7c0dfa09f05fa15cb4c7698")
 	viper.SetDefault("CONTRACT_INTERACTION_URL", "https://eth-mainnet.alchemyapi.io/v2/lZc9uHY6g2ak1jnEkrOkkopylNJXvE76")
 	viper.SetDefault("REQUIRE_NFTS", false)
+	viper.SetDefault("ADMIN_PASS", "TEST_ADMIN_PASS")
```
kaitoo1, line 89: is there any possibility that `ADMIN_PASS` wouldn't be set with the intended value, and the endpoints would be by-passable with the default value `TEST_ADMIN_PASS`?

PR #192, opened 2021-11-08
columns with token layout in updates and creation of collections

bennycio:
Changes:

- **Added** a token layout object that for the time being just holds a column field representing the number of columns in a collection
- **Added** a function to validate layouts ensuring the column constraints are met as well as defaults are set
- **Changed** update and create collection functions to optionally take in a layout object that will have its defaults set if not included in the request

### DIFF
--- a/server/collection_token.go
+++ b/server/collection_token.go
@@ -32,10 +32,11 @@ type collectionGetOutputtoken struct {
 }
 
 type collectionCreateInputToken struct {
-	GalleryID      persist.DBID   `json:"gallery_id" binding:"required"`
-	Nfts           []persist.DBID `json:"nfts" binding:"required"`
-	Name           string         `json:"name"`
-	CollectorsNote string         `json:"collectors_note"`
+	GalleryID      persist.DBID        `json:"gallery_id" binding:"required"`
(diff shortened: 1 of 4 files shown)


server/collection_token.go
```diff
@@ -232,9 +234,15 @@ func updateCollectionTokensToken(collectionsRepository persist.CollectionTokenRe
 		// ensure that there are no repeat NFTs
 		withNoRepeats := uniqueDBID(input.Nfts)
 
-		update := &persist.CollectionTokenUpdateNftsInput{Nfts: withNoRepeats}
+		layout, err := persist.ValidateLayout(input.Layout)
```
kaitoo1, line 237: Would the update collection endpoint still work if layout isn't included in the request? It looks like it's optional and we default to 3 columns in certain cases, but wasn't sure if `ValidateLayout` expects `input.Layout` to be defined

I think it would be great if this endpoint could continue to function as it does today without any frontend changes so that it can get merged and deployed independently. Maybe it does, but wanted to make sure 🙂

bennycio: It sure does thanks to no nil structs in go!

PR #201, opened 2021-12-02
remove 1 nft per collection constraint

bennycio:
Changes:

- **Removed** 1 nft per collection constraint

### DIFF
--- a/persist/mongodb/collection_token.go
+++ b/persist/mongodb/collection_token.go
@@ -52,13 +52,15 @@ func (c *CollectionTokenMongoRepository) Create(pCtx context.Context, pColl *per
 
 	if pColl.Nfts == nil {
 		pColl.Nfts = []persist.DBID{}
-	} else {
+	} /* else {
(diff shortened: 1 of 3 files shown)


persist/mongodb/collection_token.go
```diff
@@ -187,11 +189,12 @@ func (c *CollectionTokenMongoRepository) UpdateNFTsUnsafe(pCtx context.Context,
 	pUpdate *persist.CollectionTokenUpdateNftsInput,
 ) error {
 
-	if err := c.collectionsStorage.pullAll(pCtx, bson.M{}, "nfts", pUpdate.Nfts); err != nil {
-		if err != ErrDocumentNotFound {
-			return err
-		}
-	}
+	// this is to ensure that the NFTs are not being shared between collections
```
Robinnnnn, line 192: i think we should mark comments like this with "TODO" so we can easily grep / come back to clean them up

Robinnnnn (approved):

PR #203, opened 2021-12-03
cache galleries

bennycio:
Changes:

- **Added** caching for galleries. Now galleries are always returned first from the cache and optimistically updated every time an update is made to the gallery or a collection

Considerations:

- The key for cached galleries in the cache is the user ID. Not all functions that update information that would potentially update the result of a `getGalleriesByUserID` call take in a user ID. Those functions have no way of telling the cache that there is something that needs to be updated. Currently I have set an hour long TTL on the cache. Do you have any ideas of how we can ensure the cache stays up to date? We could use what a function does give us to find an owner user ID (such as using the `owner_address` of an NFT) but for functions like `BulkUpsert` there might be a lot of users to find. This still is the best solution I can think of right now.

TODO:
 
- [x] optimistically update unassigned cache
- [x] task queue for updating caches

### DIFF
--- a/server/t__routes_collection_test.go
+++ b/server/t__routes_collection_test.go
@@ -127,12 +127,6 @@ func TestGetUnassignedCollection_Success(t *testing.T) {
 func TestDeleteCollection_Success(t *testing.T) {
 	assert := setupTest(t)
 
-	// nfts := []*persist.NFTDB{
-	// 	{Description: "asd", CollectorsNote: "asd", OwnerAddress: tc.user1.address},
-	// 	{Description: "bbb", CollectorsNote: "bbb", OwnerAddress: tc.user1.address},
-	// 	{Description: "wowowowow", CollectorsNote: "wowowowow", OwnerAddress: tc.user1.address},
-	// }
-	// _, err := tc.repos.nftRepository.CreateBulk(context.Background(), nfts)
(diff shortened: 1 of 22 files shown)


server/t__routes_collection_test.go
```diff
@@ -127,6 +127,12 @@ func TestGetUnassignedCollection_Success(t *testing.T) {
 func TestDeleteCollection_Success(t *testing.T) {
 	assert := setupTest(t)
 
+	// nfts := []*persist.NFTDB{
+	// 	{Description: "asd", CollectorsNote: "asd", OwnerAddress: tc.user1.address},
+	// 	{Description: "bbb", CollectorsNote: "bbb", OwnerAddress: tc.user1.address},
+	// 	{Description: "wowowowow", CollectorsNote: "wowowowow", OwnerAddress: tc.user1.address},
+	// }
+	// _, err := tc.repos.nftRepository.CreateBulk(context.Background(), nfts)
```
Robinnnnn, line 135: did you mean to leave this commented?

Robinnnnn (approved):

server/t__routes_collection_test.go
bennycio: Thanks for pointing out robin!

PR #205, opened 2021-12-06
only use pointers where they are necessary

bennycio:
Changes:

- **Changed** most uses of pointers to not use pointers because pointers should only be used only when one of the following conditions is met:

1. Another function needs to update the value pointed to by the pointer and the calling function needs to reflect that update
2. The pointer is pointing to a very large struct that would be cheaper to copy a pointer than copy the entire struct
3. When emptiness or null-ness needs to be represented and can't be with the zero value of the value's type

At first I thought that because some of our structs can get quite large that we should be using pointer for most of them but after learning more about how pointers work, it will be cheaper to not use them because they will consistently go to another part of memory every time they are copied (heap memory) where as without pointers the copies stay on the stack. The go compiler also optimizes for copies and often ends up not copying whole structs anyway. Finally, our structs are mostly large because of large arrays or maps and maps and arrays are always reference values so when a struct is copied with an array field or map field the struct copies the pointer to the array or map and not every value stored in the array or map. This means most of our structs are actually quite small.

### DIFF
--- a/features/access.go
+++ b/features/access.go
@@ -24,7 +24,7 @@ type getUserFeaturesInput struct {
 }
 
 type getUserFeaturesOutput struct {
-	Features []*persist.FeatureFlag `json:"features"`
+	Features []persist.FeatureFlag `json:"features"`
 }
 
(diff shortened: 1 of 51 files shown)


Robinnnnn (approved):
thanks for laying everything out so clearly!

PR #207, opened 2021-12-10
services package

bennycio:
Changes:

- **Added** service package with code that pertains to the actual logic of retrieving and handling data for our application so that the top level packages just have to worry about the routing and handlers


kaitoo1:
great idea! LGTM
everything is so clean ✨

PR #210, opened 2021-12-10
add caching for user nfts

bennycio:
Changes:

- **Added** caching for getting nfts by user ID
- **Change** memory store task queue to ensure synchronous updates by key. This means that although many updates can be running asynchronously for different keys, no more than one update can be run at a time per key. This is to ensure that when updates come in for a given key, the most recent update will be the one that will be reflected in the cache.

### DIFF
--- a/service/persist/mongodb/nft.go
+++ b/service/persist/mongodb/nft.go
@@ -60,24 +60,24 @@ func (n *NFTMongoRepository) CreateBulk(pCtx context.Context, pNfts []persist.NF
 		return nil, err
 	}
 
-	// go func() {
-	// 	usersFound := map[persist.DBID]bool{}
-	// 	for _, v := range pNfts {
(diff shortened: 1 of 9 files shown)


service/persist/mongodb/nft.go
```diff
@@ -51,17 +59,73 @@ func (n *NFTMongoRepository) CreateBulk(pCtx context.Context, pNfts []persist.NF
 	if err != nil {
 		return nil, err
 	}
+
+	// go func() {
```
Robinnnnn, line 63: any reason we're keeping this?

Robinnnnn (approved):
🔥🔥🔥🔥

service/persist/mongodb/nft.go
bennycio: Just uncommented it! I commented it out to see if it broke the tests.

PR #217, opened 2021-12-17
start writing postgres persist package

bennycio:
Changes:

- **Added** package `service/persist/postgres`
- **Added** NFT postgres interaction code

TODO:

- [x] Collections
- [x] Users
- [x] Nonces
- [x] Auth
- [x] Users
- [x] Galleries
- [x] Tokens
- [x] Backups
- [x] Contracts
- [x] Features
- [x] Access
- [x] Membership
- [x] TESTS


docker/postgres/init.sql
```diff
@@ -0,0 +1,186 @@
+CREATE TABLE users (
+    ID varchar(32) PRIMARY KEY,
+    DELETED boolean NOT NULL DEFAULT false,
+    VERSION int,
+    LAST_UPDATED timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    CREATED_AT timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    USERNAME varchar(255),
+    USERNAME_IDEMPOTENT varchar(255),
+    ADDRESSES varchar(42) []
```
kaitoo1, line 9: We might want a larger limit for address length since we want to support other chains in the future - I know Solana addresses can be up to 44 characters.

Though maybe because to support multi-chain we'll need to make more schema changes anyway in the future, we could just update the limit when we need it. But just wanted to call it out here to make sure we did consider it

docker/postgres/init.sql
```diff
@@ -0,0 +1,186 @@
...
+    ID varchar(32) PRIMARY KEY,
+    DELETED boolean NOT NULL DEFAULT false,
+    VERSION int,
+    LAST_UPDATED timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    CREATED_AT timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    NAME varchar,
+    DESCRIPTION varchar,
+    COLLECTORS_NOTE varchar,
+    EXTERNAL_URL varchar,
+    CREATOR_ADDRESS varchar(255),
```
kaitoo1, line 34: any reason the `creator_address` limit is 255 and not 42 like `owner_address`?

scripts/initial_setup.sql
```diff
@@ -0,0 +1,170 @@
+CREATE TABLE users (
```
kaitoo1, line 1: when does this script get used? what's the difference between this and `init.sql`? It looks like init.sql imports data as well?

server/server.go
```diff
@@ -189,7 +209,15 @@ func newGCPStorageClient() *storage.Client {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(10)*time.Second)
 	defer cancel()
 	if viper.GetString("ENV") != "production" || viper.GetString("ENV") != "development" {
-		if _, err := os.Stat(viper.GetString("GOOGLE_APPLICATION_CREDENTIALS")); err != nil {
+
+		_, err := os.Stat(viper.GetString("GOOGLE_APPLICATION_CREDENTIALS"))
```
kaitoo1, line 213: `GOOGLE_APPLICATION_CREDENTIALS` could be a const to avoid bugs due to typos since we use it in several places

server/server.go
```diff
@@ -111,31 +118,44 @@ func setDefaults() {
...
+		collectionRepository:      postgres.NewCollectionRepository(db),
+		collectionTokenRepository: postgres.NewCollectionTokenRepository(db),
+		galleryRepository:         postgres.NewGalleryRepository(db, galleriesCache),
+		galleryTokenRepository:    postgres.NewGalleryTokenRepository(db, galleriesCacheToken),
+		historyRepository:         mongodb.NewHistoryRepository(mgoClient),
+		contractRepository:        postgres.NewContractRepository(db),
+		backupRepository:          postgres.NewBackupRepository(db),
+		membershipRepository:      postgres.NewMembershipRepository(db),
 	}
+	// galleryTokenRepo := mongodb.NewGalleryTokenRepository(mgoClient, galleriesCacheToken)
```
kaitoo1, line 141: do we want to keep the commented lines?

service/opensea/opensea.go
```diff
@@ -103,31 +103,31 @@ func PipelineAssetsForAcc(pCtx context.Context, pUserID persist.DBID, pOwnerWall
 
 	user, err := userRepo.GetByID(pCtx, pUserID)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to get user by id %s: %w", pUserID, err)
```
kaitoo1, line 106: formatted errors looking nice 👍

service/persist/mongodb/auth.go
```diff
@@ -15,37 +15,37 @@ const (
 	noncesCollName       = "nonces"
 )
 
-// LoginMongoRepository is a repository for storing login attempts in a MongoDB database
-type LoginMongoRepository struct {
+// LoginRepository is a repository for storing login attempts in a MongoDB database
```
kaitoo1, line 18: love the generalization here

kaitoo1:
Absolute monster of a PR Benny, amazing work!!!
I had a few questions that I left as comments, and I'll also start a thread in discord.

Also I imagine this changes the steps to get things running locally, would you be able to update the readme as well?

docker/postgres/init.sql
bennycio: oops! Thanks for pointing this out!

docker/postgres/init.sql
bennycio: very true! Will increase this limit!

scripts/initial_setup.sql
bennycio: This is used for GCP when the data can't be imported with the script.

server/server.go
bennycio: very true as well!

server/server.go
bennycio: No, we can remove them.

bennycio:
> Absolute monster of a PR Benny, amazing work!!! I had a few questions that I left as comments, and I'll also start a thread in discord.
> 
> Also I imagine this changes the steps to get things running locally, would you be able to update the readme as well?

Surprisingly enough, the app is still started exactly the same! Docker compose is super duper helpful!

service/persist/postgres/collection_token.go
```diff
@@ -0,0 +1,430 @@
...
+		return "", err
+	}
+	return id, nil
+}
+
+// GetByUserID returns all collections owned by a user
+func (c *CollectionTokenRepository) GetByUserID(pCtx context.Context, pUserID persist.DBID, pShowHidden bool) ([]persist.CollectionToken, error) {
+	var stmt *sql.Stmt
+	if pShowHidden {
+		stmt = c.getByUserIDAuthStmt
```
radazen, line 134: From the perspective of a newcomer to the codebase, I'd appreciate more commenting around the hidden/auth/unsafe types of methods. I understand that this particular function is fairly low-level, and maybe there's more explanation of the higher-level functions that end up calling these, but as someone reading through the code, I wonder stuff like:

- Is it the caller's job to ensure that the user is authenticated before using `getByUserIDAuthStmt` (which itself is used indirectly by `pShowHidden`)?
- What's the relationship between a user being authenticated and a collection being hidden? Is there something privileged about seeing hidden collections? Is there sensitive data involved?

If I'm understanding things correctly, is the auth stuff shorthand for "we'll give you some extra information if you're the user who owns this"? If so, it may be clearer to phrase things in terms of ownership instead of authentication.

radazen (commented):
Looking good! I left some comments and some questions. I don't have enough context (yet) to understand everything that's going on, but it all looks very reasonable to me! This is a pretty amazing thing to have put together so quickly. 🔥

docker/postgres/init.sql
```diff
@@ -0,0 +1,186 @@
+CREATE TABLE users (
+    ID varchar(32) PRIMARY KEY,
+    DELETED boolean NOT NULL DEFAULT false,
+    VERSION int,
+    LAST_UPDATED timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    CREATED_AT timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    USERNAME varchar(255),
+    USERNAME_IDEMPOTENT varchar(255),
```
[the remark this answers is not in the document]
bennycio: for a case insensitive username that will always be lowercase

docker/postgres/init.sql
```diff
@@ -0,0 +1,186 @@
+CREATE TABLE users (
+    ID varchar(32) PRIMARY KEY,
+    DELETED boolean NOT NULL DEFAULT false,
+    VERSION int,
+    LAST_UPDATED timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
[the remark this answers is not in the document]
bennycio: Awesome! Will change this, thank you!

docker/postgres/init.sql
```diff
@@ -0,0 +1,186 @@
...
+CREATE TABLE membership (
+    ID varchar(32) PRIMARY KEY,
+    DELETED boolean NOT NULL DEFAULT false,
+    VERSION int,
+    CREATED_AT timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    LAST_UPDATED timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    TOKEN_ID varchar,
+    NAME varchar,
+    ASSET_URL varchar,
+    OWNERS jsonb []
```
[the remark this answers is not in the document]
bennycio: UserID, Address, Username, and Preview NFTs as a JSON object are stored as owners.

indexer/core.go
```diff
@@ -102,7 +111,20 @@ func newIPFSShell() *shell.Shell {
...
+
+	switch viper.GetString("DATABASE") {
+	case "mongodb":
+		return mongodb.NewTokenRepository(mgoClient, nil), mongodb.NewContractRepository(mgoClient), mongodb.NewUserRepository(mgoClient)
+	case "postgres":
+		return postgres.NewTokenRepository(pgClient), postgres.NewContractRepository(pgClient), postgres.NewUserRepository(pgClient)
+	case "multi":
+		mongoToken, mongoContract := mongodb.NewTokenRepository(mgoClient, nil), mongodb.NewContractRepository(mgoClient)
+		pgToken, pgContract, pgUser := postgres.NewTokenRepository(pgClient), postgres.NewContractRepository(pgClient), postgres.NewUserRepository(pgClient)
+		return multi.NewTokenRepository(pgToken, mongoToken), multi.NewContractRepository(pgContract, mongoContract), pgUser
```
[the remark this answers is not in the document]
bennycio: The first parameter in creating a multi repository is the "Read and Write repo" which is where we will be reading from the database. Every other parameter is just being written to. Because we never write to the user's collection in the indexer, just read, I didn't think it makes sense to define a multi repo for it and put mongo as a write repo when it will never be used.

scripts/indexer_only_init.sql
```diff
@@ -0,0 +1,55 @@
...
+    CONTRACT_ADDRESS varchar(255),
+    COLLECTORS_NOTE varchar,
+    MEDIA jsonb,
+    CHAIN varchar,
+    OWNER_ADDRESS varchar(255),
+    TOKEN_URI varchar,
+    TOKEN_TYPE varchar,
+    TOKEN_ID varchar,
+    QUANTITY varchar,
+    OWNERSHIP_HISTORY jsonb [],
```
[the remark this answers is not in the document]
bennycio: Exactly, I think the key for me here is that I don't see us querying this data, just looking at it when it is returned with a token.

service/persist/persist.go
```diff
@@ -16,6 +18,15 @@ type CreationTime time.Time
...
 type LastUpdatedTime time.Time
 
+// NullString represents a string that may be null in the DB
+type NullString string
+
+// NullInt64 represents an int64 that may be null in the DB
+type NullInt64 int64
+
+// NullBool represents a bool that may be null in the DB
+type NullBool bool
```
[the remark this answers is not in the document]
bennycio: Any SQL database will panic if it tries to scan a null item into a go primitive type. Having this type allows us to control how the scanning works, and although we don't need to do this for mongo, it could also be beneficial in the future if we wanted to marshal and unmarshal an int64 a specific way.

docker/postgres/init.sql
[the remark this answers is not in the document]
radazen: Gotcha. What's the benefit of storing it in that form instead of deriving a lowercase name from the USERNAME field? Seems like you're essentially caching a lowercase form of the name. Is it for performance, convenience, or something else?

service/persist/multi/contract.go
```diff
@@ -0,0 +1,61 @@
...
+	errChan := make(chan error)
+	for _, repo := range c.WriteRepos {
+		go func(repo persist.ContractRepository) {
+			errChan <- repo.UpsertByAddress(pCtx, pAddress, pContract)
+		}(repo)
+	}
+	go func() {
+		errChan <- c.ReadRepo.UpsertByAddress(pCtx, pAddress, pContract)
+	}()
+	for i := 0; i < len(c.WriteRepos)+1; i++ {
```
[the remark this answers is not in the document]
bennycio: The +1 is because errChan is going to take in an err (or nil err) from every write repo (`len(c.WriteRepos`) and also from the single read repo (`c.ReadRepo`). You are write it would just be blocked forever and eventually go would catch it and panic.

service/persist/postgres/collection_token.go
bennycio: Very true! Can definitely do some renaming and commenting to make this more clear. Just to clear up in the meantime:

- Yes, auth represents the user who owns whatever it is
- For unsafe the function won't take in a user ID and won't check that whatever we are updating is owned by a given user.

service/persist/multi/contract.go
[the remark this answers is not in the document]
radazen: Aah! I see, the extra one is from `c.ReadRepo`. Didn't catch that when I was looking at it last night.

docker/postgres/init.sql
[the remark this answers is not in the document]
bennycio: The benefit was so that we could use that field to search and the other username field was just for storing what the user intended their username to be. I have heard that SQL has ways of searching strings with case insensitivity but mongo doesn't have that feature so that's why we have had this field. I will look into that SQL functionality and see if we can leave the field for backwards compatibility with mongo but not use it when we are doing SQL.

radazen:
Ah! That makes all of the sense. Thanks for the explanation!

kaitoo1:
broadly LGTM - I'll leave the final approval to @radazen since he's taken a much deeper look

service/persist/postgres/collection_nft.go
```diff
@@ -108,7 +108,7 @@ func NewCollectionRepository(db *sql.DB) *CollectionRepository {
...
+	return &CollectionRepository{db: db, createStmt: createStmt, getByUserIDAuthStmt: getByUserIDOwnerStmt, getByUserIDStmt: getByUserIDStmt, getByIDAuthStmt: getByIDOwnerStmt, getByIDStmt: getByIDStmt, updateInfoStmt: updateInfoStmt, updateHiddenStmt: updateHiddenStmt, updateNFTsStmt: updateNFTsStmt, nftsToRemoveStmt: nftsToRemoveStmt, deleteNFTsStmt: deleteNFTsStmt, removeNFTFromCollectionsStmt: removeNFTFromCollectionsStmt, getNFTsForAddressStmt: getNFTsForAddressStmt, deleteCollectionStmt: deleteCollectionStmt, getUserAddressesStmt: getUserAddressesStmt, getUnassignedNFTsStmt: getUnassignedNFTsStmt}
```
radazen, line 111: Probably want to update the `getByUserIDAuthStmt:` part of this too. Thanks for renaming!

service/persist/postgres/collection_token.go
```diff
@@ -113,7 +113,7 @@ func NewCollectionTokenRepository(db *sql.DB) *CollectionTokenRepository {
...
+	return &CollectionTokenRepository{db: db, createStmt: createStmt, getByUserIDAuthStmt: getByUserIDOwnerStmt, getByUserIDStmt: getByUserIDStmt, getByIDAuthStmt: getByIDOwnerStmt, getByIDStmt: getByIDStmt, updateInfoStmt: updateInfoStmt, updateInfoUnsafeStmt: updateInfoUnsafeStmt, updateHiddenStmt: updateHiddenStmt, updateHiddenUnsafeStmt: updateHiddenUnsafeStmt, updateNFTsStmt: updateNFTsStmt, updateNFTsUnsafeStmt: updateNFTsUnsafeStmt, nftsToRemoveStmt: nftsToRemoveStmt, deleteNFTsStmt: deleteNFTsStmt, removeNFTFromCollectionsStmt: removeNFTFromCollectionsStmt, getNFTsForAddressStmt: getNFTsForAddressStmt, deleteCollectionStmt: deleteCollectionStmt, getUserAddressesStmt: getUserAddressesStmt, getUnassignedNFTsStmt: getUnassignedNFTsStmt}
```
radazen, line 116: Same as above; probably want to update the getByUserIDAuthStmt: part of this too

radazen (approved):
Two small comments about renaming stuff, but otherwise looks A-OK to me! 🔥

PR #218, opened 2021-12-23
batch transfer fix, medias, logs errors, timings

bennycio:
Changes:

- **Changed** batch transfer event processing to use block number and therefore be included throughout the indexer process
- **Changed** batch transfer and single transfer event processing to use `go-ethereum`'s `abi` package for parsing the data field of a log so we don't have to manually parse it
- **Added** media retrieval for NFTs that are owned by gallery users
- **Added** error checking for logs

TODO:

- [x] handle the possibility of a `balanceOf` error for ERC-1155 contracts
- [x] check the error returned for getting token URI, get metadata, get media, and get contract metadata to determine if it is worth retrying on the fly later on or if the information we are trying to retrieve will never be available (for example, an API is permanently offline and the NFT metadata won't ever be accessible)
- [x] indexer has an endpoint that will be called when a user signs up so that the indexer can retrieve all of the media content for the NFTs they own


indexer/indexer.go
```diff
@@ -264,69 +277,110 @@ func (i *Indexer) processLogs(transfersChan chan<- []*transfer, startingBlock pe
...
-				tokenType:       persist.TokenTypeERC721,
+			// t := time.Now()
+			// erc20, err := contracts.NewIERC20Caller(pLog.Address, ethClient)
+			// if err == nil {
+			// 	_, err := erc20.Allowance(&bind.CallOpts{}, common.HexToAddress(pLog.Topics[1].Hex()), common.HexToAddress(pLog.Topics[2].Hex()))
+			// 	if err == nil {
+			// 		continue
+			// 	}
+			// }
+			// logrus.Infof("Figured out if contract was ERC20 in %s", time.Since(t))
```
bennycio, line 304: This was to test if a contract was an ERC-20 token that was trying to sneak its way through as a ERC-721 token because their event signatures look the same sometimes. Commenting this step out makes the logsToTransfers step go from about 1 minute to a few milliseconds. Want to explore any alternate solutions to just calling an ERC-20 specific function on the contract to test if it is ERC-20.

kaitoo1 (approved):

PR #220, opened 2021-12-31
use opensea to validate nfts, have signup and address add call indexer and pubsub

bennycio:
Changes:

- **Added** validate nfts endpoint to indexer to ensure that opensea NFTs match NFTs stored in database
- **Added** async calls to indexer and pubsub to create user and add addresses functions


radazen (approved):
LGTM! Only had one comment, and I deleted it after we discussed `pVar` naming conventions elsewhere. 😉

PR #226, opened 2022-01-13
increase opensea limit to 20k

bennycio:
Changes:

- **Added** check to start reversing results after opensea result 10k so we can receive a max of 20k nfts


service/opensea/opensea.go
```diff
@@ -255,22 +255,28 @@ func fetchAssetsForWallets(pCtx context.Context, pWalletAddresses []persist.Addr
...
 	}
 
-	qsLst := []string{}
-	for k, v := range qsArgsMap {
-		qsLst = append(qsLst, fmt.Sprintf("%s=%s", k, v))
+	dir := "desc"
+	offset := pOffset
+	if pOffset > 10000 {
+		dir = "asc"
+		offset = pOffset - 10000
```
radazen, line 274: I think this will actually end up missing some NFTs! 10000 is the last valid offset, so if we get here we'll have pulled in 10050 NFTs and will now have an offset of 10050, which means we'll subtract 10k and start our ascending queries at 50 instead of 0.

PR #227, opened 2022-01-14
validate and update nfts if not accounted for

bennycio:
Changes:

- **Added** processing for unaccounted for NFTs from the check against opensea

### DIFF
--- a/indexer/server.go
+++ b/indexer/server.go
@@ -8,9 +8,11 @@ import (
 	"time"
 
 	"cloud.google.com/go/storage"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/gin-gonic/gin"
 	shell "github.com/ipfs/go-ipfs-api"
+	"github.com/mikeydub/go-gallery/contracts"
 	"github.com/mikeydub/go-gallery/service/media"
 	"github.com/mikeydub/go-gallery/service/opensea"
 	"github.com/mikeydub/go-gallery/service/persist"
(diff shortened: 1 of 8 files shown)


service/persist/token.go
```diff
@@ -353,7 +354,7 @@ func (uri TokenURI) Type() URIType {
 }
 
 func (id TokenID) String() string {
-	return strings.ToLower(util.RemoveLeftPaddedZeros(string(id)))
+	return strings.ToLower(id.BigInt().Text(16))
```
radazen, line 357: For my own understanding: given that the tokenID itself is already a string, what's the difference between that string and the string we return from this `String()` function? String -> BigInt -> String seems kind of indirect, but it'd probably make more sense if I understood the difference between the starting and ending strings!

service/persist/token.go
```diff
@@ -371,13 +372,39 @@ func (id *TokenID) Scan(src interface{}) error {
 	return nil
 }
 
+// DetectBase returns the base of the token URI
+func (id TokenID) DetectBase() int {
```
radazen, line 376: This is an interesting one! Why do we end up needing to detect the base of a token ID? It's worth noting that it's unlikely but possible for a hex string not to have any letters in it. The longer the string, the less likely it is, but it's still possible (unless token IDs happen to adhere to a spec that always includes some bytes that translate into A-F).

Ideally we'd know the base of the token at the time we first encounter it, and then store the token either in "0x" format as appropriate, or make the token a struct with a "base" field so we don't have to guess later.

radazen:
Looks good! 🔥  I left some questions/comments, but nothing major!

service/persist/token.go
radazen: Follow-up thought: the best approach is probably for our own TokenID type to be consistently in a single base. E.g. our TokenID is always base 16, and then when we parse an ID and create a new TokenID type, we _hopefully_ know the base of the ID at the point where we're receiving it from somewhere, but if we really don't even know the base at that point, then we try our best guess with auto-detection. But the idea would be that, once we've got it as a TokenID type in our own codebase, we're working with a consistent base and don't need to guess every time we're using it.

util/random.go
```diff
@@ -5,6 +5,7 @@ import (
 )
 
 const alphanumeric = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+const hex = "0123456789abcdefABCDEF"
```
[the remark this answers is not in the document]
bennycio: Ethereum addresses are often passed in mixed case so I thought it might be best to include this as a part of our tests. I will document that for sure, thank you!

indexer/server.go
```diff
@@ -231,6 +241,7 @@ func validateUsersNFTs(tokenRepository persist.TokenRepository, userRepository p
 			}
 			if !af {
 				unaccountedFor[asset.Contract.ContractAddress.String()+" -- "+asset.TokenID.String()] = true
+				allUnaccountedForAssets = append(allUnaccountedForAssets, asset)
```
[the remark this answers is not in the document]
bennycio: Very true! Will make some changes to this part.

indexer/server.go
```diff
@@ -256,3 +276,91 @@ func validateUsersNFTs(tokenRepository persist.TokenRepository, userRepository p
 
 	}
 }
+
+func processUnaccountedForNFTs(ctx context.Context, assets []opensea.Asset, tokenRepository persist.TokenRepository, contractRepository persist.ContractRepository, userRepository persist.UserRepository, ethcl *ethclient.Client, ipfsClient *shell.Shell, stg *storage.Client) error {
+	store := map[persist.TokenIdentifiers]*persist.Token{}
+	allTokens := make([]persist.Token, 0, len(assets))
+	ctrcts := make([]persist.Contract, 0, len(assets))
```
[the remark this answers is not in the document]
bennycio: I wish we could call it contracts but unfortunately one of our packages is called contracts. I can change it to `cntracts` to make it a little more clear.

service/persist/token.go
bennycio: This is almost a way of normalizing the token ID because it will detect if the original string is base 10 or base 16 and convert it to hex with no zero padding.

service/persist/token.go
bennycio: Definitely. Right now the only time this makes a big difference is with opensea because opensea gives us token IDs as base 10. I can make that more explicit though by creating a new type in the opensea package for opensea token ID and converting it to our token ID so we can assume that it will always be hex.  Then I can remove the detect base stuff and the string -> Big -> string normalization I mentioned above. Thanks for pointing this out!

PR #228, opened 2022-01-14
Minor style clean-ups

radazen:
Figured I'd try this as a PR! Just a few suggestions for removing redundant use of literals and replacing them with variables to improve maintainability. Also added a few comments to help with exploring Postgres parameter limits in the future.

Oh, and I added .idea to .gitignore so IntelliJ will stop trying to commit every configuration change I make.

### DIFF
--- a/indexer/indexer.go
+++ b/indexer/indexer.go
@@ -486,7 +486,7 @@ func processTransfers(i *Indexer, transfers []transfersAtBlock, uris chan<- toke
 							return
 						}
 						var fromBalance, toBalance *big.Int
-						if from.String() != "0x0000000000000000000000000000000000000000" {
+						if from.String() != persist.ZeroAddress.String() {
 							fromBalance, err = ierc1155.BalanceOf(&bind.CallOpts{Context: ctx}, from.Address(), tokenID.BigInt())
 							if err != nil {
 								ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
@@ -508,7 +508,7 @@ func processTransfers(i *Indexer, transfers []transfersAtBlock, uris chan<- toke
 								return
 							}
 						}
-						if to.String() != "0x0000000000000000000000000000000000000000" {
+						if to.String() != persist.ZeroAddress.String() {
 							toBalance, err = ierc1155.BalanceOf(&bind.CallOpts{Context: ctx}, to.Address(), tokenID.BigInt())
 							if err != nil {
 								ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
--- a/service/persist/postgres/nft.go
+++ b/service/persist/postgres/nft.go
@@ -261,10 +261,15 @@ func (n *NFTRepository) BulkUpsert(pCtx context.Context, pUserID persist.DBID, p
 
 	resultIDs := make([]persist.DBID, len(pNFTs))
 
-	// Postgres can't upsert more than 3400 tokens at a time
-	if len(pNFTs) > (65535/21 - 1000) {
-		next := pNFTs[(65535/21 - 1000):]
-		current := pNFTs[:(65535/21 - 1000)]
+	// Postgres only allows 65535 parameters at a time.
+	// TODO: Consider trying this implementation at some point instead of chunking:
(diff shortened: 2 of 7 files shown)


service/persist/postgres/nft.go
```diff
@@ -261,10 +261,15 @@ func (n *NFTRepository) BulkUpsert(pCtx context.Context, pUserID persist.DBID, p
...
 
-	// Postgres can't upsert more than 3400 tokens at a time
-	if len(pNFTs) > (65535/21 - 1000) {
-		next := pNFTs[(65535/21 - 1000):]
-		current := pNFTs[:(65535/21 - 1000)]
+	// Postgres only allows 65535 parameters at a time.
+	// TODO: Consider trying this implementation at some point instead of chunking:
+	//       https://klotzandrew.com/blog/postgres-passing-65535-parameter-limit
+	paramsPerRow := 21
+	rowsPerQuery := 65535 / paramsPerRow
```
radazen, line 268: @bennycio I know we discussed adding the ` - 1000` part here and I was in favor of it, but when I was trying to add a comment explaining why it's there, I couldn't come up with a good reason beyond "because of a gut feeling that it'd be nice to have some wiggle room." A comment like that is a code smell (imho), and we know that the math should work out here, so in the end I don't think we need the extra breathing room.

service/persist/postgres/token.go
```diff
@@ -303,19 +303,24 @@ func (t *TokenRepository) BulkUpsert(pCtx context.Context, pTokens []persist.Tok
 		}
 	}
 
-	// Postgres can't upsert more than 3400 tokens at a time
-	if len(allTokens) > (65535/19 - 1000) {
-		next := allTokens[(65535/19 - 1000):]
-		current := allTokens[:(65535/19 - 1000)]
+	// Postgres only allows 65535 parameters at a time.
+	// TODO: Consider trying this implementation at some point instead of chunking:
```
radazen, line 307: Ran into this while googling for the limit. Sounds like a technique worth exploring at some point!

service/persist/token.go
```diff
@@ -98,6 +98,9 @@ const (
 // InvalidTokenURI represents an invalid token URI
 const InvalidTokenURI TokenURI = "INVALID"
 
+// ZeroAddress is the all-zero Ethereum address
+const ZeroAddress Address = "0x0000000000000000000000000000000000000000"
```
radazen, line 102: Not sure if this is the best place for the `ZeroAddress` const, but it looks like this is where we keep other similar constants.

indexer/indexer.go
```diff
@@ -486,7 +486,7 @@ func processTransfers(i *Indexer, transfers []transfersAtBlock, uris chan<- toke
 							return
 						}
 						var fromBalance, toBalance *big.Int
-						if from.String() != "0x0000000000000000000000000000000000000000" {
+						if from.String() != persist.ZeroAddress.String() {
```
radazen, line 489: I'd be tempted to remove the `.String()` from both of these, but I'm not sure if the lowercasing function of `.String()` is something we actually depend on. Do we ever expect addresses with an uppercase "X" in the "0X0000" format? Or is it universally accepted that, even when there are uppercase hexadecimal letters in the address, the leading `x` is lowercase?

service/persist/postgres/nft.go
bennycio: Sounds good to me! That implementation looks great with the unnest, in their example they used a typed int array, I wonder if it could work generically for our arrays of interfaces. I'll definitely look into this.

service/persist/postgres/token.go
bennycio: Oops, commented on this in the conversation above.

service/persist/token.go
bennycio: I think that is a marvelous place for it!

indexer/indexer.go
bennycio: True, especially in this case there would be no way the case would matter. I am in favor of removing the `.String()` on both sides. It is always a lower case `x` in `0x`.

bennycio (approved):
Looks fantastic! Thanks for these changes!

PR #231, opened 2022-01-20
allowlist from multiple addresses

bennycio:
Changes:

- **Changed** env variable `CONTRACT_ADDRESS` to `CONTRACT_ADDRESSES` that has a simple format as follows:
```
{contractAddress}={tokenID,otherTokenID,moreAllowlistedTokenIDs}|{anotherContract}={moreCommaSeparatedIDs}
```
the backend will parse this format and use it where needed to allowlist users.


server/handler.go
```diff
@@ -53,21 +54,21 @@ func authHandlersInitNFT(parent *gin.RouterGroup, repos *repositories, ethClient
 	// AUTH
 	authGroup.GET("/get_preflight", middleware.JWTOptional(), getAuthPreflight(repos.userRepository, repos.nonceRepository, ethClient))
 	authGroup.GET("/jwt_valid", middleware.JWTOptional(), auth.ValidateJWT())
-	authGroup.GET("/is_member", middleware.JWTOptional(), hasNFTs(repos.userRepository, ethClient, auth.RequiredNFTs))
+	authGroup.GET("/is_member", middleware.JWTOptional(), hasNFTs(repos.userRepository, ethClient, membership.PremiumCards, membership.MembershipTierIDs))
```
kaitoo1, line 57: I forget what this endpoint is for.. I see we're only checking against Premium Cards here but I don't think we use this endpoint currently so it might not matter

kaitoo1 (approved):

PR #238, opened 2022-01-25
Ignore unset local service key

Robinnnnn:
if service key is unset locally, app will still run instead of crashing

image

### DIFF
--- a/server/handler.go
+++ b/server/handler.go
@@ -17,7 +17,9 @@ func handlersInit(router *gin.Engine, repos *repositories, ethClient *ethclient.
 	apiGroupV2 := router.Group("/glry/v2")
 
 	nftHandlersInit(apiGroupV1, repos, ethClient, psub)
-	tokenHandlersInit(apiGroupV2, repos, ethClient, ipfsClient, psub, storageClient)
+	if storageClient != nil {
+		tokenHandlersInit(apiGroupV2, repos, ethClient, ipfsClient, psub, storageClient)
+	}
 
 	return router
 }
--- a/server/handler.go
+++ b/server/handler.go
@@ -1,7 +1,6 @@
 package server
 
 import (
-	"cloud.google.com/go/storage"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/gin-gonic/gin"
 	shell "github.com/ipfs/go-ipfs-api"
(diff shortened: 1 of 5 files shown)


server/server.go
```diff
@@ -228,35 +228,30 @@ func newGCPPubSub() pubsub.PubSub {
 func newGCPStorageClient() *storage.Client {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(10)*time.Second)
 	defer cancel()
-	if viper.GetString("ENV") != "production" || viper.GetString("ENV") != "development" {
```
Robinnnnn, line 231: main change 1 was getting rid of this line, since it always evaluates to true

server/server.go
```diff
@@ -228,35 +228,30 @@ func newGCPPubSub() pubsub.PubSub {
...
+	if err != nil {
+		logrus.Errorf("err getting app creds: %s", err)
+		client, err := storage.NewClient(ctx, option.WithCredentialsJSON([]byte(viper.GetString("GCLOUD_SERVICE_KEY"))))
 		if err != nil {
-			logrus.Errorf("err getting app creds: %s", err)
-			client, err := storage.NewClient(ctx, option.WithCredentialsJSON([]byte(viper.GetString("GCLOUD_SERVICE_KEY"))))
-			if err != nil {
-				panic(err)
+			if viper.GetString("ENV") != "production" && viper.GetString("ENV") != "development" {
+				logrus.Info("not using gcp storage client in local environment")
```
Robinnnnn, line 245: main change 2

server/server.go
```diff
@@ -228,35 +228,30 @@ func newGCPPubSub() pubsub.PubSub {
...
-			return client
-		}
-		client, err := storage.NewClient(ctx, option.WithCredentialsFile(appCredentials))
-		if err != nil {
 			panic(err)
 		}
 		return client
 	}
-	client, err := storage.NewClient(ctx)
+	client, err := storage.NewClient(ctx, option.WithCredentialsFile(appCredentials))
```
Robinnnnn, line 252: line 250 was just moved down here, git thinks it's an addition

bennycio, line 252: I think maybe we could move some of the logic around to allow this to also continue working in the app engine environment. Right now if we always use the credentials file then this might not work in the app engine environment. Maybe we could do something like this:

```
 if env != local {
  client, err := storage.NewClient(ctx)
  ...
  return client
}
// now we can assume this is local env
appCredentials := viper.GetString("GOOGLE_APPLICATION_CREDENTIALS")
... // logic to check if credentials file exists somewhere
if credentialsFile != nil {
  client, err := storage.NewClient(ctx, option.WithCredentialsFile(appCredentials))
  ...
  return client
}
// no credentials file exists locally, lets just return nil instead of panicking
return nil

```

bennycio (changes requested):

server/handler.go
```diff
@@ -17,7 +17,9 @@ func handlersInit(router *gin.Engine, repos *repositories, ethClient *ethclient.
 	apiGroupV2 := router.Group("/glry/v2")
 
 	nftHandlersInit(apiGroupV1, repos, ethClient, psub)
-	tokenHandlersInit(apiGroupV2, repos, ethClient, ipfsClient, psub, storageClient)
+	if storageClient != nil {
```
jarrel-b, line 20: Just realized that this would prevent me from testing v2 routes. Would it make to have some sort of dummy client?

PR #240, opened 2022-01-26
add endpoint to get current user and change logout to POST

kaitoo1:
## Context:
On the front end we are removing all JWT related code. This means that we are no longer storing the jwt in local storage which we used to quickly determine the user's 'authenticated' state upon page load.

This PR adds an endpoint that returns the currently authenticated user if a valid cookie is included in the request, which will allow the front end to continue to determine the user's logged in state.

## Changes:
- Added a new endpoint `users/get/current` to retrieve the currently authenticated user
  - If a user is found for the auth cookie, return the user
  - If there is no auth cookie or the provided auth cookie is invalid, return a 204 No Content
  - If the auth cookie is valid but no corresponding user is found, return a 500
- Added related tests for the endpoint
  - success: valid cookie return user
  - fail: no cookie returns 204
  - fail: invalid cookie returns 204 

I also changed the following to support related front end changes
- Changed /logout from GET to POST

### DIFF
--- a/server/user_token.go
+++ b/server/user_token.go
@@ -72,6 +72,34 @@ func getUser(userRepository persist.UserRepository) gin.HandlerFunc {
 	}
 }
 
+func getCurrentUser(userRepository persist.UserRepository) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		authed := c.GetBool(auth.AuthContextKey)
+		if !authed {
+			util.ErrResponse(c, http.StatusUnauthorized, errors.New("not authorized"))
+			return
+		}
+		userID := auth.GetUserIDFromCtx(c)
+
+		output, err := user.GetUser(
+			c,
+			user.GetUserInput{UserID: userID},
+			userRepository,
+		)
+		if err != nil {
+			status := http.StatusInternalServerError
+			switch err.(type) {
+			case persist.ErrUserNotFoundByAddress, persist.ErrUserNotFoundByID, persist.ErrUserNotFoundByUsername:
+				status = http.StatusNotFound
+			}
+			util.ErrResponse(c, status, err)
+			return
+		}
+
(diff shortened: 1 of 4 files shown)


service/auth/auth.go
```diff
@@ -54,7 +54,7 @@ var errAddressSignatureMismatch = errors.New("address does not match signature")
 var ErrNonceMismatch = errors.New("incorrect nonce input")
 
 // ErrInvalidJWT is returned when the JWT is invalid
-var ErrInvalidJWT = errors.New("invalid JWT")
+var ErrInvalidJWT = errors.New("invalid or expired auth token")
```
kaitoo1, line 57: I just updated this to be  a bit more user friendly

Robinnnnn (approved):

server/user_token.go
```diff
@@ -72,6 +71,34 @@ func getUser(userRepository persist.UserRepository) gin.HandlerFunc {
 	}
 }
 
+func getCurrentUser(userRepository persist.UserRepository) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		authed := c.GetBool(auth.AuthContextKey)
+		if !authed {
+			c.JSON(http.StatusNoContent, util.SuccessResponse{Success: false})
+			return
+		}
```
bennycio, line 80: This is unnecessary if you use `middleware.AuthRequired()` as a middleware for this endpoint instead of `middleware.AuthOptional()`. Unless you need the response status to be StatusNoContent?

bennycio (approved):

server/user_token.go
Robinnnnn: in this case we do need it to be StatusNoContent to handle a special case on the frontend!

kaitoo1: yep what Robin said!

PR #241, opened 2022-01-26
admin backend start

bennycio:
Changes:

- **Added** new backend service for admin
- **Added** some endpoints for getting and deleting a user and executing raw sql stmts

TODO:

- [ ] Spec out everything we want an endpoint for
- [ ] Some ideas I've already heard that we probably want an endpoint for: Creating a user, getting galleries, getting NFTs, getting collections, updating galleries and collections, deleting specific collections, refreshing opensea

Considerations:

- Right now there is a raw exec endpoint that can exec write statements for the DB as well as an option for the raw query endpoint that can allow queries to write. I am in favor of not allowing any user inputted sql statements be able to write (even if we can guarantee that only we can access this API). I think if we are doing any sort of manual updates there should be an explicit endpoint for it with inputs and outputs. What do you all think?


Robinnnnn:
> I am in favor of not allowing any user inputted sql statements be able to write (even if we can guarantee that only we can access this API). I think if we are doing any sort of manual updates there should be an explicit endpoint for it with inputs and outputs

agreed with this!

in terms of an endpoint that would be useful, i think **merging accounts** would be a big one. so far i've had to do many of the same manual migration by hand, which goes something like:
1. user has 2 separate accounts on gallery they want to merge
2. i delete their secondary user
3. i delete their galleries attached to said user
4. i add the deleted user's address to the main one

^ in fact, this merge functionality is something we want to allow directly from the main frontend too, so that any user can click on a button for themselves (as long as they can prove that they own both wallets)

PR #245, opened 2022-01-28
merge users endpoint

bennycio:
Changes:

- **Added** an authenticated endpoint to merge users
- **Added** 2 success case tests for this endpoint


service/persist/postgres/user.go
```diff
@@ -247,3 +273,76 @@ func (u *UserRepository) RemoveAddresses(pCtx context.Context, pID persist.DBID,
...
+		return err
+	}
+	defer tx.Rollback()
+
+	if _, err := tx.StmtContext(pCtx, u.addAddressStmt).ExecContext(pCtx, user.ID, pq.Array(secondUser.Addresses)); err != nil {
+		return err
+	}
+
+	deleteGalleryStmt := tx.StmtContext(pCtx, u.deleteGalleryStmt)
+	mergedCollections := make([]persist.DBID, 0, 10)
```
Robinnnnn, line 301: is there a reason w set the capacity to 10?

bennycio: It seems like the average amount of collections in a gallery will be less than 10. I thought it would be a reasonable middle-ground number. Although it could be lower given that it is unlikely that a merging user has created a full gallery. Will definitely lower this number.

server/t__routes_user_test.go
```diff
@@ -338,6 +338,160 @@ func TestUserRemoveAddresses_AllAddresses_Failure(t *testing.T) {
...
+	}
+	collID, err := tc.repos.collectionRepository.Create(context.Background(), coll)
+	assert.Nil(err)
+
+	gallery := persist.GalleryDB{
+		OwnerUserID: userID,
+		Collections: []persist.DBID{collID},
+	}
+	galleryID, err := tc.repos.galleryRepository.Create(context.Background(), gallery)
+	assert.Nil(err)
```
Robinnnnn, line 371: this looks like a nifty block to have as a util, like `createPopulatedGallery!`

service/persist/postgres/user.go
Robinnnnn: AH, i assumed this was rigid and couldn't be increased. i was more worried that it was too low, since there are def people with many more than 10 collections. but it seems like it doesn't matter if it's malleable

Robinnnnn (approved):

Robinnnnn (approved):

PR #248, opened 2022-01-31
First cut of integration test

jarrel-b:
Adds the testing framework and a small test that logins in an already existing user. The test will pull down a redis and postgres image, start those containers, run the test, and then delete the containers and volumes. That's so it doesn't wipe your local docker instance if you're testing out a particular thing.

I had to increase the number of open files allowed on my laptop when running the tests on my laptop because it wasn't able to create more web sockets (also ran into the issue on `main`)
```
ulimit -n 2048
```
It was originally set to 256, I think that could mean there are file handlers not being closed somewhere in the test or app since commenting out a portion of the tests also got around the issue. Was planning to look into it after this. 

Tests:
```
> go test ./server/...
ok  	github.com/mikeydub/go-gallery/server	141.322s

> go test ./server/... -run TestUserTestSuite
ok  	github.com/mikeydub/go-gallery/server	10.260s
```

Next steps:
* Verify that this works in CICD
* Adding more steps i.e. create new user ->  login -> update bio -> create a gallery -> update description -> logout


server/server.go
```diff
@@ -43,7 +43,7 @@ type repositories struct {
 }
 
 // Init initializes the server
-func init() {
+func Init() {
```
jarrel-b, line 46: I renamed this to prevent side effects from importing the package

server/t__test_utils_test.go
```diff
@@ -42,6 +44,7 @@ var ts *httptest.Server
 
 type TestUser struct {
 	id       persist.DBID
+	pk       *ecdsa.PrivateKey
```
jarrel-b, line 47: Added the private key so that tests can sign messages

server/t__integration_utils_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}()
+	redis.NewCache(0).Close(false)
+	return
+}
+
+func initPostgres(pool *dockertest.Pool) (*dockertest.Resource, *sql.DB) {
+	pg, err := pool.RunWithOptions(
+		&dockertest.RunOptions{
+			Repository: "postgres",
+			Tag:        "14",
```
radazen, line 46: Good call using the tags we're using in production (Postgres 14, Redis 6) instead of just using `latest` like we do in docker-compose.yml. It's probably worth defining these as consts (this one, Redis, any others in the future) and grouping them all together with a comment about keeping them in sync with what we're doing in docker-compose.yml.

Oooor...this might sound a little over-the-top, but here's a silly question: is there a decent/easy way to read and parse yaml here, such that we could just read the versions from our docker-compose.yml file?

I think it'd probably make sense to put our docker-compose stuff on the same versions we use in prod, or at least on specific versions in case we're starting to do local work on updating to a new image in dev/prod. So I guess my ideal flow would be:

- We decide to change our GCP environment
- We use specific tags in docker-compose.yml to reflect that
- These tests just pull version numbers out of docker-compose.yml and now our test environment reflects our intended deployed environment

Or if parsing yaml here is cumbersome, just a comment in docker-compose.yml noting that these tags should also be updated when we change the tags in docker-compose.yml.

(would also appreciate thoughts from @Robinnnnn or @bennycio here, cause I'm still learning about how we configure stuff 😄)

server/t__test_utils_test.go
```diff
@@ -85,7 +93,10 @@ func initializeTestEnv(a *assert.Assertions, v int) *TestConfig {
 		db = postgres.NewClient()
 	}
 
-	gin.SetMode(gin.ReleaseMode) // Prevent excessive logs
```
radazen, line 88: Accidental deletion, or is there a reason we should no longer be setting this?

server/t__test_utils_test.go
radazen: 🙌🙌🙌

radazen (commented):
Looking great! I left some small comments. Really excited about the direction this is heading in!

server/t__test_utils_test.go
jarrel-b: Whoops good catch that was accident!

server/t__integration_utils_test.go
Robinnnnn: great point! would love to have a single source of truth for this, though i'm also not sure what the best practice is here. i like the short-term idea of adding a comment here (and also the docker-compose file) that we should keep the versions in sync whenever we change them

server/t__test_utils_test.go
Robinnnnn: NOICE

server/t__integration_utils_test.go
bennycio: I agree with Robin, the short term solution with writing a comment in docker-compose.yaml is great. I don't think it would be too cumbersome to read the yaml file, there are lots of libraries that make loading yaml files into structs and maps super easy. I also think the compose file being the source of truth would make it easiest from a technical perspective. If we had another file or location as the source of truth, I can't think of a way off the top of my head for the docker-compose file to read from the source of truth.

server/server.go
bennycio, line 46: I am super down to remove the init that could cause side-effects.

If we are going to remove this side-effect, it would be a good idea to manually call `Init` now where it was previously just imported. Right now we are importing this in `cmd/server/main.go`. 

Thanks Jarrel!

bennycio (approved):
One little comment but other than that, looks fantastic! Thanks so much for helping out with this, Jarrel!

server/t__integration_utils_test.go
jarrel-b: Sounds good! I added logic to read it from the yaml file instead

Robinnnnn (approved):

.circleci/config.yml
```diff
@@ -31,7 +32,44 @@ jobs:
...
+    machine:
+      image: ubuntu-2004:202111-02
+      docker_layer_caching: false  # can enable when on performance plan
+    environment:
+      GO_VERSION: 1.16.13
+      GOTESTSUM_VERSION: 1.7.0
+    steps:
+      - checkout
+      - run:
+          name: Install Go
```
Robinnnnn, line 49: dumb question, but why do we need to install go? AFAIK the unit tests would run without installation (line 35)

.circleci/config.yml
```diff
@@ -31,7 +32,44 @@ jobs:
...
+            echo 'export GOPATH=~/go' >> $BASH_ENV
+            echo 'export PATH=$GOROOT/bin:$GOPATH/bin:$PATH' >> $BASH_ENV
+      - restore_cache:
+          keys:
+            - go-mod-usr-v1-{{ checksum "go.sum" }}
+      - run:
+          name: Install Dependencies
+          command: |
+            go mod download -x
+            go get gotest.tools/gotestsum@v$GOTESTSUM_VERSION
```
Robinnnnn, line 63: i guess same question here- L35 seems to call `gotestsum` without downloading

.circleci/config.yml
```diff
@@ -3,6 +3,7 @@ workflows:
   main:
     jobs:
       - build
+      - integration-tests
```
Robinnnnn, line 6: nit: we have `build`, which contains unit tests, as well as `integration-tests`. i wonder if we should now separate them out into:
```
- build
- unit-tests <-- could this easily build on top of the previous job?
- integration-tests
```

jarrel-b: Good idea, I split that job into build and unit-tests, and made unit-tests dependent on build

.circleci/config.yml
jarrel-b: Ah so we need to install go because this job is using a `machine` executor vs a `docker` executor. The machine executor gives us special privileges to run docker, but it also means we need to add the correct runtime and dependencies. The `ubuntu-2004` does have go installed, but it's not the same version of go used in the project.

.circleci/config.yml
jarrel-b: Same reason as above!

.circleci/config.yml
Robinnnnn: ahhh that makes so much sense, thank you for the explanation!

sorry to nitpick, but could we add a comment block above `integration-tests` to explain that? i think we can say something like:
```
  # using a machine executor to run docker, which in turn installs postgres + redis and runs underlying tests
  # https://circleci.com/docs/2.0/executor-types/#using-machine
  integration-tests:
    ...
```

Robinnnnn (approved):

PR #261, opened 2022-02-02
snapshot endpoint and opensea refreshing

bennycio:
Changes:

- **Added** endpoint for getting the current general card snapshot allowlist from GCP so that the admin interface can add to it and remove from it
- **Added** endpoint for refreshing a user's opensea NFTs

Considerations:

- Is it alright that the snapshot allowlist can be edited by the admin UI? We have to make sure that we keep an accurate list so that we don't remove addresses accidentally from the allowlist when we allowlist new addresses. The allowlist has to be a long running list that we keep off-chain of every address we want to mint. If we are alright with it being edited in the admin UI, are we alright with GCP hosting the source of truth? Personally, I actually think this is the best place we can keep track of the file overall despite us using it in the admin UI.


kaitoo1:
>  so that we don't remove addresses accidentally from the allowlist when we allowlist new addresses.

Is it possible to overwrite the list like this via the admin console? or is it strictly just *adding* addresses?


And more broadly, I'd say it's ok that it can be updated by the admin UI. But we should keep using our spreadsheet to maintain a list of addresses that we are manually adding to the allowlist so that we remember who they are and why we are adding them manually.

admin/admin.go
```diff
@@ -36,7 +39,18 @@ func CoreInit(pqClient *sql.DB) *gin.Engine {
 	router := gin.Default()
 	router.Use(middleware.ErrLogger())
 
-	return handlersInit(router, pqClient, newStatements(pqClient), newEthClient())
+	var s *storage.Client
+	var err error
+	if viper.GetString("ENV") != "local" {
+		s, err = storage.NewClient(context.Background())
+	} else {
+		s, err = storage.NewClient(context.Background(), option.WithCredentialsFile("./_deploy/service-key.json"))
```
Robinnnnn, line 47: haha thanks for fixing this...

Robinnnnn:
i'm fine with being able to update the allowlist on admin as long as we can keep track of historical changes (i.e. state history) for easy reverts. we should also have a big warning to confirm the the update before sending the change!

like kaito said, i think it would be great to retroactively bucket the users into the partnerships they came through, so that they can get special badges or POAPs down the line, and discover the "class" they were in when they were inaugurated into gallery. we don't need to do this on the backend because we're [already keeping track on the frontend](https://github.com/gallery-so/gallery/blob/main/src/scenes/MembershipMintPage/GeneralCardAllowlist.ts#L2841-L2845) – but TL;DR is that when we introduce the next snapshot of users, we'll place them in their own array on the frontend

PR #265, opened 2022-02-02
Add GraphQL support

radazen:
### GraphQL Support

This is a WIP to add GraqhQL support to go-gallery via gqlgen.

### What does it add?

- GraphQL calls can hit the `/glry/graphql/query` route
- You can also point your web browser to the playground GUI at `/glry/graphql/playground`
- Currently supports the `getAuthNonce`, `login`, and `createUser` mutations
- Known/expected backend errors are sent as GraphQL types
- Auth mechanisms are decoupled from functionality that requires auth (e.g. login, createUser), so these methods can be reused with any future auth mechanism we add

### How does it work?

gqlgen parses our graphql schema and generates input/output structs and stubbed-out resolvers. We fill out those resolvers to bind our data to gqlgen responses. A gin handler sends GraphQL queries to gqlgen, and gqlgen uses our resolvers to retrieve data and send it back to the client.

### Implementation details

- Moved server.repositories to persist.Repositories to allow other packages to pass around a repository collection. Necessary only because most GraphQL stuff is in the `graphql` package instead of the `server` package, but we could move everything into the `server` package if need be.

- gqlgen resolvers don't receive a `gin.Context` directly, so we use middleware to store the gin.Context and retrieve it from within a resolver

- In our existing REST implementation, we use hand-written input/output structs to bind and serialize data. When converting a REST route to GraphQL, I'm updating the internal logic to use gqlgen's generated input/output structs instead. REST routes get converted to simple wrappers around the GraphQL functions, moving data between the two different struct types. REST-specific wrappers now end with `REST` (e.g. `LoginREST`) to denote that they're REST wrapper methods. For example, `LoginREST` just calls `Login` and shuffles data between structs accordingly.

- Any error returned by the backend goes through a remapping method that checks whether a GraphQL counterpart exists for the error and whether the current resolver can return errors of that type. If so, the backend error will be returned as an expected GraphQL type. Example: `CreateUser` can return `CreateUserResult | ErrUserExistsWithAddress | ErrSignatureVerificationFailed | ErrAddressDoesNotOwnRequiredNFT`, so the frontend knows what errors it should expect to handle and can test for them easily.

- Authentication has now uses an `Authenticator` interface on the backend. Any auth implementation can implement `Authenticator` and return a Gallery `userID` (if the user exists) and a list of authenticated `addresses` belonging to the user. From what I can tell, this is the minimum set of common outputs that any auth mechanism should supply. The only Authenticator implementation right now is `EthereumNonceAuthenticator`, which handles our existing auth mechanism and has allowed me to remove a bunch of auth code from various methods and consolidate them with a simple `authenticator.Authenticate()` call. In GraphQL, this lets us call `login` like this:
```
mutation LoginMutation {
  login(authMechanism: {
    ethereumEoa:{
      address:"0xSomeAccountAddress",
      nonce:"Gallery uses this cryptographic signature in place of a password: blah blah blah",
      signature:"0xSomeSignatureFortheNonce"
    }
  })
```

- It's worth noting that an authenticator can be supplied _anywhere_ we need to verify that a user has control of a particular set of addresses. E.g. it could be used to merge accounts (logged in user sends an authenticator for some other account to be linked -- not just via nonce verification, but through any future mechanism we use too). This should make it straightforward to do things like adding wallets from other chains to an existing Gallery user.

### What's left?

- Per-resolver auth handling: I'm currently using the `AuthOptional` middleware on the `/glry/graphql/query` route to grab the user's JWT if it's available, but there's no auth beyond that yet
- Resolvers: we only have resolvers for the `getAuthNonce`, `login`, and `createUser` mutations right now
- Validation: GraphQL inputs are not currently validated


graphql/schema/schema.graphql
```diff
@@ -0,0 +1,228 @@
...
+
+  # User Mutations
+  removeUserAddress(address: String!): RemoveUserAddressPayload
+  updateUserInfo(input: UpdateUserInfoInput): UpdateUserInfoPayload
+  # We don't know the auth situation until we speak to Kaito
+  # addUserAddress()
+
+  refreshOpenSeaNfts: RefreshOpenSeaNftsPayload
+
+  getLoginNonce(address: String!): GetLoginNoncePayload
```
radazen, line 221: This is the equivalent of `getPreflight`, and actually calls our existing GetPreflight method internally. @tbezman's original schema had a mutation for getting a nonce; I renamed it to `getLoginNonce` to be a little more explicit, but I'm curious: any objections to renaming GetPreflight? I always found it a _little_ odd that it's called GetPreflight but isn't actually what Chrome dev tools considers to be a Preflight request.

graphql/schema/schema.graphql
```diff
@@ -0,0 +1,228 @@
...
+
+  # Login with wallet type 0
+  loginWithEOA(address: String!, nonce: String!, signature: String!): LoginPayload
```
radazen, line 224: @tbezman Any particular reason the two login methods were separated in the original schema? I'd like to condense them into a single `login` mutation that takes a wallet type, since they take the same parameters server-side and logging in with a smart contract might still involve different wallet types (wallet type 1 is Gnosis, but I can see a future where we're sending different wallet types for smart contract logins).

util/helpers.go
```diff
@@ -82,3 +82,9 @@ func GetValueFromMapUnsafe(m map[string]interface{}, key string, searchDepth int
 	}
 	return nil
 }
+
+// StringToPointer simply returns a pointer to the parameter string. It's useful for taking the address of a string concatenation,
+// a function that returns a string, or any other string that would otherwise need to be assigned to a variable before becoming addressable.
+func StringToPointer(str string) *string {
```
radazen, line 88: I added this because I was tired of using two lines to take the address of a `NullString.String()` or a concatenated string

graphql/schema/schema.graphql
tbezman: That makes sense. I think the different use cases warrant separate mutations but I'll defer to you on this one.

gqlgen.yml
```diff
@@ -0,0 +1,56 @@
+# Where are all the schema files located? globs are supported eg  src/**/*.graphqls
```
Robinnnnn, line 1: are the questions in this file posed to the reviewer, or did they come like that?

either way, i think the locations you chose are great!

radazen: Haha. I can understand why you'd think these were my questioning comments, but they're part of the default config file. I didn't change much in here beyond splitting things into subdirectories.

graphql/schema/schema.graphql
Robinnnnn: in favor! let's just make sure to leave the REST endpoint alone to `get_preflight` so we don't have to bump the frontend

graphql/schema/schema.graphql
Robinnnnn: if they're similar enough it's probably fine to merge- the only thing is we'll want to support logins with more wallet types (solana, tezos, clover, etc.)

Robinnnnn: and also possibly oauth-type logins like nifty gateway / foundation

radazen: Gotcha. Happy to split it up by use case with things like OAuth and multi-chain on the horizon. `loginWithEthereumEOA`, `loginWithGnosisWallet`, `loginWithSolanaSomething`, etc? We'll probably end up with a lot of login methods, but we can reassess in the future.

service/auth/auth.go
```diff
@@ -118,78 +120,108 @@ func GenerateNonce() string {
 	return nonceStr
 }
 
-// LoginAndMemorizeAttempt will run the login pipeline and memorize the result
-func LoginAndMemorizeAttempt(pCtx context.Context, pInput LoginInput,
+// LoginAndRecordAttemptREST will run the login pipeline and memorize the result
+func LoginAndRecordAttemptREST(pCtx context.Context, pInput LoginInput,
```
radazen, line 124: Renamed this from "LoginAndMemorizeAttempt" because "memoize" is an actual thing that's very easy to confuse with "memorize"

graphql/resolver/resolver.go
```diff
@@ -0,0 +1,42 @@
...
+
+type Resolver struct {
+	Repos     *persist.Repositories
+	EthClient *ethclient.Client
+}
+
+func GinContextFromContext(ctx context.Context) (*gin.Context, error) {
+	ginContext := ctx.Value(GinContextKey)
+	if ginContext == nil {
+		err := fmt.Errorf("gin.Context not found in current context")
```
bennycio, line 29: maybe instead of initializing this error every time ginContext is nil we can add it as a top level variable since it is a constant error.

`var errGinCtxNotFoundInCtx = errors.New("gin.Context not found in current context")`

server/server.go
```diff
@@ -118,27 +103,25 @@ func setDefaults() {
 	}
 }
 
-func newRepos(db *sql.DB) *repositories {
-
+func newRepos(db *sql.DB) *persist.Repositories {
```
bennycio, line 106: Could we add this to the persist package where `persist.Repositories` is defined? Is there any world where we are going to be using different packages to implement the repository interfaces across the application? Or possibly use different redis db's?

bennycio (commented):
Looks amazing Ezra! So excited to see graphQL in action soon!

server/server.go
radazen: I initially tried to add this to the `persist` package...

...and then I had my first experience with golang not allowing import cycles. Initially, I wasn't planning to move `repositories` or `newRepos` at all, but that meant the `graphql` package referenced `server` to get access to this struct, and `server` referenced `graphql` to set up Gin handlers, so we had a cycle.

I moved the struct and `newRepos` into the `persist` package, but then...`persist` needed to import `postgres` to call the `NewXyzRepository` functions, but `postgres` imports `persist` to use our database models. Import cycle again.

Soooo my best alternative solution was to put the `Repositories` struct in persist, but delegate the construction of said struct to the `server` package, since it's the place where we pick `postgres` for our repository implementations.

graphql/resolver/resolver.go
```diff
@@ -0,0 +1,42 @@
...
+		logrus.Errorf("could not retrieve gin.Context: %v", err)
+		return nil, err
+	}
+
+	gc, ok := ginContext.(*gin.Context)
+	if !ok {
+		err := fmt.Errorf("gin.Context has wrong type")
```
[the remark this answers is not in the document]
radazen: Totally! On a related note, these errors are a little annoying, since most errors would be pushed into the Gin context and then logged as part of the error handler middleware...but if we're hitting these errors, we don't have access to the Gin context. Do you think logging these directly with logrus here is sufficient, or is there something else I should be doing too?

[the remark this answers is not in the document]
bennycio: I think that because the resolver returns the error it should be alright to just log it here so that we can see it in app engine logs. Another possibility would be to panic here. We should never encounter this error if the codebase was setup properly. A user cannot trigger this error. That would make this a good candidate for panic I think. I am not sure how gqlgen responds to panics, however.

server/server.go
bennycio: Yeah I think if we did want to move it, it would make sense to put it in the `postgres` package. I guess it just comes down to whether you think we will ever want to do something in the future where we have a mix of repositories. Although in that case we could also just setup an alternate function for mixing different implementations. We could have a `postgres.NewRepositories` func and in a package specific `newRepos` if we ever needed it. Right now it seems like we only need postgres repositories.

radazen: Yeah, agreed. Honestly happy with either option. I'll move it to `postgres` and see how it feels. Like you said, we can always change it in the future if we need to.

graphql/resolver/resolver.go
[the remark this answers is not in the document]
radazen: Ah, good call. AFAIK, gqlgen just returns the error message for any resolver that encounters a panic. Can't resolve anything without the gin context, so you're totally right, this is a great candidate for a panic.

graphql/schema/schema.graphql
```diff
@@ -0,0 +1,228 @@
...
+}
+
+type GetLoginNoncePayload {
+  nonce: String
+  user_exists: Boolean
+}
+
+type LoginPayload {
+  signatureValid: Boolean
+  jwtToken: String # probably not needed anymore?
```
kaitoo1, line 198: correct,  the front end doesn't use this field anymore 👍

graphql/resolver/schema.resolvers.go
```diff
@@ -0,0 +1,149 @@
...
+	// TODO: This currently gets its value from the AuthOptional middleware applied to the entire GraphQL endpoint,
+	// but that won't suffice when we transition auth-only endpoints to GraphQL too.
+	authed := gc.GetBool(auth.AuthContextKey)
+
+	output, err := auth.GetAuthNonce(gc, persist.Address(address), authed, r.Repos.UserRepository, r.Repos.NonceRepository, r.EthClient)
+
+	if err != nil {
+		// Map known errors to GraphQL return types
+		if errorType, ok := r.errorToGraphqlType(err); ok {
+			if returnType, ok := errorType.(model.GetAuthNoncePayload); ok {
```
radazen, line 62: I'm hoping that this whole error remapping block can be moved into a more succinct function when generics are released in go 1.18. As it stands, this 10-line block needs to be copy/pasted with an updated `model.Type` in every resolver.

Robinnnnn: classic!

util/helpers.go
```diff
@@ -82,3 +85,31 @@ func GetValueFromMapUnsafe(m map[string]interface{}, key string, searchDepth int
...
+	return &str
+}
+
+// GinContextFromContext retrieves a gin.Context previously stored in the request context via the GinContextToContext middleware,
+// or panics if no gin.Context can be retrieved (since there's nothing left for the resolver to do if it can't obtain the context).
+func GinContextFromContext(ctx context.Context) *gin.Context {
```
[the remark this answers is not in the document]
Robinnnnn: seems chill

graphql/resolver/schema.resolvers.helpers.go
```diff
@@ -0,0 +1,65 @@
...
+
+	if mappedErr != nil {
+		return mappedErr, true
+	}
+
+	return nil, false
+}
+
+// authMechanismToAuthenticator takes a GraphQL AuthMechanism and returns an Authenticator that can be used for auth
+func (r *Resolver) authMechanismToAuthenticator(m model.AuthMechanism) (auth.Authenticator, error) {
```
Robinnnnn, line 41: love this

graphql/resolver/schema.resolvers.go
Robinnnnn: just so i understand what's happening in this block:
- if there's an error, see if it's an error that corresponds to one that's defined specifically as a return type for this resolver
- if it's not defined (or doesn't map properly), just return a pure error

server/handler.go
```diff
@@ -1,80 +1,116 @@
...
 	"github.com/mikeydub/go-gallery/service/pubsub"
+	"github.com/spf13/viper"
 )
 
-func handlersInit(router *gin.Engine, repos *repositories, ethClient *ethclient.Client, ipfsClient *shell.Shell, psub pubsub.PubSub) *gin.Engine {
+func handlersInit(router *gin.Engine, repos *persist.Repositories, ethClient *ethclient.Client, ipfsClient *shell.Shell, psub pubsub.PubSub) *gin.Engine {
 
 	apiGroupV1 := router.Group("/glry/v1")
 	apiGroupV2 := router.Group("/glry/v2")
+	graphqlGroup := router.Group("/glry/graphql")
```
Robinnnnn, line 23: man should we just drop the `glry` at this point? lol

service/auth/auth.go
```diff
@@ -359,17 +443,34 @@ func GetPreflight(pCtx context.Context, pInput GetPreflightInput, pPreAuthed boo
...
 
-	return output, nil
+	return &output, nil
+}
+
+// GetAuthNonceREST will determine whether a user is permitted to log in, and if so, generate a nonce to be signed
+func GetAuthNonceREST(pCtx context.Context, pInput GetPreflightInput, pPreAuthed bool,
+	userRepo persist.UserRepository, nonceRepo persist.NonceRepository, ethClient *ethclient.Client) (*GetPreflightOutput, error) {
+
+	gqlOutput, err := GetAuthNonce(pCtx, pInput.Address, pPreAuthed, userRepo, nonceRepo, ethClient)
```
Robinnnnn, line 463: this technically isn't `gqlOutput` right? bc we're in a function that's REST

service/user/user.go
```diff
@@ -267,24 +256,49 @@ func CreateUser(pCtx context.Context, pInput AddUserAddressesInput, userRepo per
...
+		Address:    pInput.Address.String(),
+		Nonce:      pInput.Nonce,
+		Signature:  pInput.Signature,
+		WalletType: pInput.WalletType,
+		UserRepo:   userRepo,
+		NonceRepo:  nonceRepo,
+		EthClient:  ethClient,
+	}
+
+	gqlOutput, err := CreateUser(pCtx, authenticator, userRepo, galleryRepo)
```
Robinnnnn, line 292: nit: same comment

graphql/resolver/schema.resolvers.go
radazen: Yeup. Gqlgen takes this line from our schema:

`union GetAuthNoncePayload = AuthNonce | ErrAddressDoesNotOwnRequiredNFT`

and generates an interface named `model.GetAuthNoncePayload`, along with `model.AuthNonce` and `model.ErrAddressDoesNotOwnRequiredNFT` structs that implement the `model.GetAuthNoncePayload` interface. Once we get a remapped error back, we try to cast it to `model.GetAuthNoncePayload` to see if it's one of the valid return types in the union.

server/handler.go
radazen: Haha. I'm okay either way! But if we do want to drop it, this is a convenient time. 😁

service/auth/auth.go
radazen: (the nuance is that `model.AuthNonce` isn't technically specific to GraphQL, or at least it doesn't have to be...but it _is_ generated by gqlgen and used primarily for GraphQL support)

graphql/schema/schema.graphql
```diff
@@ -0,0 +1,265 @@
...
+  updateCollectionInfo(input: UpdateCollectionInfoInput!): UpdateCollectionInfoPayload
+  updateCollectionNfts(input: UpdateCollectionNftsInput!): UpdateCollectionNftsPayload
+
+  # Gallery Mutations
+  updateGalleryCollections(input: UpdateGalleryCollectionsInput): UpdateGalleryCollectionsPayload
+
+  # User Mutations
+  removeUserAddress(address: String!): RemoveUserAddressPayload
+  updateUserInfo(input: UpdateUserInfoInput): UpdateUserInfoPayload
+  # We don't know the auth situation until we speak to Kaito
```
kaitoo1, line 256: hello

kaitoo1: am I a blocker here 👀

radazen: Oh! Haha. No, that was a comment from Terence's first schema draft. When we were brainstorming about what we'd need, we weren't sure if Gnosis had any special requirements beyond what we use for regular Ethereum wallets. But AFAIK -- and please correct me if I'm wrong! -- Gnosis uses the same REST routes with the same input fields (address, nonce, signature, walletType) as regular EOA auth, so nothing here really needs special handling.

service/auth/auth.go
```diff
@@ -118,78 +150,128 @@ func GenerateNonce() string {
...
+	WalletType WalletType
+	UserRepo   persist.UserRepository
+	NonceRepo  persist.NonceRepository
+	EthClient  *ethclient.Client
+}
 
-	output, err := LoginPipeline(pCtx, pInput, userRepo, nonceRepo, ec)
-	if err != nil {
-		return LoginOutput{}, err
+func (e EthereumNonceAuthenticator) Authenticate(pCtx context.Context) (*AuthResult, error) {
```
kaitoo1, line 163: love this refactor!

graphql/schema/schema.graphql
kaitoo1: ah gotcha. yup that's correct Gnosis and EOA use the same REST routes

kaitoo1 (approved):
Great work @radazen 🙌

Thanks for the great PR description and comments in the code - they helped a lot to digest these changes.
I'm excited to get closer to switching the frontend to authenticate using graphql!!

PR #269, opened 2022-02-04
fix uri type bug, ensure data is not oversized, cache media separately

bennycio:


cmd/migration/main.go
```diff
@@ -133,6 +158,81 @@ func getAllUsers(ctx context.Context, pgClient *sql.DB) map[persist.DBID][]persi
...
+	if err != nil {
+		return nil, err
+	}
+	allTokens := make([]persist.Token, 0, 5)
+	asURI := persist.TokenURI(nft.ImageURL)
+	media := persist.Media{}
+
+	bs, err := rpc.GetDataFromURI(ctx, asURI, ipfsClient, arweaveClient)
+	if err == nil {
+		mediaType := persist.SniffMediaType(bs)
```
radazen, line 173: 🙌🙌🙌

cmd/migration/main.go
```diff
@@ -133,6 +158,81 @@ func getAllUsers(ctx context.Context, pgClient *sql.DB) map[persist.DBID][]persi
...
+		if mediaType != persist.MediaTypeUnknown {
+			media = persist.Media{
```
radazen, line 175: Nit: could set the fields of the existing `media` struct instead of allocating a new one

cmd/update/main.go
```diff
@@ -0,0 +1,128 @@
...
+	wp := workerpool.New(6)
+	go func() {
+		for {
+			time.Sleep(time.Minute)
+			logrus.Warnf("Workerpool queue size: %d", wp.WaitingQueueSize())
+		}
+	}()
+	for u, addrs := range users {
+		userID := u
+		addresses := addrs
```
radazen, line 75: Any reason this shouldn't just be `for userID, addresses := range users {`?

indexer/indexer.go
```diff
@@ -545,28 +532,19 @@ func processTransfers(i *Indexer, transfers []transfersAtBlock, uris chan<- toke
 				}
 				cancel()
 
-				id, err := util.HexToBigInt(string(tokenID))
-				if err != nil {
-					panic(fmt.Sprintf("error converting tokenID to bigint: %s", err))
-				}
-
-				uriReplaced := persist.TokenURI(strings.TrimSpace(strings.ReplaceAll(u.String(), "{id}", id.Text(16))))
+				uriReplaced := persist.TokenURI(strings.TrimSpace(strings.ReplaceAll(u.String(), "{id}", tokenID.ToUint256String())))
```
radazen, line 535: Probably worth adding a utility method for `strings.TrimSpace(strings.ReplaceAll(u.String(), "{id}", tokenID.ToUint256String()))`, since we do it elsewhere too

radazen (approved):
LGTM! 🚀

cmd/update/main.go
bennycio: Yes, this is a weird go thing where using a value from a for loop in a goroutine (or in this case in a worker pool made up of goroutines) will use the current value that loop holds when the goroutine runs unless the value is captured inside the loop like I have done there. If I don't do that, it causes a race condition that will lead to unexpected results as the array values at `u` and `addrs` change each loop despite the goroutines running separately from the loop.

cmd/migration/main.go
```diff
@@ -95,7 +115,12 @@ func getNewCollections(ctx context.Context, pgClient *sql.DB, userIDs map[persis
 					if _, ok := err.(persist.ErrTokenNotFoundByIdentifiers); !ok {
 						panic(err)
 					} else {
-						logrus.Infof("Token equi not found for %s-%s in collection %s", fullNFT.OpenseaTokenID, fullNFT.Contract.ContractAddress, coll)
+						logrus.Infof("Token equi not found for %s-%s in collection %s. Making token...", fullNFT.OpenseaTokenID, fullNFT.Contract.ContractAddress, coll)
```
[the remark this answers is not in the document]
bennycio: 👌 yes it does, I will change that to make it more clear

cmd/migration/main.go
```diff
@@ -27,20 +38,26 @@ func run() {
...
 	userIDs := getAllUsers(ctx, pgClient)
 
-	usersToNewCollections := getNewCollections(ctx, pgClient, userIDs, nftRepo, tokenRepo)
+	usersToNewCollections := getNewCollections(ctx, pgClient, userIDs, nftRepo, tokenRepo, ethClient, ipfsClient, arweaveClient)
 
 	updateCollections(ctx, pgClient, usersToNewCollections)
 }
 
 func setDefaults() {
-	viper.SetDefault("ENV", "local")
```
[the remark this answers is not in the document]
bennycio: It is never used in the migration script because the migration script is only run locally anyway.

cmd/update/main.go
radazen: Doh! I've totally read about this being a Golang gotcha! Thanks for the explanation.

PR #270, opened 2022-02-07
only remove old NFTs in opensea refresh

bennycio:
Changes:

- **Removed** the step in the user remove addresses endpoint that will go through a user's collections and remove any NFT associated with the addresses being removed
- **Added** a step at the end of the opensea refresh that will ensure that all of the NFTs that a user has in their collections are owned by addresses that they own
- **Added** a test to ensure that this last step in the opensea refresh will remove NFTs that are associated with old addresses

The issue that this fixes is when a user loses progress on their gallery after moving their NFTs between their wallet addresses and removes an address from their gallery. The reason why this would occur was because although they had sent their NFTs between their wallets, the NFTs were not changing owner addresses on the backend because they had not refreshed opensea yet. That means that the NFTs in their collections were associated with the address they are removing and were then removed from their collections because of this.


Robinnnnn (approved):
🎉🎉🎉

PR #283, opened 2022-02-11
Read test-wallet.json from _internal directory

radazen:
Small tweak to read test-wallet.json from `_internal`, since that's where it ends up by default on a developer's machine


jarrel-b (approved):
Makes sense!

PR #287, opened 2022-02-18
instantiate collection in each iteration to reset previous values

kaitoo1:
Description: 
This PR tries to fix a bug I saw when testing whitespace blocks.
After saving some collections with whitespaces in a collection, I tried to retrieve my Gallery and found that in the response, collections were getting their whitespace values combined with whitespace values from other collections.

it's easier to see with examples, so here are 2 cases:
2 collections (one with [0,0,0,1] and one with no whitespace)
db: 
Screen Shot 2022-02-18 at 23 45 51
response:
CleanShot 2022-02-18 at 23 43 01@2x

2 collections (one with [3,3,3,3,6,7,8] and one with [2,2,2,2,3])
db:
Screen Shot 2022-02-18 at 23 49 45

response:
CleanShot 2022-02-18 at 23 48 53@2x

The db query was fine and returning results as expected. It looks like the issue was with how we hold on to the `collection` value outside of the `rows.Scan` loop. We overwrite fields on `collection` with each iteration, but it looks like if there are any fields that aren't present in the current row, the old value doesn't get cleared. It also appears that when there is a whitespace value present, it doesn't fully replace the previous value, it just overwrites part of the list.

My solution was to re instantiate `collection` in every iteration to eliminate the risk of holding on to any stale values  from the previous row.

### DIFF
--- a/service/persist/postgres/gallery_nft.go
+++ b/service/persist/postgres/gallery_nft.go
@@ -208,9 +208,9 @@ func (g *GalleryRepository) GetByUserID(pCtx context.Context, pUserID persist.DB
 	galleries := make(map[persist.DBID]persist.Gallery)
 	collections := make(map[persist.DBID][]persist.Collection)
 	var gallery persist.Gallery
-	var collection persist.Collection
+	var lastCollID persist.DBID
 	for rows.Next() {
-		lastCollID := collection.ID
+		var collection persist.Collection
 		var nft persist.CollectionNFT
 
 		err := rows.Scan(&gallery.ID, &gallery.Version, &gallery.OwnerUserID, &gallery.CreationTime, &gallery.LastUpdated,
@@ -252,6 +252,7 @@ func (g *GalleryRepository) GetByUserID(pCtx context.Context, pUserID persist.DB
 		}
 
 		collections[gallery.ID] = colls
+		lastCollID = collection.ID
 	}
 	if err := rows.Err(); err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


service/persist/postgres/gallery_token.go
```diff
@@ -319,10 +321,10 @@ func (g *GalleryTokenRepository) GetByID(pCtx context.Context, pID persist.DBID)
 	galleries := make(map[persist.DBID]persist.GalleryToken)
 	collections := make(map[persist.DBID][]persist.CollectionToken)
 	var gallery persist.GalleryToken
-	var collection persist.CollectionToken
-	var nft persist.TokenInCollection
```
kaitoo1, line 323: In this file I noticed the `GetByUserID` method has  `var nft persist.TokenInCollection` defined within the for `rows.Next()`  loop, but the `GetByID` method has `var nft persist.TokenInCollection` outside of the `rows.Next()`  loop.

Should they be the same? I moved it so that they're consistent, but lmk if they should be different.

This made me a bit concerned about how we are maintaining the same or very similar logic in 4 places (getByUserId and getById in gallery_token and gallery_nft) for 2 main reasons:
One, this makes it challenging to keep them all in sync when we make a change like with this PR where we want to change how we retrieve a gallery. Two, when there are slight differences, it's difficult to tell if they're deliberate differences, or if they're unintentional and got out of sync.

I know we are moving off of gallery_nft soon so we don't have to over invest in improving it, but 
even with gallery_token, maybe some of the shared logic in getByUserId and getById can be abstracted into a shared method or two. Maybe I'm missing slight nuances, but the logic in `for rows.Next()` for both methods looked very similar

kaitoo1: @bennycio  would love to get your thoughts. especially the first part, to make sure I'm not merging something incorrect 🙏

bennycio: Yes, the nft should definitely be within the loop, thank you for fixing this. 

I would also love to abstract some of this logic, the issue I had when I was upkeeping both token and nft persist funcs was that because the entities in the loop are different types (`Collection` vs `CollectionToken` and `CollectionToken` vs `TokenInCollection`) we could not make a single function that handles both types.

service/persist/postgres/gallery_nft.go
```diff
@@ -351,7 +352,9 @@ func (g *GalleryRepository) GetByID(pCtx context.Context, pID persist.DBID) (per
 		}
 
 		collections[gallery.ID] = colls
+		lastCollID = collection.ID
```
bennycio, line 355: I believe this line also needs to be added to `gallery_token.go`'s `GetByID` function. Thanks for figuring this out!

bennycio (approved):
One little change and this is good to go! Thanks Kaito!

service/persist/postgres/gallery_nft.go
kaitoo1: good catch, added 👍

kaitoo1: thanks for looking Benny 😎

PR #289, opened 2022-02-19
GraphQL - Phase 2

radazen:
**_NOTE TO REVIEWERS: Feel free to ignore generated files -- anything named `generated.go` or ending in `_gen.go`._**

### Schema updates
- Queries and mutations can be tagged with the `@authRequired` directive, which will reject unauthenticated requests. This is similar to the `AuthRequired` middleware we apply to authenticated REST routes.

- Replaced `ErrSessionExpired` and `ErrNoCookie` with `ErrNotAuthorized`, which has an `errorType` field denoting what kind of auth failure was encountered. This should make it easy to add auth failures to any return type union just by adding `ErrNotAuthorized`.

- Adopted a convention for return type unions: `XyzOrError`. It sounds kind of dumb, but I've found it's a useful way to describe operations that will return a union with a return type or an error. E.g. the `Viewer` query now returns `ViewerOrError`, which is a union of `Viewer` and all of the different error types the query might return.

### Auth updates
- Cookies have a 10-year MaxAge now so they won't expire before their JWT does

- GraphQL routes marked with `@authRequired` will return appropriate auth errors based on cookie/JWT status

- Some minor refactoring of the way we store auth info in the request context -- auth context keys (`user_id`, `authenticated`) are namespaced to avoid accidental key name collisions (now `auth.user_id` and `auth.authenticated`), and there are helper methods to set those fields instead of letting other packages set them directly.

### Resolvers
- `Viewer` and `UserByUsername` root queries are available. Lots of other fields are available or partially available, but there's a lot here that's still in-progress. If a field comes back null or returns an unhandled error, it probably isn't implemented yet.

- GraphQL database lookups are cached via dataloaders, which makes it much easier to write modular/reusable resolvers. Dataloaders can also be used to batch database queries, but I haven't set that up yet.

### Other Notes
- Schema types like `ImageNFT` and `VideoNFT` are difficult to implement at the moment because we're not storing media type info on the backend. We _will_ be storing this information when we switch from Opensea to our own indexer, but we may have to wait until then to fully implement this approach in GraphQL.


graphql/schema/schema.graphql
```diff
@@ -102,16 +111,16 @@ type ViewerGallery {
 
 type Viewer {
   user: GalleryUser
-  wallets: [Wallet]
-  viewerGallery: ViewerGallery
+  wallets: [Wallet] # TODO: Is this field necessary?
```
radazen, line 114: @tbezman Not sure if we need this field, since the user object has `wallets` on it too. Do you remember if there was a specific reason for it?

graphql/resolver/schema.resolvers.helpers.go
```diff
@@ -63,3 +67,138 @@ func (r *Resolver) authMechanismToAuthenticator(m model.AuthMechanism) (auth.Aut
...
+
+	return output, nil
+}
+
+// addressesToWalletModels converts a slice of persist.Address to a slice of model.Wallet
+func addressesToWalletModels(ctx context.Context, r *Resolver, addresses []persist.Address) []*model.Wallet {
+	wallets := make([]*model.Wallet, len(addresses))
+	for i, address := range addresses {
+		wallets[i] = &model.Wallet{
+			ID:      "", // TODO: What's a wallet's ID?
```
radazen, line 197: Unlike our other GraphQL node types, wallets don't have a database model. They're just address strings. As such, I'm not sure what's appropriate here -- is a wallet's ID just its address? Feels a little inconsistent, given that every other node's ID is its database ID. Maybe wallets just don't implement the `Node` interface and don't have IDs?

service/auth/jwt.go
```diff
@@ -35,9 +35,8 @@ func ValidateJWT() gin.HandlerFunc {
 	}
 }
 
-// JWTParse parses the JWT token from the request and returns whether the token is valid and the user ID associated with it
-func JWTParse(pJWTtokenStr string,
-	pJWTsecretKeyStr string) (bool, persist.DBID, error) {
+// JWTParse parses the JWT token from the request and returns the user ID associated with it
+func JWTParse(pJWTtokenStr string, pJWTsecretKeyStr string) (persist.DBID, error) {
```
radazen, line 39: I removed `valid` as a return value here because it was redundant -- an invalid JWT always returns an error, so we might as well use errors as the source of truth for validity.

service/user/user.go
```diff
@@ -98,8 +98,9 @@ type errUserNotFound struct {
 	username string
 }
 
-type ErrUserExistsWithAddress struct {
-	Address persist.Address
+type ErrUserAlreadyExists struct {
```
radazen, line 101: Reworking some error types to make them work with Authenticators

graphql/dataloader/dataloaders.go
```diff
@@ -0,0 +1,227 @@
...
+	"time"
+)
+
+const loadersKey = "graphql.dataloaders"
+const defaultMaxBatch = 100
+const defaultWaitTime = 1 * time.Millisecond
+
+// Loaders will cache and batch lookups. They are short-lived and should never persist beyond
+// a single request, nor should they be shared between requests (since the data returned is
+// relative to the current request context, including the user and their auth status).
```
Robinnnnn, line 24: 😍

graphql/resolver/auth.go
```diff
@@ -0,0 +1,70 @@
...
+		if viper.GetBool("REQUIRE_NFTS") {
+			user, err := dataloader.For(ctx).UserByUserId.Load(userID.String())
+			if err != nil {
+				notAuthorized := model.ErrNotAuthorized{Message: err.Error(), ErrorType: model.AuthFailureTypeInternalError}
+				return notAuthorized, nil
+			}
+
+			has := false
+			for _, addr := range user.Addresses {
+				allowlist := auth.GetAllowlistContracts()
```
Robinnnnn, line 54: note: i know this just implements our existing middleware, but i wonder if this adds significant delay to authenticated requests

graphql/resolver/schema.resolvers.helpers.go
Robinnnnn: i could see wallets being its own database model down the line

service/auth/auth.go
```diff
@@ -95,6 +99,11 @@ type GetPreflightOutput struct {
 }
 
 type Authenticator interface {
+	// GetDescriptor returns information about the authenticator for error and logging purposes.
+	// NOTE: GetDescriptor should NOT include any sensitive data (passwords, auth tokens, etc)
+	// that we wouldn't want showing up in logs!
+	GetDescriptor() AuthDescriptor
```
Robinnnnn, line 105: nice!

service/auth/jwt.go
Robinnnnn: confirmed we don't seem to use this on frontend.

we do use `signature_valid`, but shouldn't affect this function:
- https://github.com/gallery-so/gallery/blob/main/src/components/WalletSelector/authRequestUtils.ts#L9
- https://github.com/gallery-so/gallery/blob/aaa8a6a3c8293428f6dcea0e675d840d94ccae44/src/components/WalletSelector/AddWalletPending/AddWalletPendingDefault.tsx#L87

service/persist/persist.go
```diff
@@ -213,6 +216,36 @@ func (n *NullInt64) Scan(value interface{}) error {
...
+// Scan implements the database/sql Scanner interface for the NullInt32 type
+func (n *NullInt32) Scan(value interface{}) error {
+	if value == nil {
+		*n = NullInt32(0)
+		return nil
+	}
+	// database/sql spec says integer values should be returned as int64, even if the underlying column is int32
+	*n = NullInt32(value.(int64))
+	return nil
+}
```
Robinnnnn, line 247: nice!

graphql/schema/schema.graphql
tbezman: Is that public information? `Viewer` is meant to represent things only the logged in user could see. @Robinnnnn can you confirm it's okay for everyone to see a given user's wallets?

graphql/schema/schema.graphql
```diff
@@ -185,25 +194,26 @@ type AuthNonce {
...
 
-union GetAuthNoncePayload = AuthNonce | ErrAddressDoesNotOwnRequiredNFT
+union GetAuthNoncePayloadOrError = AuthNonce | ErrAddressDoesNotOwnRequiredNFT
 
-type ErrAddressDoesNotOwnRequiredNFT implements Error {
-  message: String!
-}
+# InternalError means that something unexpected happened. The error message may give more detail,
+# but it isn't an error type that is expected or requires specific handling.
+enum AuthFailureType { NoCookie, InvalidToken, DoesNotOwnRequiredNFT, InternalError }
```
tbezman, line 201: I don't like the idea of encoding `InternalError` as a type here. Technically anything can fail so if we were following this, every type should be a union of `X | InternalError`. Can we just leave this out and use standard GraphQL unhandled errors?

tbezman: I also am not a huge fan of using an enum here. I think it should just be a union type. In the case that we need to pass more information along with a given type. I think there are valid use cases for enums but I don't think this is one of them. Happy to have some back and forth on this one.

tbezman (commented):
Leaving some comments on the schema. I think that's where my review is most helpful. I will defer to @bennycio for good ol' Golang review :)

graphql/resolver/auth.go
Robinnnnn: added a task to to our board!

Robinnnnn:
> can you confirm it's okay for everyone to see a given user's wallets?
yup! totally cool

service/auth/jwt.go
kaitoo1: confirmed we don't use the return value

graphql/schema/schema.graphql
tbezman: @radazen Marking this as resolved since we talked through it on Meet today.

graphql/resolver/auth.go
radazen: Follow-up: I misread the underlying NFT-querying code; I actually don't know enough about the underlying Ethereum library to know if the current NFT membership check scales well with a large number of NFTs or not. That being said, caching this status for a while still sounds like a good idea.

PR #284, opened 2022-02-17
Add whitespace

jarrel-b:
**Changes**
* Adds `Whitespace` to Layout
* Adds integration test for collections
* Fixes a small bug where missing collections returned an internal error instead of a not found

PR #290, opened 2022-02-21
Remove bson tags

jarrel-b:
**Changes**
* Follow up to https://github.com/gallery-so/go-gallery/pull/284#discussion_r808728971 to remove `bson` tags since mongo is no longer used


Robinnnnn (approved):
NICE!

PR #296, opened 2022-02-23
side effects, organization, handle subscribed logs differently

bennycio:
Changes:

- **Added** side effect handling for ensuring that NFTs transferred away from a user leave their collections
- **Added** new indexer pipeline funcs for subscribed logs where we want to grab more data such as media content
- **Changed** the indexer file to be a bit more organized and separated out some common logic

Considerations:

- Right now media content (the media field on a token) is only retrieved in the "listening" stage of the indexer after the initial indexing process. This is because the initial indexing process is very expensive and would be very time consuming if it had to figure out for every NFT if there was a user associated with it and then subsequently grab media content for that NFT which could take anywhere from a few milliseconds to a few minutes (if the context allows for it). I believe handling side-effects like grabbing media content and checking if NFTs are transferred out of someone's collections is fine in the listening stage given that we receive NFTs one at a time and only once an Ethereum block has been produced. 
- I decided that it would be good to handle the side-effect of removing NFTs from collections of users who transferred away the NFTs because that would prevent us from having some sort of manual refresh step for users to get rid of NFTs they no longer have. This side-effect is handled by first checking if the `from` address in a transfer is owned by a user (meaning there is a database call for every NFT that comes through the pipeline) and if there is a user going through their collections to see if they have the NFT anywhere and remove it (2 more database calls). Another option to offload this kind of side-effect being handled for every NFT that comes through the indexer would be to run this when a user logs in. It could be a background process that is async. Kind of like how right now when users transfer NFTs they are still in their collections until they refresh opensea, except rather than refreshing it will happen automatically on the backend on login.

### DIFF
--- a/features/access.go
+++ b/features/access.go
@@ -85,7 +85,11 @@ func upsertAccessState(pCtx context.Context, userID persist.DBID, userRepo persi
 	for _, feature := range allFeatures {
 		switch feature.TokenType {
 		case persist.TokenTypeERC1155:
-			address, tokenID := feature.RequiredToken.GetParts()
+			address, tokenID, err := feature.RequiredToken.GetParts()
+			if err != nil {
+				logrus.WithError(err).Error("failed to get parts from required token")
+				return err
+			}
 			ca, err := contracts.NewIERC1155Caller(address.Address(), ethClient)
 			if err != nil {
 				logrus.WithError(err).Error("failed to initialize ERC1155 caller")
@@ -102,7 +106,11 @@ func upsertAccessState(pCtx context.Context, userID persist.DBID, userRepo persi
 			}
 			tis[feature.RequiredToken] = totalBal.Uint64()
 		case persist.TokenTypeERC721:
-			address, tokenID := feature.RequiredToken.GetParts()
+			address, tokenID, err := feature.RequiredToken.GetParts()
+			if err != nil {
+				logrus.WithError(err).Error("failed to get parts from required token")
+				return err
+			}
 			ca, err := contracts.NewIERC721Caller(address.Address(), ethClient)
 			if err != nil {
 				logrus.WithError(err).Error("failed to initialize ERC1155 caller")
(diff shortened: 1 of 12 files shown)


Robinnnnn:
benny, thanks so much for the detailed breakdown!

i would definitely prefer to handle the side-effect on user login to reduce calls to the DB. i don't think it's a huge deal that a user's gallery could show somewhat stale data. if people start complaining about this, we can consider making it more real-time.

thanks for thinking through this!

radazen:
> i would definitely prefer to handle the side-effect on user login to reduce calls to the DB. i don't think it's a huge deal that a user's gallery could show somewhat stale data. if people start complaining about this, we can consider making it more real-time.

My biggest concern with this approach is that the data could get _very_ stale, right? If I decide to sell off all my NFTs and become a tech-averse hermit, never logging into Gallery again...I feel like we'd want _some_ event to trigger the removal of those NFTs from my collections. Even though they're _my_ collections, other people may view and interact with them, so I don't think the update mechanism should be tied to some event that I alone can trigger. At the point where we're coming up with other ways to trigger this event, making sure we're triggering it often enough or at the right time...I feel like we're just adding unnecessary complexity.

With that in mind, I think I lean toward the first option. Seems like the overwhelmingly common case for doing this as we process each NFT would be "this doesn't belong to one of our users, early return," so the real overhead is probably closer to a single database lookup to see if the NFT warrants further processing. What kind of database access are we already doing on a per-NFT basis when we pull them in? If the overhead of an extra user lookup is significant at scale, we could also try building a cache of addresses owned by Gallery users for faster lookup.

radazen (approved):
Everything looks A-OK to me!

bennycio:
> Seems like the overwhelmingly common case for doing this as we process each NFT would be "this doesn't belong to one of our users, early return," so the real overhead is probably closer to a single database lookup to see if the NFT warrants further processing. What kind of database access are we already doing on a per-NFT basis when we pull them in? If the overhead of an extra user lookup is significant at scale, we could also try building a cache of addresses owned by Gallery users for faster lookup.
> 
When we receive NFTs we are not doing any sort of DB access right now until the upsert stage. You are right, the majority of cases will be they aren't a user so return early, and the DB call is pretty quick so this should be fine.

> [Edited to add] Do we currently check whether the "to" address is one of our users? Like, are we already doing a user lookup to do additional processing if the NFT is coming into our ecosystem?

We only check if "to" is one of our users in the step where we decide to grab media content for an NFT. If "to" address is a user, we do want to grab the media associated with the NFT.

radazen:
Ah, that makes sense. Thinking out loud: can we delay the "remove from collection" thing until the step where we grab media content? Maybe merge them into a single DB lookup that checks whether either address is associated with one of our users and then does media grabbing and collection-removal accordingly?

PR #300, opened 2022-03-01
cache galleries optimistically

bennycio:
Changes:

- **Added** caching for galleries in postgres


service/persist/postgres/gallery_nft.go
```diff
@@ -455,3 +480,23 @@ func addUnaccountedForCollections(pCtx context.Context, g *GalleryRepository, pU
 	}
 	return appendDifference(pColls, colls), nil
 }
+
+func (g *GalleryRepository) cacheByUserID(pCtx context.Context, pUserID persist.DBID) error {
+	err := g.RefreshCache(pCtx, pUserID)
```
Robinnnnn, line 485: how often do we refresh the cache?

Robinnnnn (approved):
woooo thanks for updating tests!

service/persist/postgres/gallery_nft.go
bennycio: Whenever there is an update to a cached item.

PR #301, opened 2022-03-05
GraphQL 3: Much TODO About Some Things

radazen:
**_NOTE TO REVIEWERS: Feel free to ignore generated files -- anything named `generated.go` or ending in `_gen.go`._**

**_Also, fair warning: this is a refactoring-heavy PR that has a whole lot of diff lines._**

## What's new?
- Added resolvers for all mutations defined in the schema
- Created a `publicapi` package to encapsulate our public API
- Mapped IDs and Addresses in GQL schema to backend types to reduce conversion boilerplate

## What's up with the PR title?
- There are TODOs everywhere. In the interest of getting a usable API up for people to work with, I've deferred some optimizations that we'll definitely want in place before we're relying on GraphQL in production. Some queries or mutations may be slower than they should be, may hit the database more than necessary, may do things inefficiently, etc.

## What's missing?
- `ErrInvalidInput` is declared in the schema to handle validation errors, but nothing is actually getting validated yet
- NFTs are still being returned as `GenericNft` and not as media types like `ImageNft`
- ~~Two fields are actually **not** resolved correctly at the moment: `DeleteCollectionPayload.Gallery` and `GalleryCollection.Gallery`. They'll return `null`.~~ [Edit: I just added another commit that enables these]

## Why all the refactoring?
To implement most of these mutations, I needed to call existing backend functionality used by our REST routes. That functionality lived in the `server` package, and attempting to reference it from the `graphql` package caused an import cycle (which is not allowed in golang). Something like this: `server.setUpARoute(server.someFunction)` is fine, but `server.setUpARoute(graphql.someFunction)` means that `graphql` can no longer reference any `server` methods without creating a cycle.

The best solution was to move the functionality shared by REST and GraphQL routes into a different package. Coincidentally, I was also trying to solve another problem at the same time: figuring out where to handle auth, input validation, and input sanitization. Best practices would suggest **not** delegating these to GraphQL, since it's the top layer of the request-handling stack and we'd only guarantee auth/validation/sanitization for requests coming in through GraphQL. If possible, they should live lower in the stack and be applied to any request -- even future requests that may not use GraphQL. Ideally, auth/validation/sanitization would also be handled in one location so you can immediately tell whether a particular point of entry is appropriately secured.

All of this led to the creation of `publicapi`: a package designed to consolidate user-facing endpoints into a single API that's easy to reason about. All exported functions in this package are entry points into backend functionality, and everything is available from a single object that can be passed to consumers of the API.

Nice things about the public API:
- It doesn't export anything that could be accidentally misused (like exporting a repository that consumers of the API could interact with directly, unknowingly skipping auth/validation/sanitization)
- It makes it easy to answer questions about auth/validation/sanitization without having to trace a bunch of code, inspect structs, or jump between files
- It's transport-agnostic and not tied to REST or GraphQL

Our GraphQL implementation is currently using the public API for mutation resolvers, and as I move more functionality into the public API, query resolvers will start using it too (and stop accessing repositories directly).


publicapi/collection.go
```diff
@@ -0,0 +1,117 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/mikeydub/go-gallery/graphql/dataloader"
+	"github.com/mikeydub/go-gallery/service/persist"
+	"github.com/mikeydub/go-gallery/service/pubsub"
+	"github.com/mikeydub/go-gallery/validate"
```
Robinnnnn, line 11: shouldn't these import from `/gallery-so/go-gallery` instead of `mikeydub`?!

publicapi/user.go
```diff
@@ -0,0 +1,55 @@
...
+	return user.RemoveAddressesFromUser(ctx, userID, addresses, api.repos.UserRepository)
+}
+
+func (api UserAPI) UpdateUserInfo(ctx context.Context, username string, bio string) error {
+	userID, err := getAuthenticatedUser(ctx)
+	if err != nil {
+		return err
+	}
+
+	// TODO: Validation of username and bio, sanitize one or both also
```
Robinnnnn, line 48: +1 to thwart sql injections haha

server/handler.go
```diff
@@ -26,29 +27,31 @@ func handlersInit(router *gin.Engine, repos *persist.Repositories, ethClient *et
...
-	if viper.GetString("ENV") != "production" {
-		// Keep graphql route out of prod while it's under development
-		parent.POST("/query", middleware.AddAuthToContext(), graphqlHandler(repos, ethClient))
+func graphqlHandlersInit(parent *gin.RouterGroup, repos *persist.Repositories, ethClient *ethclient.Client, pubsub pubsub.PubSub) {
+	parent.POST("/query", middleware.AddAuthToContext(), graphqlHandler(repos, ethClient, pubsub))
 
+	if viper.GetString("ENV") != "production" {
 		// TODO: Consider completely disabling introspection in production
 		parent.GET("/playground", graphqlPlaygroundHandler())
 	}
```
Robinnnnn, line 41: i think a blocker for officially switching over on prod is to set up some kind of logs (maybe datadog?) to get visibility into graphql requests, which are further obfuscated by doing SSR. in other words, it's been really easy to debug users' issues by popping open the browser network tab and inspecting the requests, and i'm afraid of losing this. @tbezman any thoughts here?

Robinnnnn (approved):
you're a freaking wizard. i say we merge this and do a round of bug-bashing alongside terence's PR on dev

kaitoo1 (approved):

kaitoo1:
+1 to what robin said.
looking forward to seeing this in action!!

publicapi/user.go
radazen: I should clarify that this already happens for our existing REST routes, just making a note for myself to handle in the public API layer!

PR #303, opened 2022-03-08
community endpoints

bennycio:
Changes:

- **Added** new table in DB for communities that is very similar to membership tiers
- **Added** persist code for getting and upserting communities
- **Added** package for updating community data
- **Added** endpoints for getting specific communities and all communities

Considerations:

- Do we want to be able to dynamically get data on any community or have a specific list of communities we want pages for? Right now, these endpoints work with a specific list we hard code but with a quick call to opensea we can get the necessary data to fill out data for any "community" (any opensea collection).

TODO:

- [ ] Tests
- [ ] GraphQL Resolver (maybe not required for this PR)

### DIFF
--- a/service/community/community.go
+++ b/service/community/community.go
@@ -17,9 +17,9 @@ type communityData struct {
 	TokenIDRanges   []persist.TokenIDRange
 }
 
-// right now these communities are hard coded, if we wanted to have this be dynamic we can get all of the information we need on a collection
+// Communities - right now these Communities are hard coded, if we wanted to have this be dynamic we can get all of the information we need on a collection
 // from opensea or the indexed NFTs (indexed NFTs won't have a banner image)
-var communities = map[persist.Address]communityData{
+var Communities = map[persist.Address]communityData{
 	// cryptocoven
 	"0x5180db8f5c931aae63c74266b211f580155ecac8": {
 		Name:            "CryptoCoven",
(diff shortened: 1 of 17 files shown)


service/community/community.go
```diff
@@ -0,0 +1,222 @@
...
+	for _, anN := range allNFTs {
+		user, err := userRepository.GetByAddress(pCtx, anN.OwnerAddress)
+		if err != nil {
+			continue
+		}
+		galleries, err := galleryRepository.GetByUserID(pCtx, user.ID)
+		if err != nil {
+			continue
+		}
+	outer:
```
Robinnnnn, line 80: holy crap, TIL you can do this

service/community/community.go
```diff
@@ -0,0 +1,222 @@
...
+	for i := 0; i < len(communities); i++ {
+		if err := <-done; err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// UpdateCommunity updates a community in the database
+func UpdateCommunity(pCtx context.Context, addr persist.Address, nftRepository persist.NFTRepository, userRepository persist.UserRepository, galleryRepository persist.GalleryRepository, communityRepository persist.CommunityRepository) error {
```
Robinnnnn, line 51: curious what the cost is of this operation. let me know if i have the steps straight:
1. look up all NFTs for a given contract address
2. select all owners of those NFTs (wallet address)
3. query all users who exist on gallery with that wallet address
4. get all of those users' galleries

it seems potentially a bit heavy, but adequate for the short run. ideally in the long-run we generate this list 1-time at the beginning, then from then on out we "listen" for these tokens moving around and being featured in galleries, and use that diff to add/remove users from the community list.

for the latter solution let's wait to have a better idea of a "feed" and "events" and such!

Robinnnnn (approved):

kaitoo1:
hey @bennycio , seconding Robin that we would want the list of community "members" to be any user that has that community's NFT in their gallery, rather than querying all the owners on the blockchain or thru opensea.

We're getting close to finalizing the specs for this feature, and they should be ready by early next week. Let's sync up then to go over the new requirements! We're going to have several iterations of the feature, so we can discuss the future direction as well. Thanks for getting a head start on this, I think we'll be able to leverage a lot of what you have here already

PR #304, opened 2022-03-08
Initial validation for public API functions

radazen:
Start doing validation and sanitization at the public API level. Still some TODOs to handle in a future pass, but this PR covers the basics.


Robinnnnn (approved):
🔥🔥

PR #307, opened 2022-03-09
Add discord bot

jarrel-b:
**Changes**
* Adds a feedbot app service which sends messages to discord, will follow up with a PR to add the deployment to dev,
* Adds an event module that handles the registering and dispatching of events to configured handlers. The handlers get registered and are added to the gin context similar to the publicapi pattern.
* Adds `Public*` interfaces to publicapi and added `*WithDispatch` versions of those interfaces which call the event handlers for the specific events we're interested in
* Adds new event schemas to persist and migrations
* Added a feed-event queue manually to the dev project

**Testing**
* Locally tested the `createCollection`, `updateCollectionInfo`, `updateCollectionNfts`, and `updateUserInfo` mutations manually and confirmed that tasks were created to a local task queue
* Need to do more testing with the feedbot, namely that it can accept tasks and that it gets routed correctly

**TODO**
- [x] Need to push image with new schemas to CircleCI before merging
- [x] Need to update server dev deployment with feedbot params before merging
- [x] Add documentation to Notion
- [ ] Add NFT events when more mutations get added


publicapi/user_dispatch.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/mikeydub/go-gallery/service/persist"
+)
+
+type UserWithDispatch struct {
+	PublicUserAPI
+	gc *gin.Context
+}
+
+// Forwarding methods to wrapped API
+func (u UserWithDispatch) UpdateUserInfo(ctx context.Context, username, bio string) error {
```
Robinnnnn, line 18: is this method set up to trigger at the right moment?

e.g. all user creations have empty `""` usernames until they subsequently set their username. so we would need to detect a change from unset => set, rather than triggering on _any_ user update

service/event/cloudtask/cloudtask.go
```diff
@@ -0,0 +1,71 @@
...
+
+func createTask(ctx context.Context, createdOn time.Time, eventID persist.DBID, eventCode persist.EventCode) error {
+	client, err := newClient(ctx)
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	queuePath := viper.GetString("GCLOUD_FEED_TASK_QUEUE")
+	scheduleOn := createdOn.Add(time.Duration(viper.GetInt("GCLOUD_FEED_TASK_BUFFER_SECS")) * time.Second)
```
Robinnnnn, line 43: 😍

service/persist/event.go
```diff
@@ -0,0 +1,24 @@
...
+
+type EventCode int16
+
+const (
+	UserEventCode = iota + 1
+	NftEventCode
+	CollectionEventCode
+)
+const (
+	UserCreatedEvent = (UserEventCode << 8) + iota + 1
```
Robinnnnn, line 11: 🔥

publicapi/user_dispatch.go
jarrel-b: Good question! I was torn between filtering events here vs. in the discord bot. Decided to leave it up to the receiver to handle the filtering. Was thinking an empty username might not be useful for discord, but a user update event in general, could be useful to another receiver in the future.

radazen:
Awesome stuff! I'm a big fan of the overall architecture of this. My only real question is about the PublicAPI dispatch wrappers. How would you feel about invoking the event stuff directly from the public API functions themselves, instead of wrapping/dispatching? Something like:

```
func (api CollectionAPI) UpdateCollection(ctx context.Context, collectionID persist.DBID, name string, collectorsNote string) error {
	// Validate
	if err := validateFields(api.validator, validationMap{
		"collectionID":   {collectionID, "required"},
		"name":           {name, "required,collection_name"},
		"collectorsNote": {collectorsNote, "required,collection_note"},
	}); err != nil {
		return err
	}

	// Sanitize
	name = validate.SanitizationPolicy.Sanitize(name)
	collectorsNote = validate.SanitizationPolicy.Sanitize(collectorsNote)

	userID, err := getAuthenticatedUser(ctx)
	if err != nil {
		return err
	}

	update := persist.CollectionUpdateInfoInput{
		Name:           persist.NullString(name),
		CollectorsNote: persist.NullString(collectorsNote),
	}

	err = api.repos.CollectionRepository.Update(ctx, collectionID, userID, update)
        if err != nil {
            return err
        }

        invokeEventHandlingHere()
        ^^^^^^^^^^^^^^^^^^^^^^^^^

        return nil
}
```

What I like about the above is that it's significantly less complex: no need to update multiple interface implementations when we change the public API, no need for a layer of indirection, less glue, etc. It doesn't separate concerns as well as your current implementation, but I think it'd be easier to maintain.

Alternatively, I can see a future where we can register handler functions with the public API to receive callbacks when various things happen. Same basic idea as the code snippet above, but `invokeEventHandlingHere` invokes any registered callbacks with whatever info/parameters would be useful to pass on to receivers. That'd separate concerns and be pretty easy to reuse for future features that might want to hook into API events, but imho we probably don't need to work on that kind of extensibility until we have multiple things that want to hook into the API.

jarrel-b:
> Awesome stuff! I'm a big fan of the overall architecture of this. My only real question is about the PublicAPI dispatch wrappers. How would you feel about invoking the event stuff directly from the public API functions themselves, instead of wrapping/dispatching? Something like:
> 
> ```
> func (api CollectionAPI) UpdateCollection(ctx context.Context, collectionID persist.DBID, name string, collectorsNote string) error {
> 	// Validate
> 	if err := validateFields(api.validator, validationMap{
> 		"collectionID":   {collectionID, "required"},
> 		"name":           {name, "required,collection_name"},
> 		"collectorsNote": {collectorsNote, "required,collection_note"},
> 	}); err != nil {
> 		return err
> 	}
> 
> 	// Sanitize
> 	name = validate.SanitizationPolicy.Sanitize(name)
> 	collectorsNote = validate.SanitizationPolicy.Sanitize(collectorsNote)
> 
> 	userID, err := getAuthenticatedUser(ctx)
> 	if err != nil {
> 		return err
> 	}
> 
> 	update := persist.CollectionUpdateInfoInput{
> 		Name:           persist.NullString(name),
> 		CollectorsNote: persist.NullString(collectorsNote),
> 	}
> 
> 	err = api.repos.CollectionRepository.Update(ctx, collectionID, userID, update)
>         if err != nil {
>             return err
>         }
> 
>         invokeEventHandlingHere()
>         ^^^^^^^^^^^^^^^^^^^^^^^^^
> 
>         return nil
> }
> ```
> 
> What I like about the above is that it's significantly less complex: no need to update multiple interface implementations when we change the public API, no need for a layer of indirection, less glue, etc. It doesn't separate concerns as well as your current implementation, but I think it'd be easier to maintain.
> 
> Alternatively, I can see a future where we can register handler functions with the public API to receive callbacks when various things happen. Same basic idea as the code snippet above, but `invokeEventHandlingHere` invokes any registered callbacks with whatever info/parameters would be useful to pass on to receivers. That'd separate concerns and be pretty easy to reuse for future features that might want to hook into API events, but imho we probably don't need to work on that kind of extensibility until we have multiple things that want to hook into the API.

Sounds good! Like keeping it as simple as possible.

publicapi/publicapi.go
```diff
@@ -26,6 +27,10 @@ type PublicAPI struct {
 	Nft        *NftAPI
 }
 
+type PublicNftAPI interface {
```
radazen, line 30: Is this left over from the wrapper dispatch stuff?

radazen (approved):
Looks great! Can't wait to see it in action.

kaitoo1 (approved):

kaitoo1:
Really awesome stuff @jarrel-b !! Super cool seeing the updates in the bot test channel and excited for this to go live.

I left a couple of clarification questions, mostly around the expected behavior. 
Code looks super clean, nice work , and thanks for adding the architecture diagram to notion btw!

publicapi/publicapi.go
jarrel-b: Good catch! removed

feedbot/nft.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Don't send with empty notes.
+	if event.Data.CollectorsNote == "" {
+		return nil
+	}
+
+	eventsSince, err := nftEventRepo.GetEventsSince(ctx, event, time.Now())
```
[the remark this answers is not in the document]
jarrel-b: yup spot on!

Robinnnnn (commented):
LFG

PR #308, opened 2022-03-09
cache changes to collections

bennycio:
Changes:

- **Added** cache refreshing to collection changes
- **Added** small fix to postgres token metadata to sanitize it for a specific bad character before putting it in the database. This bad character causes postgres to freak out and therefore the indexer to freak out.


Robinnnnn (approved):
this is awesome benny, thanks for the fix!

do you think this is an issue we could've caught using tests? curious how we can ensure that updates to a collection trigger an update to the gallery when it's fetched; we'd need to stand up some extra testing infra right?

bennycio:
> this is awesome benny, thanks for the fix!
> 
> do you think this is an issue we could've caught using tests? curious how we can ensure that updates to a collection trigger an update to the gallery when it's fetched; we'd need to stand up some extra testing infra right?

Yes, I definitely think there can be some tests for this! When I was working on caching originally the gallery tests were super helpful at getting me there but there are a lot more places in the application that affect what the get gallery output is. I will write some tests for updating collections and NFTs and how they affect the gallery output.

PR #313, opened 2022-03-11
GraphQL: Media type mapping and validation tweaks

radazen:
## What's new?
- GraphQL queries return a new `Nft` type that matches the format we expect to use when switching to the indexer. The fields on the `Nft` type are currently backed by Opensea data, but will be backed by indexer data in the future, and that switchover should (hopefully) be transparent to the frontend.
- `Nft` types now have a `media` field with for media-specific data
- A few validation tweaks for incoming request parameters

## Notes for implementors
- cc @tbezman 
- The `UnknownMedia` type is the default if the backend can't figure out what media type an NFT is. The default case in the existing frontend detection logic is `Animation`, and we probably want the frontend to handle unknown media the same way it's been handling `Animation` up to this point.
- `Media.mediaType` is nil right now. There's no v1 concept that would give the values we expect from the v2 API here.
- All `Media` types have a `previewURLs` field that returns preview images in various sizes. They get mapped from v1 concepts like this:
```
raw: image_original_url
small: image_thumbnail_url
medium: image_preview_url
large: image_url
```

- @Robinnnnn and I decided that `contentRenderingURL(s)` would be a pretty explicit way to name the set of URLs we want to use when actually rendering non-preview content. Every media type has one of these fields: `ImageMedia` has a set of URLs identical to the preview URLs list above, `VideoMedia` has a similar set of raw/small/medium/large values, and everything else just has a single `contentRenderingURL` field.
- **We may need to tweak the media detection and URL mapping logic.** The frontend logic for determining this stuff is fairly complex. If there are any cases where the backend isn't returning usable media information, we'll need to adjust accordingly.


Robinnnnn:
thanks so much for doing this @radazen and laying out the deets! 🔥 🔥 🔥

@tbezman whenever you get around to migrating the actual rendering of images / content feel free to hit me up and i'm happy to pair. the logic is def a bit wonky

Robinnnnn (approved):

graphql/plugin/gqlidgen/code_generator.gotpl
```diff
@@ -0,0 +1,63 @@
+{{ reserveImport "fmt" }}
```
Robinnnnn, line 1: 😍😍😍

graphql/plugin/gqlidgen/gqlidgen.go
```diff
@@ -0,0 +1,377 @@
...
+}
+
+func readTemplateFile(filename string) (string, error) {
+	_, callerFile, _, _ := runtime.Caller(1)
+	rootDir := filepath.Dir(callerFile)
+	templatePath := filepath.Join(rootDir, filename)
+
+	data, err := ioutil.ReadFile(templatePath)
+	return string(data), err
+}
```
Robinnnnn, line 377: wow this file is 🔥🔥🔥

radazen:
Realizing now that I should have said this before anyone started reviewing (cc @Robinnnnn): you can ignore the stuff in modelgen_custom. It's like 99.9% gqlgen and .1% a few small mods I made and copied to a new plugin instead of forking gqlgen itself.

Robinnnnn (approved):
you, ser, are a tour de force

kaitoo1:
@tbezman gave me the rundown on this pr and he was hyping up the changes here and he wasn't kidding this is extraordinary @radazen ! the code generation is super cool, it didn't occur to me that we could do something like that. super excited to use the standardized media types and asset urls on the front end too

PR #317, opened 2022-03-15
Update feed bot to filter events

jarrel-b:
This is a follow-up PR to have the feed bot compare the event to the most recent past event in order to determine if it should send a message or not.

Also adding an additional index to each table to allow for indexed queries when using the `last_updated` column.


kaitoo1 (approved):

Robinnnnn (approved):
awesome, thank you for this!

bennycio (approved):
Looks fantastic! Thanks so much!

PR #320, opened 2022-03-19
Add sentry to app

jarrel-b:
**Changes**
* Sentry setup on the backend:
    - panics in a resolver get routed by a hook provided by gqlgen
    - gin errors added to the error stack and then captured at the end of a request
    - remaining errors/panics outside of gin get handled at the top level
* Adds userID to the event and removes JWT  before sending to Sentry


cmd/server/main.go
```diff
@@ -2,14 +2,27 @@ package main
...
+	"time"
 
+	"github.com/getsentry/sentry-go"
 	"github.com/mikeydub/go-gallery/server"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/appengine"
 )
 
 func main() {
+	defer func() {
```
Robinnnnn, line 14: as a noob, mind explaining what this does?

server/server.go
```diff
@@ -164,3 +174,39 @@ func newGCPPubSub() pubsub.PubSub {
...
+		Dsn:              viper.GetString("SENTRY_DSN"),
+		Environment:      viper.GetString("ENV"),
+		AttachStacktrace: true,
+		BeforeSend: func(event *sentry.Event, _ *sentry.EventHint) *sentry.Event {
+			if event.Request == nil {
+				return event
+			}
+
+			var scrubbed []string
+			for _, c := range strings.Split(event.Request.Cookies, "; ") {
```
Robinnnnn, line 196: fire catch

Robinnnnn (approved):
looks really awesome to me! let's just wait on a review from someone else on backend!

cmd/server/main.go
jarrel-b: `defer` functions are called when the outer function returns. `recover()` is go's way of catching an exception, so this is essentially a `try...catch`. Here we catch a `panic`, capture it, and then re-raise it

graphql/resolver/schema.resolvers.helpers.go
```diff
@@ -40,10 +38,15 @@ func (r *Resolver) errorToGraphqlType(err error) (gqlError model.Error, ok bool)
 		mappedErr = model.ErrInvalidInput{Message: message, Parameters: validationErr.Parameters, Reasons: validationErr.Reasons}
 	}
 
-	if mappedErr != nil {
+	return mappedErr
+}
+
+// errorToGraphqlType converts a golang error to its matching type from our GraphQL schema.
```
kaitoo1, line 44: love the comments, much appreciated 👏

server/server.go
```diff
@@ -164,3 +174,39 @@ func newGCPPubSub() pubsub.PubSub {
 	client.CreateTopic(ctx, viper.GetString("ADD_ADDRESS_TOPIC"))
 	return client
 }
+
+func initSentry() {
+	if viper.GetString("ENV") == "local" {
+		log.Info("skipping sentry init")
```
kaitoo1, line 180: 🔥

cmd/server/main.go
```diff
@@ -2,14 +2,27 @@ package main
...
+		if err := recover(); err != nil {
+			sentry.CurrentHub().Recover(err)
```
radazen, line 16: If `sentry.CurrentHub()` returns a nil pointer for some reason, what will happen to the resulting nil reference error in `sentry.CurrentHub().Recover(err)`? Will it get re-raised automatically? I assume so, but I'm not too familiar with error recovery in go!

jarrel-b: Good question! Sentry creates an empty hub to start off so `.CurrentHub()` should never be `nil`: https://github.com/getsentry/sentry-go/blob/v0.13.0/hub.go#L29. 

But to answer your question, yup that would cause a runtime error and get re-raised

server/handler.go
```diff
@@ -49,7 +56,37 @@ func graphqlHandler(repos *persist.Repositories, ethClient *ethclient.Client, pu
 
 	h := handler.NewDefaultServer(generated.NewExecutableSchema(config))
 
+	h.SetRecoverFunc(func(ctx context.Context, err interface{}) error {
+		gc := util.GinContextFromContext(ctx)
+		if hub := sentrygin.GetHubFromContext(gc); hub != nil {
+			hub.Recover(err)
+		}
+
+		return gqlerror.Errorf("internal server error")
```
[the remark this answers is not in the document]
jarrel-b: Nice! Switched to using the default recover

graphql/resolver/schema.resolvers.helpers.go
```diff
@@ -18,11 +19,8 @@ import (
 
 var errNoAuthMechanismFound = fmt.Errorf("no auth mechanism found")
 
-// errorToGraphqlType converts a golang error to its matching type from our GraphQL schema.
-// If no matching type is found, ok will return false
-func (r *Resolver) errorToGraphqlType(err error) (gqlError model.Error, ok bool) {
+func GraphqlErrType(err error) (mappedErr model.Error) {
```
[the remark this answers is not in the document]
jarrel-b: Good call, removed the receiver and renamed it to `ErrorToGraphqlType`.

Robinnnnn (approved):

graphql/resolver/auth.go
```diff
@@ -31,8 +32,10 @@ func AuthRequiredDirectiveHandler(ethClient *ethclient.Client) func(ctx context.
 			switch authError {
 			case auth.ErrNoCookie:
 				gqlModel = model.ErrNoCookie{Message: errorMsg}
+				addError(ctx, authError, gqlModel)
```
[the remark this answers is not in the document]
radazen: I like the way you have it set up here!

radazen (approved):
Looks great! I did some local testing and can confirm that everything works as expected, too. Thanks @jarrel-b!

BRING ON THE SENTRY REPOOOOOOOOORRRRRTTTTTTTSSSS!

PR #321, opened 2022-03-21
do not clean strings

bennycio:
Changes:

- **Removed** clean string for `NullString` and `TokenMetadata` types so that non print characters can show up (such as new lines)


Robinnnnn (approved):
thanks so much benny!!

you think we could write a test for this so that we don't encounter this regression in the future? can be a separate PR so that we can push out this fix for now :)

PR #324, opened 2022-03-22
refresh metadata endpoint

bennycio:
Changes:

- **Added** an endpoint for refreshing a token's metadata (URI, JSON metadata, and Media)

TODO:

- [x] Tests

### DIFF
--- a/server/t__routes_token_test.go
+++ b/server/t__routes_token_test.go
@@ -65,6 +65,37 @@ func TestGetTokenByID_NotFoundError(t *testing.T) {
 	assert.NotEmpty(fmt.Sprintf("token not found by ID:: %s", nonexistentNftID), body.Error)
 }
 
+func TestRefreshMetadata_Success(t *testing.T) {
+	assert := setupTest(t, 2)
+
+	// seed DB with nft
+	name := "very cool nft"
+	token := persist.Token{
+		Name:            persist.NullString(name),
+		CollectorsNote:  "this is a bad note",
+		OwnerAddress:    tc.user1.address,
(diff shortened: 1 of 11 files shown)


server/t__routes_token_test.go
```diff
@@ -65,6 +65,37 @@ func TestGetTokenByID_NotFoundError(t *testing.T) {
 	assert.NotEmpty(fmt.Sprintf("token not found by ID:: %s", nonexistentNftID), body.Error)
 }
 
+func TestRefreshMetadata_Success(t *testing.T) {
```
Robinnnnn, line 68: lets gooo

service/membership/membership.go
```diff
@@ -266,30 +275,29 @@ func processCurrentTierToken(ctx context.Context, pTokenID persist.TokenID, ethC
...
+		addr := o
 		wp.Submit(func() {
-			logrus.Debugf("Processing event for ID %s %+v %d", id, event.ToAccount.Address, i)
-			if event.ToAccount.Address != persist.ZeroAddress {
+			logrus.Debugf("Processing event for ID %s %+v %d", id, addr, i)
+			if addr != persist.ZeroAddress {
 				logrus.Debug("Event is to real address")
 				// does to have the NFT?
-				membershipOwner := fillMembershipOwner(ctx, event.ToAccount.Address, id, ethClient, userRepository, galleryRepository)
+				membershipOwner := fillMembershipOwner(ctx, addr, id, ethClient, userRepository, galleryRepository)
```
Robinnnnn, line 300: non-blocking, but each time we call this we make a trip to the DB right? in the future do you think we could batch this into one big request to postgres?

Robinnnnn (approved):

PR #325, opened 2022-03-23
backups and opensea claiming

bennycio:
Changes:

- **Added** backup repo functions for getting backups and restoring backups
- **Added** tests for backup repo additions
- **Changed** claim NFTs step in opensea to not remove NFTs that are transferred between addresses

### DIFF
--- a/service/persist/postgres/backup.go
+++ b/service/persist/postgres/backup.go
@@ -87,7 +87,7 @@ func (b *BackupRepository) Insert(pCtx context.Context, pGallery persist.Gallery
 	}
 	defer res.Close()
 
-	currentBackups := map[persist.DBID]persist.CreationTime{}
+	currentBackups := []persist.Backup{}
 	for res.Next() {
 		var id persist.DBID
 		var createdAt persist.CreationTime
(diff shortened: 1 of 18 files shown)


service/persist/postgres/backup.go
```diff
@@ -41,33 +87,135 @@ func (b *BackupRepository) Insert(pCtx context.Context, pGallery persist.Gallery
...
+		currentBackups[id] = createdAt
+	}
+
+	if err = res.Err(); err != nil {
+		return err
+	}
+
+	doInsert := true
+	for id, createdAt := range currentBackups {
+		if time.Since(createdAt.Time()) > time.Hour*24*5 {
```
Robinnnnn, line 107: Will this delete backups over 5 days old?

Robinnnnn: If so I think we wanna keep them in case a user who hasn't updated their profile in a long time end up coming back later and have their gallery messed up

service/persist/postgres/collection_nft.go
```diff
@@ -332,33 +349,90 @@ func (c *CollectionRepository) ClaimNFTs(pCtx context.Context, pUserID persist.D
...
+			if removeOpenseaID == openseaID {
+				remove = false
+
+				var newAddress persist.NullString
+				err := c.getOwnerAddressStmt.QueryRowContext(pCtx, id).Scan(&newAddress)
+				if err != nil {
+					return err
+				}
+
+				_, err = deleteStmt.ExecContext(pCtx, id)
```
Robinnnnn, line 403: Will this line delete the nft?

Robinnnnn: Just figured this was the "don't delete the NFT" block lol

service/persist/postgres/backup.go
Robinnnnn: I think the logic should be that:
1) store up to 5 backups
2) if a user makes an update and it's been <12 hours since the last backup, ignore it
3) if a user makes an update and it's been >12 hours since the last backup, append it (and pop off the earliest backup if there are 5)
4) keep old backups as long as the user has 5 or less

service/persist/postgres/collection_nft.go
bennycio: This line deletes the NFT that we are joining with. Removes the duplicate basically.

Robinnnnn (approved):
incredible !!! 🔥 🔥 🔥

PR #328, opened 2022-03-25
Better error remapping

radazen:
## Better error remapping

We use a lot of union types with GraphQL to define expected response types. For example, looking up a user by username returns this union:
- `union UserByUsernameOrError = GalleryUser | ErrUserNotFound | ErrInvalidInput`

On the backend, we check golang errors to see if they have a corresponding GraphQL model type:

- `persist.ErrUserNotFound` (golang error type) -> `model.ErrUserNotFound` (GraphQL model type)

If the error does have a corresponding GraphQL model type, we use a typecast to see if the model type is allowed for the resolver being invoked:

- `if remapped, ok := gqlErrorModel.(model.UserByUsernameOrError)`

All of this works well, but up to this point it's required a lot of boilerplate to set up the appropriate typecast on a per-resolver basis. Every resolver that _could_ remap an error had to declare something like this:
```
// Map known errors to GraphQL return types
remapError := func(err error) (model.UserByUsernameOrError, error) {
    if errorType, ok := errorToGraphqlType(err); ok {
        if returnType, ok := errorType.(model.UserByUsernameOrError); ok {
            addError(ctx, err, errorType)
            return returnType, nil
        }
    }

    return nil, err
}
``` 

This PR adds a new gqlgen plugin that greatly simplifies the remapping process via code generation. The generated code maps a GraphQL type name to an appropriate typecast:

- `"UserByUsernameOrError"` -> `obj.UserByUsernameOrError`

And then we can use a gqlgen middleware to intercept the field and automatically map it to the correct output type if possible:

```
func RemapErrors(ctx context.Context, next gqlgen.Resolver) (res interface{}, err error) {
    res, err = next(ctx)

    if err == nil {
        return res, err
    }

    fc := gqlgen.GetFieldContext(ctx)
    typeName := fc.Field.Field.Definition.Type.NamedType

    // If a resolver returns an error that can be mapped to that resolver's expected GQL type,
    // remap it and return the appropriate GQL model instead of an error. This is common for
    // union types where the result could be an object or a set of errors.
    if fc.IsResolver {
        if remapped, ok := errorToGraphqlType(ctx, err, typeName); ok {
            return remapped, nil
        }
    }

    return res, err
}
```

Our resolvers no longer need to do _anything_ to handle error remapping. They can just return errors like normal, and the middleware function will take care of the remapping step before returning a response to the client.


server/handler.go
```diff
@@ -58,6 +57,8 @@ func graphqlHandler(repos *persist.Repositories, ethClient *ethclient.Client, ip
 	schema := generated.NewExecutableSchema(config)
 	h := handler.NewDefaultServer(schema)
 	h.AroundOperations(graphql.ScrubbedRequestLogger(schema.Schema()))
+	h.AroundFields(graphql.RemapErrors)
```
Robinnnnn, line 60: 🤯

Robinnnnn (approved):
🔥🔥🔥🔥🔥

graphql/plugin/remapgen/remapgen.go
```diff
@@ -0,0 +1,68 @@
...
+
+var (
+	_ plugin.ConfigMutator = &Plugin{}
+)
+
+func (m *Plugin) Name() string {
+	return "remapgen"
+}
+
+func (m *Plugin) MutateConfig(cfg *config.Config) error {
```
jarrel-b, line 37: Wow, this is really clever, amazed how you figured this out!

graphql/model/remapgen_gen.go
```diff
@@ -0,0 +1,120 @@
+// Code generated by github.com/99designs/gqlgen, DO NOT EDIT.
+
+package model
+
+var typeConversionMap = map[string]func(object interface{}) (objectAsType interface{}, ok bool){
```
jarrel-b, line 5: This is all generated from the plugin? Crazy. Was reading through the gqlgen docs and there's hardly any info on this feature haha

radazen: Yeah! It definitely took me some digging to figure out how to set up my own plugin, but it's pretty straightforward once you have everything set up. This whole file is generated from `remapgen.go` and `remapgen.gotpl`. The go file checks the GQL schema and stores union + interfaces in a slice, and then the template file loops through that slice and builds all of the mappings.

jarrel-b: ah `remapgen.gotpl` was the piece I was missing, this makes so much more sense now 👍

PR #330, opened 2022-03-28
Update gallery backup logic

Robinnnnn:
### Changes

Updated behavior:
- **Store up to 50 backups per gallery.** Will keep an eye on the size of this DB ballooning. Cool thing is if this is too frequent, we can simply update the pruning rules to slowly clear out the DB.
- **Reduce backup frequency from 12 hours to 5 mins**, if change was made in past 24h

New behavior:
- Backups in the past 7 days should be stored no more often than every 1 hour
- Backups beyond the past 7 days should be stored no more often than every 1 day

(See additional notes in files)

- [x] Automated tests
- [x] Manual testing in dev

### Manual Testing on Dev

Initial state:
Screen Shot 2022-03-29 at 2 14 53 AM

The latest backup inserted was at `03-29 06:11`. I made changes to my gallery at `06:12`, `06:13`, etc. and did not see any inserts to the DB. This was expected because the changes were within 5 minutes of the previous update ✅

I then made an update to a collection at `06:18`, 7 minutes after the last insert:
Screen Shot 2022-03-29 at 2 20 02 AM

The `06:18` backup was inserted. ✅

The previous `06:11` backup remained intact, since it had been >5 minutes since the last update. ✅

All but one entry from `03-24` were pruned, since the changes occurred within 1 hour of each other. Specifically, the latest one was kept. ✅

### DIFF
--- a/service/persist/postgres/backup.go
+++ b/service/persist/postgres/backup.go
@@ -120,45 +120,38 @@ func (b *BackupRepository) Insert(pCtx context.Context, pGallery persist.Gallery
 		}
 	}
 
-	// prune backups
-	if len(currentBackups) > 0 {
+	if len(currentBackups) >= 2 {
 		day := time.Hour * 24
 		week := day * 7
 
-		prev := currentBackups[0]
-		prevCreationTime := prev.CreationTime.Time()
-		for i := 1; i < len(currentBackups); i++ {
-			curr := currentBackups[i]
-			currCreationTime := curr.CreationTime.Time()
+		curr := currentBackups[len(currentBackups)-1]
+		currCreationTime := curr.CreationTime.Time()
 
-			// if two backups are over a week old but within a day apart, keep the older one
-			if time.Since(prevCreationTime) > week &&
-				time.Since(currCreationTime) > week &&
-				currCreationTime.Sub(prevCreationTime) < day {
-				b.deleteBackupStmt.ExecContext(pCtx, curr.ID)
+		// iterate from latest to oldest backup
+		for i := len(currentBackups) - 2; i >= 0; i-- {
+			prev := currentBackups[i]
+			prevCreationTime := prev.CreationTime.Time()
+
+			// backups in the past day and are within 5 mins of each other
+			updatedInPastDay := time.Since(currCreationTime) <= day &&
+				currCreationTime.Sub(prevCreationTime) < 5*time.Minute
+			// backups in the past week and are within 1 hour of each other
+			updatedInPastWeek := time.Since(currCreationTime) > day &&
+				time.Since(currCreationTime) <= week &&
+				currCreationTime.Sub(prevCreationTime) < time.Hour
+			// backups older than 1 week and are within 1 day of each other
+			updatedOverWeekAgo := time.Since(currCreationTime) > week &&
+				currCreationTime.Sub(prevCreationTime) < day
+
+			if updatedInPastDay || updatedInPastWeek || updatedOverWeekAgo {
+				b.deleteBackupStmt.ExecContext(pCtx, prev.ID)
+				// continue statement here ensures that `curr` remains anchored
+				// for the next comparison, avoiding a cascade
 				continue
 			}
 
-			// if two backups are under a week old, over a day old, and within an hour apart, keep the older one
-			if time.Since(prevCreationTime) < week &&
-				time.Since(currCreationTime) < week &&
-				time.Since(prevCreationTime) > day &&
-				time.Since(currCreationTime) > day &&
-				currCreationTime.Sub(prevCreationTime) < time.Hour {
-				b.deleteBackupStmt.ExecContext(pCtx, curr.ID)
-				continue
-			}
-
-			// if two backups are within a day old but within 5 minutes apart, keep the older one
-			if time.Since(prevCreationTime) < day &&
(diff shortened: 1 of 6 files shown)


service/persist/postgres/backup.go
```diff
@@ -47,7 +49,7 @@ func NewBackupRepository(db *sql.DB) *BackupRepository {
 	deleteBackupStmt, err := db.PrepareContext(ctx, `DELETE FROM backups WHERE ID = $1;`)
 	checkNoErr(err)
 
-	insertBackupStmt, err := db.PrepareContext(ctx, `INSERT INTO backups (ID, GALLERY_ID, VERSION, GALLERY) VALUES ($1, $2, $3, $4);`)
+	insertBackupStmt, err := db.PrepareContext(ctx, `INSERT INTO backups (ID, GALLERY_ID, VERSION, GALLERY, CREATED_AT) VALUES ($1, $2, $3, $4, $5);`)
```
Robinnnnn, line 52: require creation time to be provided when inserting backup. this allows us to insert backups at arbitrary times to test our pruning logic.

- [x] ensure that manually setting a timestamp (new behavior) follows the same format as the default timestamp (previous behavior)

image

service/persist/postgres/t__backup_test.go
```diff
@@ -3,73 +3,16 @@ package postgres
...
-	"github.com/mikeydub/go-gallery/service/memstore/redis"
 	"github.com/mikeydub/go-gallery/service/persist"
-	"github.com/sirupsen/logrus"
 )
 
 func TestBackupRestore_Success(t *testing.T) {
 	a, db := setupTest(t)
+	userID, collID, nftIDs, g, backupRepo, galleryRepo, collectionRepo, nftRepo, _ := createMockGallery(t, a, db)
 
-	galleryRepo := NewGalleryRepository(db, redis.NewCache(0))
```
Robinnnnn, line 15: removed a lot of boilerplate to a test util function

service/persist/postgres/t__backup_test.go
```diff
@@ -107,64 +50,48 @@ func TestBackupRestore_Success(t *testing.T) {
 	a.Len(galleries[0].Collections, 1)
 }
 
-func TestOneBackupADay_Success(t *testing.T) {
+// backups made within 5 mins of each other should be de-duped
+func TestThrottle_Success(t *testing.T) {
```
Robinnnnn, line 54: no new behavior in this test, can ignore

service/persist/postgres/backup.go
```diff
@@ -102,21 +104,62 @@ func (b *BackupRepository) Insert(pCtx context.Context, pGallery persist.Gallery
 		return err
 	}
 
+	// skip insert if we've created a backup very recently
```
kaitoo1, line 107: thanks for all the comments 👏 helps a lot

service/persist/postgres/t__utils_test.go
```diff
@@ -29,3 +33,60 @@ func setupTest(t *testing.T) (*assert.Assertions, *sql.DB) {
 
 	return assert.New(t), db
 }
+
+func createMockGallery(t *testing.T, a *assert.Assertions, db *sql.DB) (persist.DBID, persist.DBID, []persist.DBID, persist.Gallery, *BackupRepository, *GalleryRepository, *CollectionRepository, *NFTRepository, *UserRepository) {
```
kaitoo1, line 37: niceeee

service/persist/postgres/t__backup_test.go
```diff
@@ -175,22 +102,100 @@ func TestOneBackupADay_Success(t *testing.T) {
...
+	backupRepo.insertBackupStmt.ExecContext(context.Background(), persist.GenerateID(), g.ID, g.Version, g, persist.CreationTime(time.Now().Local().Add(time.Hour)))
 
 	backups, err = backupRepo.Get(context.Background(), userID)
 	a.NoError(err)
-	a.Len(backups, 1)
+	a.Len(backups, 2)
 	a.Len(backups[0].Gallery.Collections, 1)
 }
+
+// Backup pruning rules:
```
kaitoo1, line 124: 🙌🙌🙌

radazen (commented):
Looks great! Thanks for all the thorough testing.

One additional thought: instead of enforcing a maximum number of backups by deleting the oldest when adding a new one, can we use pruning to enforce the maximum number of backups, and stop pruning when the user is below the max?

I'm thinking of the scenario where a user runs into trouble and reports it -- so far so good, we have lots of 5m backups near the issue -- but then we're on a 24h clock to try and use the backups before they're at risk of being pruned (down to 1h backups) the next time the user makes a change.

I'm thinking something like this:
- Add a backup
- Are we over max backups? If so, start pruning
- If pruning gets us below max backups, stop pruning
- If we finish pruning and we're still over the maximum, delete the oldest N backups to get us down to the maximum

Then, ideally, we'd only start to lose the 5m snapshots if the user continues to make lots of updates to their gallery after the issue occurs...and if that's the case, I'm not sure a backup is going to help them since they've got all sorts of new changes.

Robinnnnn:
@kaitoo1 @radazen thanks so much for your awesome comments!

i sync'd with ezra and let him know that i'll be tackling these in a follow-up PR since my day is pretty packed, and this at least handles a number of low-hanging fruits.

PR #331, opened 2022-03-29
Don't select deleted NFTs when building a list of NFTs to delete

radazen:
When reconciling user NFTs with OpenSea, we build a list of NFTs that are pending removal. An NFT in that list will be removed **unless** we also have a new NFT from OpenSea with the same OpenSea ID -- essentially, if the same NFT is being deleted and added at the same time, we keep the older (pending deletion) version because its Gallery ID is the one we'd have used in existing collections.

An issue arises if the older NFT is already deleted, however, since we'd be opting to keep a deleted version of the NFT instead of keeping the newer, non-deleted version.

This PR addresses the issue by ignoring already-deleted NFTs when building a list of NFTs to delete.

--

Other changes in here:
- Don't error on empty wallets during OpenSea refresh, but don't filter them out of asset reconciliation either. Otherwise, NFTs that _were_ in a now-empty wallet aren't considered when updating IDs during a sync.
- Since we're not hard-deleting NFTs anymore, it's easy to violate the (wallet_address, opensea_id) uniqueness constraint when multiple deleted NFTs have the same owner and OpenSea ID. This PR fixes the issue by relaxing the uniqueness constraint to only be enforced for non-deleted NFTs. The corresponding DB migration (to update the index) will need to be run against dev and prod.


Robinnnnn (approved):
thank you ezra!!

PR #332, opened 2022-03-29
backup before refresh, ensure restore deleted NFTs, ensure no double deletes

bennycio:
Changes:

- **Added** an undelete step for NFTs that were deleted when restoring backups
- **Added** a backup before opensea refresh
- **Changed** collection getting to not get deleted NFTs so that we don't continually retrieve deleted NFTs and re-delete them
- **Changed** indexer transfers step to run side-effects for ERC-721 as well (small fix)


indexer/indexer.go
```diff
@@ -590,6 +590,13 @@ func findFields(i *Indexer, transfer rpc.Transfer, key persist.TokenIdentifiers,
 			previousOwners <- prevOwner
 		}()
 
+		if sideEffects {
```
radazen, line 593: Is this an unrelated change that managed to slip in? 🥷

indexer/indexer.go
```diff
@@ -730,7 +737,11 @@ func updateCollections(ctx context.Context, contractAddress persist.Address, tok
 		didUpdate := false
 		for i, nft := range coll.NFTs {
 			if nft.ContractAddress.String() == contractAddress.String() && nft.TokenID.String() == tokenID.String() {
-				coll.NFTs = append(coll.NFTs[:i], coll.NFTs[i+1:]...)
+				if i+1 > len(coll.NFTs) {
```
radazen, line 740: This is a tricky one. Given that we're modifying a slice while iterating over it, I believe you'll actually end up skipping an entry whenever there's a match. I threw together some test code because I wasn't sure how golang handles this sort of thing. C# would throw an exception if you modify a collection while iterating over it, but go seems happy to let you do it. The result is a little surprising, though:

```
entries := []string{"a", "b", "c"}
for i, entry := range entries {
    fmt.Printf("index: %d, entry: %s\n", i, entry)
    if entry == "a" || entry == "b" {
        if i+1 > len(entries) {
            entries = entries[:i]
        } else {
            entries = append(entries[:i], entries[i+1:]...)
        }
    }
}

fmt.Printf("entries: %v\n", entries)
```

You'd think it would filter out both `a` and `b`, but it actually only filters out `a` because `b` gets skipped when the array is modified:

```
Output:
index: 0, entry: a
index: 1, entry: c
index: 2, entry: c
entries: [b c]
```

I was actually just tweaking this in a different PR, and I ended up using the technique recommended here: https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating

service/persist/postgres/collection_nft.go
```diff
@@ -57,7 +57,7 @@ func NewCollectionRepository(db *sql.DB, galleryRepo *GalleryRepository) *Collec
 	n.MULTIPLE_OWNERS,n.NAME,n.CONTRACT,n.TOKEN_COLLECTION_NAME,n.CREATOR_ADDRESS,n.CREATOR_NAME, 
 	n.IMAGE_URL,n.IMAGE_THUMBNAIL_URL,n.IMAGE_PREVIEW_URL,n.ANIMATION_ORIGINAL_URL,n.ANIMATION_URL,n.CREATED_AT 
 	FROM collections c, unnest(c.NFTS) WITH ORDINALITY AS u(nft, ordinality)
-	LEFT JOIN nfts n ON n.ID = nft
+	LEFT JOIN nfts n ON n.ID = nft AND n.DELETED = false
```
radazen, line 60: I think you want `AND n.DELETED = false` in the WHERE clause instead of the JOIN condition. If it's in the join, you'll still have a row for this NFT (from the `unnest(c.NFTS)`), it just won't have any of the columns from the NFT table filled out. If you move it to the WHERE clause, the row will be omitted from the output entirely.*

\* - I only know this because I was experimenting with these sorts of statements last week. 😅

service/persist/postgres/collection_nft.go
```diff
@@ -69,7 +69,7 @@ func NewCollectionRepository(db *sql.DB, galleryRepo *GalleryRepository) *Collec
 	n.MULTIPLE_OWNERS,n.NAME,n.CONTRACT,n.TOKEN_COLLECTION_NAME,n.CREATOR_ADDRESS,n.CREATOR_NAME, 
 	n.IMAGE_URL,n.IMAGE_THUMBNAIL_URL,n.IMAGE_PREVIEW_URL,n.ANIMATION_ORIGINAL_URL,n.ANIMATION_URL,n.CREATED_AT 
 	FROM collections c,unnest(c.NFTS) WITH ORDINALITY AS u(nft, ordinality) 
-	LEFT JOIN nfts n ON n.ID = nft 
+	LEFT JOIN nfts n ON n.ID = nft AND n.DELETED = false
```
radazen, line 72: Same as above re: JOIN vs WHERE

service/persist/postgres/collection_nft.go
```diff
@@ -80,7 +80,7 @@ func NewCollectionRepository(db *sql.DB, galleryRepo *GalleryRepository) *Collec
 	n.MULTIPLE_OWNERS,n.NAME,n.CONTRACT,n.TOKEN_COLLECTION_NAME,n.CREATOR_ADDRESS,n.CREATOR_NAME, 
 	n.IMAGE_URL,n.IMAGE_THUMBNAIL_URL,n.IMAGE_PREVIEW_URL,n.ANIMATION_ORIGINAL_URL,n.ANIMATION_URL,n.CREATED_AT 
 	FROM collections c, unnest(c.NFTS) WITH ORDINALITY AS u(nft, ordinality)
-	LEFT JOIN nfts n ON n.ID = nft
+	LEFT JOIN nfts n ON n.ID = nft AND n.DELETED = false
```
radazen, line 83: Same as above re: JOIN vs WHERE

service/persist/postgres/collection_nft.go
```diff
@@ -106,7 +106,7 @@ func NewCollectionRepository(db *sql.DB, galleryRepo *GalleryRepository) *Collec
 	n.MULTIPLE_OWNERS,n.NAME,n.CONTRACT,n.TOKEN_COLLECTION_NAME,n.CREATOR_ADDRESS,n.CREATOR_NAME,
 	n.IMAGE_URL,n.IMAGE_THUMBNAIL_URL,n.IMAGE_PREVIEW_URL,n.ANIMATION_ORIGINAL_URL,n.ANIMATION_URL,n.CREATED_AT 
 	FROM collections c, unnest(c.NFTS) WITH ORDINALITY AS u(nft, ordinality)
-	LEFT JOIN nfts n ON n.ID = nft
+	LEFT JOIN nfts n ON n.ID = nft AND n.DELETED = false
```
radazen, line 109: Same as above re: JOIN vs WHERE

service/persist/postgres/collection_token.go
```diff
@@ -52,7 +52,7 @@ func NewCollectionTokenRepository(db *sql.DB, galleryRepo *GalleryTokenRepositor
 		c.LAYOUT,c.HIDDEN,c.CREATED_AT,c.LAST_UPDATED,
 		n.ID,n.OWNER_ADDRESS,n.CHAIN,n.NAME,n.DESCRIPTION,n.TOKEN_TYPE,n.TOKEN_URI,n.TOKEN_ID,n.MEDIA,n.TOKEN_METADATA,n.CONTRACT_ADDRESS,n.CREATED_AT 
 		FROM collections_v2 c, unnest(c.NFTS) WITH ORDINALITY AS u(nft, ordinality) 
-		JOIN tokens n ON n.ID = nft 
+		JOIN tokens n ON n.ID = nft AND n.DELETED = false
```
radazen, line 55: Same as above re: JOIN vs WHERE

service/persist/postgres/collection_token.go
```diff
@@ -62,7 +62,7 @@ func NewCollectionTokenRepository(db *sql.DB, galleryRepo *GalleryTokenRepositor
 	c.LAYOUT,c.HIDDEN,c.CREATED_AT,c.LAST_UPDATED,
 	n.ID,n.OWNER_ADDRESS,n.CHAIN,n.NAME,n.DESCRIPTION,n.TOKEN_TYPE,n.TOKEN_URI,n.TOKEN_ID,n.MEDIA,n.TOKEN_METADATA,n.CONTRACT_ADDRESS,n.CREATED_AT 
 	FROM collections_v2 c, unnest(c.NFTS) WITH ORDINALITY AS u(nft, ordinality) 
-	JOIN tokens n ON n.ID = nft
+	JOIN tokens n ON n.ID = nft AND n.DELETED = false
```
radazen, line 65: Same as above re: JOIN vs WHERE

radazen (commented):
🔥🔥🔥 Looking good, Benny! I left a few comments.

service/persist/postgres/collection_nft.go
bennycio: Amazing! I kept putting this back and forth between the where and join clause when I was working on this and noticed exactly what you are talking about. Having this in the where clause will save us some iterations 🎉

Robinnnnn (approved):

PR #334, opened 2022-03-31
validate new address NFTs

bennycio:
Changes:

- **Changed** create user and add address funcs for users in v2 endpoints to ensure that any new addresses call the indexer validate functions to ensure they have all of their NFTs as well as have all of the media content for their NFTs

### DIFF
--- a/publicapi/user.go
+++ b/publicapi/user.go
@@ -44,7 +44,7 @@ func (api UserAPI) AddUserAddress(ctx context.Context, address persist.Address, 
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


bennycio:
Small note: We used to have the main backend call endpoints on the indexer. I think it is better to reuse the code instead of calling an indexer endpoint because the indexer has enough to listen to already from the blockchain, it should not have to listen to thousands of requests from the backend as well. I think the computational energy used to validate NFTs and their media content for a new user is better put on the service that is creating that new user. I would love to hear opinions on this, however.

PR #336, opened 2022-03-31
Clear cookie if getCurrentUser returns a 204

Robinnnnn:


radazen (approved):
Pushed a small tweak, but honestly not sure if my change is worthwhile or not. Without the change, we'd try to clear the user's cookie even if the reason they're not authed is that they don't have a cookie...but now that I've pushed it, I'm honestly not sure it's a meaningful distinction. 🤷‍♂️

In any case, LGTM!

radazen:
And thanks!

PR #339, opened 2022-04-06
GraphQL optimization + organization

radazen:
### sqlc
- Added an initial [sqlc](https://sqlc.dev) implementation
- sqlc generates structs and query methods based on raw SQL statements
- Currently using `sqlc` as a namespace to make it obvious what's coming from `persist` and what's coming from `sqlc`
- sqlc is using the native `pgx` Postgres driver, which has significant performance benefits over using pgx in `database/sql` compatibility mode
- This is a work in progress, and I'll probably be reorganizing it in future PRs
- **Note:** sqlc is using the native `pgx` driver, but the majority of the backend is still using `database/sql` via pgx in compatibility mode. Consequently, we're setting up _two_ database clients during initialization right now. This shouldn't be an issue in practice, but at some point we should move everything over to native pgx.

### Dataloaders
- Dataloaders now handle query batching in addition to caching (which they already supported)
- Batching is enabled via sqlc's support for pgx query batching
- Concrete example of what this means: instead of doing multiple round trips to the database to look up each of a user's collections, our dataloaders will wait for about 2ms to aggregate all similar queries, and then send them to the database in a single round trip via batching

### GraphQL
- Schema now supports the `node(id: ID!)` root query to look up any `Node` by its ID
- `MembershipOwner` types no longer implement the `Node` interface because refetching them by ID is a pain and they aren't well-suited to being nodes


db/queries/query.sql
```diff
@@ -0,0 +1,83 @@
...
+SELECT * FROM collections WHERE id = $1 AND deleted = false;
+
+-- name: GetCollectionByIdBatch :batchone
+SELECT * FROM collections WHERE id = $1 AND deleted = false;
+
+-- name: GetCollectionsByGalleryId :many
+SELECT c.* FROM galleries g, unnest(g.collections)
+    WITH ORDINALITY AS x(coll_id, coll_ord)
+    INNER JOIN collections c ON c.id = x.coll_id
+    WHERE g.id = $1 AND g.deleted = false AND c.deleted = false ORDER BY x.coll_ord;
```
bennycio, line 47: I noticed there are no queries that do both joins to Collections and NFTs from the top level Gallery. I also noticed there was a TODO comment in a file further on mentioning that we aren't going to always fill in the information for a gallery. Just to clarify, is this going to be the new process for getting the NFTs for the top level Gallery view:

 1. Get all collections in the gallery.
 2. For each collection in the gallery, iteratively get all NFTs in the collection but only the fields we need (image, name, etc.)

db/sqlc/models.go
```diff
@@ -0,0 +1,8 @@
+package sqlc
+
+// TokenLayout defines the layout of a collection of tokens
+type TokenLayout struct {
+	Columns    int   `json:"columns"`
+	Whitespace []int `json:"whitespace"`
+	// Padding         int   `bson:"padding" json:"padding"`
```
Robinnnnn, line 7: what's padding for?

graphql/generated/generated.go
```diff
@@ -6496,6 +6482,45 @@ func (ec *executionContext) _PreviewURLSet_large(ctx context.Context, field grap
...
+	if err != nil {
+		ec.Error(ctx, err)
+		return graphql.Null
+	}
+	if resTmp == nil {
+		return graphql.Null
+	}
+	res := resTmp.(model.Node)
+	fc.Result = res
+	return ec.marshalONode2githubᚗcomᚋmikeydubᚋgoᚑgalleryᚋgraphqlᚋmodelᚐNode(ctx, field.Selections, res)
```
Robinnnnn, line 6521: woah what's going on with these characters

image

Robinnnnn (approved):

graphql/generated/generated.go
radazen: Haha, that's a gqlgen thing. Not sure why they use those characters! I read something related in their changelog once, but I don't remember the reasoning.

radazen:
> MONSTER PRRR!!! 🔥
> 
> > at some point we should move everything over to native pgx
> 
> does this essentially translate to "at some point we should use sqlc everywhere"?

At some point I would indeed like to use sqlc everywhere, though I think it's probably much easier to switch existing queries over to native pgx as a first step!

db/queries/query.sql
radazen: Pretty close, yes! You are correct; there are no lookups that join gallery/collection/nft or collection/nft to fill out a hierarchy. Getting a top level gallery view looks like this:
- Look up the gallery in the DB
- Look up the collections
- Look up the NFTs

GraphQL tends to be modular in this way. You only try to look up the data you need, so by default, querying for the gallery doesn't look up any collections or NFTs. When we do need collections and NFTs, there are extra round trips involved. The benefit of this approach is that it keeps things very modular and keeps the resolvers themselves simple and maintainable. GraphQL also parallelizes requests where possible, and the dataloaders do batching, so a gallery with lots of collections would do something like:

```
DB request 1: look up this batch of 10 collections' NFTs
DB request 2: look up another batch of 10 collections' NFTs
DB request 3: look up another batch of 10 collections' NFTs

Now execute all 3 requests in parallel
```

In practice, this scheme is surprisingly fast! While we could write additional queries to fill out collections/NFTs in a single round trip, I don't think it's worth the additional complexity in our resolvers atm. I suspect we'll want to start using redis with GraphQL at some point in the near future, though, and I think that would be a good opportunity to avoid round trips where possible.

> 2. For each collection in the gallery, iteratively get all NFTs in the collection but only the fields we need (image, name, etc.)

For something like an NFT, we'll still pull an entire row from the database, but GraphQL will only return the pieces the client asked for.

bennycio: So cool!!! Thanks for the detailed explanation!

PR #341, opened 2022-04-07
community endpoint

bennycio:
Changes:

- **Added** endpoint for getting communities by address
- **Added** persist functions for caching and grabbing information on a community


service/persist/community.go
```diff
@@ -0,0 +1,27 @@
+package persist
+
+import "context"
+
+// Community represents a community and is only cached (has no ID)
```
Robinnnnn, line 5: is there a reason we're only caching? is it because it's unnecessary to store, since we always infer it from opensea/indexer?

db/migrations/000001_create_core_tables.up.sql
```diff
@@ -132,6 +133,7 @@ CREATE TABLE IF NOT EXISTS contracts (
     NAME varchar,
     SYMBOL varchar,
     ADDRESS varchar(255),
+    CREATOR_ADDRESS varchar(255),
```
Robinnnnn, line 136: how do we ensure this field gets added/migrated in prod?

service/persist/community.go
bennycio: Exactly! Since the data for a community is always just conglomerated from other parts of the DB, if we stored it in the DB it would just be like using postgres as a cache. I did not think it would be necessary given that this data was pretty easily accessible internally and it is not like we could get rate limited and not have access to any data like for opensea.

db/migrations/000001_create_core_tables.up.sql
bennycio: We can do an update table, add column statement. Since there are no creator addresses that exist right now (and we aren't retrieving them yet), there is nothing to migrate, so we might as well just add the column. I can do this before we push this to production.

radazen: I _think_ the ideal way to do this with the migration stuff Jarrel set up is to create a new sequence file that adds the column, and then use the `migrate` command to bump the dev/prod databases up to the new migration level. I did it last week when modifying an index, and it worked very smoothly!* Migration tool details are in the `go-gallery` README.

\* - ...well, it worked smoothly after I updated to the branch where my new migration file was.

service/persist/community.go
```diff
@@ -0,0 +1,27 @@
...
+type Community struct {
+	LastUpdated LastUpdatedTime `json:"last_updated"`
+
+	ContractAddress Address `json:"contract_address"`
+	CreatorAddress  Address
```
kaitoo1, line 10: should this have `json:"creator_address"` ?

server/community.go
```diff
@@ -0,0 +1,35 @@
...
+			return
+		}
+
+		community, err := communityRepo.GetByAddress(c, input.ContractAddress)
+		if err != nil {
+			util.ErrResponse(c, http.StatusInternalServerError, err)
+			return
+		}
+
+		c.JSON(http.StatusOK, getCommunityOutput{Community: community})
```
kaitoo1, line 33: could we return a not found error if a community isn't found for a given address?

kaitoo1: (though if we're going to have the graphql resolver soon we can just return not found for that, and leave the rest endpoint as is )

service/persist/community.go
bennycio: yes it should!

server/community.go
kaitoo1: thanks!

kaitoo1 (approved):

Robinnnnn (approved):

PR #343, opened 2022-04-07
Jarrel/use dockertest for test

jarrel-b:
**Changes**
Unit tests start a new container per test file (the tests were super flaky if containers were started each test), instead of using the local containers and volumes for testing. The `setupDoubles` fixture is used to set up the containers and tear them down and patches the environment. The `setupTest` fixture sets up the server and wipes the tables after each unit test. 

Also moved some functions previously in `server` to `docker` and `db` so they can be re-used.

**Testing**
Ran `go test -short ./...`

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -115,16 +115,15 @@ func getBuildImage(path string, s Service) ([]string, error) {
 	return nil, errors.New("no `FROM` directive found in dockerfile")
 }
 
-func InitPostgres(composePath string) *dockertest.Resource {
+func InitPostgres(composePath string) (resource *dockertest.Resource, callback func()) {
 	pool, err := dockertest.NewPool("")
 	pool.MaxWait = 3 * time.Minute
 	if err != nil {
 		log.Fatalf("could not connect to docker: %s", err)
 	}
 
-	absPath, _ := filepath.Abs(composePath)
-	apps := loadComposeFile(absPath)
-	imgAndVer, err := getBuildImage(absPath, apps.Services["postgres"])
+	apps := loadComposeFile(composePath)
+	imgAndVer, err := getBuildImage(composePath, apps.Services["postgres"])
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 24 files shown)


docker/docker.go
```diff
@@ -0,0 +1,189 @@
...
+	for _, cmd := range res {
+		if cmd.Cmd == "FROM" {
+			return getImageAndVersion(cmd.Value[0])
+		}
+	}
+
+	return nil, errors.New("no `FROM` directive found in dockerfile")
+}
+
+func InitPostgres(composePath string) *dockertest.Resource {
```
radazen, line 118: How would you feel about getting rid of the `composePath` parameter and just defaulting to the `docker-compose.yml` file at the project root? Or splitting it into two functions so we have an `InitPostgres` function that takes no parameters and calls `InitPostgresWithComposePath` or something with the default docker-compose file location? I like having the path be configurable, but it seems like all uses of this function pass the same `../../../docker-compose.yml` style parameter, and each of them is responsible for figuring out the correct level of `../..` based on how deeply the particular test file is nested.

radazen (approved):
I love everything about this! 🎉🎉🎉 Left one small comment re: docker-compose.yml path.

🙌🙌🙌🙌🙌🙌🙌🙌🙌🙌🙌🙌🙌🙌

Robinnnnn (approved):

Robinnnnn (approved):

docker/docker.go
radazen: Ah, that makes sense. I tried some googling to see if there's an easy way to get the project root, but it doesn't seem like there's a reliable way to do it. Maybe check for `./docker-compose.yml`, `../docker-compose.yml`, `../../docker-compose.yml`, etc until we find a file or hit some depth threshold? I'm mostly imagining myself trying to figure out why tests are broken some day when I refactor stuff and move files around.

jarrel-b: ooo like this idea, added this for finding `docker-compose.yml` and also for finding the migration files to seed the containers.

radazen (approved):
Yup, this is awesome.

PR #344, opened 2022-04-08
GraphQL: updateNftInfo / extra NFT fields / Dead Ringers

radazen:
## What's new?

- updateNftInfo mutation allows modifying an NFT's collector's note
- creatorAddress and openseaCollectionName fields have been added to the GraphQL `Nft` type
- GraphQL resolvers handle remapping the Dead Ringers URL to a smaller resized version
- Fixed an issue in the Public API where collection names and descriptions couldn't be updated to an empty string

### DIFF
--- a/graphql/resolver/schema.resolvers.go
+++ b/graphql/resolver/schema.resolvers.go
@@ -70,6 +70,77 @@ func (r *membershipOwnerResolver) User(ctx context.Context, obj *model.Membershi
 	return resolveGalleryUserByUserID(ctx, obj.Dbid)
 }
 
+func (r *mutationResolver) AddUserAddress(ctx context.Context, address persist.Address, authMechanism model.AuthMechanism) (model.AddUserAddressPayloadOrError, error) {
+	api := publicapi.For(ctx)
+
+	authenticator, err := r.authMechanismToAuthenticator(authMechanism)
+	if err != nil {
+		return nil, err
+	}
+
+	err = api.User.AddUserAddress(ctx, address, authenticator)
+	if err != nil {
+		return nil, err
+	}
+
+	output := &model.AddUserAddressPayload{
+		Viewer: resolveViewer(ctx),
+	}
+
+	return output, nil
+}
+
+func (r *mutationResolver) RemoveUserAddresses(ctx context.Context, addresses []persist.Address) (model.RemoveUserAddressesPayloadOrError, error) {
+	api := publicapi.For(ctx)
+
+	err := api.User.RemoveUserAddresses(ctx, addresses)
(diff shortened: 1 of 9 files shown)


graphql/resolver/schema.resolvers.go
```diff
@@ -70,172 +70,192 @@ func (r *membershipOwnerResolver) User(ctx context.Context, obj *model.Membershi
 	return resolveGalleryUserByUserID(ctx, obj.Dbid)
 }
 
-func (r *mutationResolver) CreateCollection(ctx context.Context, input model.CreateCollectionInput) (model.CreateCollectionPayloadOrError, error) {
+func (r *mutationResolver) AddUserAddress(ctx context.Context, address persist.Address, authMechanism model.AuthMechanism) (model.AddUserAddressPayloadOrError, error) {
```
radazen, line 73: This file is a little tricky because it's half codegen and half manual code. gqlgen generates all of the methods and then we fill them in. Consequently, it seems like sometimes when we regenerate it to pull in new schema changes, the file gets shuffled around and it looks like there are many more changes than were actually made.

The majority of real changes in this file were just adding the `UpdateNftInfo` function.

service/persist/nft_event.go
```diff
@@ -10,7 +10,7 @@ import (
 type NftEventRecord struct {
 	ID           DBID            `json:"id"`
 	UserID       DBID            `json:"user_id"`
-	NftID        TokenID         `json:"nft_id"`
+	NftID        DBID            `json:"nft_id"`
```
radazen, line 13: I switched this over to `DBID` for the time being since we're still using v1 NFTs

publicapi/nft.go
```diff
@@ -89,3 +92,51 @@ func (api NftAPI) RefreshOpenSeaNfts(ctx context.Context, addresses string) erro
...
+	api.loaders.ClearAllCaches()
+
+	// Send event
+	nftData := persist.NftEvent{CollectorsNote: persist.NullString(collectorsNote)}
+	dispatchNftEvent(ctx, persist.NftCollectorsNoteAddedEvent, userID, nftID, nftData)
+
+	return nil
+}
+
+func dispatchNftEvent(ctx context.Context, eventCode persist.EventCode, userID persist.DBID, nftID persist.DBID, nftData persist.NftEvent) {
```
radazen, line 131: I just copied the way we dispatch collection events and did it for NFTs, too. I think it should work correctly, but we'll see what happens when we start using the feedlot!

Robinnnnn: nice!

Robinnnnn (approved):

PR #345, opened 2022-04-08
no previews, no membership

bennycio:
Changes:

- **Changed** membership endpoint to not include members that have no NFTs displayed. We determine this by looking at whether or not they have any preview NFTs.

### DIFF
--- a/service/membership/membership.go
+++ b/service/membership/membership.go
@@ -299,8 +299,13 @@ func processOwners(ctx context.Context, id persist.TokenID, metadata alchemyNFTM
 				logrus.Debug("Event is to real address")
 				// does to have the NFT?
 				membershipOwner := fillMembershipOwner(ctx, addr, id, ethClient, userRepository, galleryRepository)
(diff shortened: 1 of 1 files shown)


service/membership/membership.go
```diff
@@ -341,25 +346,22 @@ func processOwners(ctx context.Context, id persist.TokenID, metadata alchemyNFTM
 func fillMembershipOwner(ctx context.Context, pAddress persist.Address, id persist.TokenID, ethClient *ethclient.Client, userRepository persist.UserRepository, galleryRepository persist.GalleryRepository) persist.MembershipOwner {
 	membershipOwner := persist.MembershipOwner{Address: pAddress}
 
-	hasNFT, _ := eth.HasNFT(ctx, PremiumCards, id, pAddress, ethClient)
-	if hasNFT {
```
bennycio, line 345: Removed this because we can already guarantee the user has the premium card. Alchemy's response is a list of all addresses that own the NFT.

Robinnnnn (approved):

PR #346, opened 2022-04-09
Fallthrough on media type checking

Robinnnnn:
switch/case in golang [doesn't fall through](https://stackoverflow.com/a/40822747). this was causing `.glb` files to be returned as `UnknownMedia` types.

confirmed this fix works by testing on frontend


kaitoo1 (approved):

radazen:
Yikes, thanks for catching that!

PR #353, opened 2022-04-13
Fix a bug when parsing address lists for Opensea refresh

radazen:
Refreshing Opensea assets via list of addresses would accidentally add the entire comma-separated list of addresses as a single "address" in the resulting output slice.


kaitoo1 (approved):
nice catch!

Robinnnnn (approved):

PR #357, opened 2022-04-18
Debug Tools

radazen:
## What's new?
This PR adds two things:

1. A `DebugAuthenticator` type that can be used to bypass authentication and impersonate any user or address
2. One trillion safeguards to stop a `DebugAuthenticator` from ever being usable in production

### Why?
One really cool thing about using `Authenticators` for auth is that we have a very modular auth system. Anything that requires authentication accepts an `Authenticator` type, and we can use that to allow auth from all sorts of sources (nonce-based, OAuth, other integrations) anywhere we want to verify ownership of a user account or an address.

This is _also_ very handy for local debugging. A `DebugAuthenticator` can be used to impersonate any user or validate any address for the sake of testing/debugging from the perspective of the given user. This is great for investigating issues that are happening to specific users, or for seeing what would happen if a particular user disconnected a wallet, etc.

### Sounds cool! And also scary. How do we make sure this is only usable locally?
I've implemented a bunch of safeguards to make sure a `DebugAuthenticator` can only be used in a `local` environment.

1. Our GraphQL schema now supports a `restrictEnvironment` directive. Any attempt to use a field annotated with `restrictEnvironment` will throw an error if the current environment is not in the directive's `allowed` parameters. In the case of `DebugAuthenticator`, it looks like this:

```
# DebugAuth always succeeds and returns the supplied userId and addresses.
# Only available for local development.
input DebugAuth @restrictEnvironment(allowed: ["local"]){
    userId: DBID
    addresses: [Address!]!
}
```

2. Once the GraphQL resolvers receive a `DebugAuth` input, they'll only map it to a `DebugAuthenticator` object if `debugtools.Enabled` is true and we're in a `local` environment:

```
if debugtools.Enabled {
    if viper.GetString("ENV") == "local" && m.DebugAuth != nil {
        userID := persist.DBID("")
        if m.DebugAuth.UserID != nil {
            userID = *m.DebugAuth.UserID
        }
        return debugtools.NewDebugAuthenticator(userID, m.DebugAuth.Addresses), nil
    }
}
```

3. `debugtools.Enabled` is part of a new compilation-based `debugtools` package. Unlike the other safeguards, which are all runtime checks, `debugtools` uses a golang build tag to choose which file to compile: either `debugtools_enabled.go` or `debugtools_disabled.go`. Only a build compiled with the `-tags debug_tools`* flag will contain a working implementation of a `DebugAuthenticator`, meaning that even if our production environment were somehow returning `ENV="local"`, the version of `DebugAuthenticator` returned by `debugtools` would still throw an error.

\* - Also worth noting: the `debug_tools` build tag is something I made up, so it only exists in builds where we explicitly set it. 

4. After all that, the working version of a `DebugAuthenticator` will still check the environment before returning successfully:

```
if viper.GetString("ENV") != "local" {
    return nil, errors.New("DebugAuthenticator may only be used in a local environment")
}

authResult := auth.AuthResult{
    Addresses: d.Addresses,
    UserID:    d.UserID,
}

return &authResult, nil
```

### DIFF
--- a/debugtools/debugtools_common.go
+++ b/debugtools/debugtools_common.go
@@ -1,0 +1,26 @@
+// debugtools_common.go is always compiled and is not dependent on a build tag.
+// It contains shared code used by both debugtools_enabled.go and debugtools_disabled.go.
+
+package debugtools
+
+import (
+	"fmt"
+	"github.com/mikeydub/go-gallery/service/auth"
+	"github.com/mikeydub/go-gallery/service/persist"
+)
+
+type DebugAuthenticator struct {
+	UserID    persist.DBID
+	Addresses []persist.Address
+}
+
+func (d DebugAuthenticator) GetDescription() string {
+	return fmt.Sprintf("DebugAuthenticator(userId: %s, addresses: %v)", d.UserID, d.Addresses)
+}
+
+func NewDebugAuthenticator(userID persist.DBID, addresses []persist.Address) auth.Authenticator {
+	return DebugAuthenticator{
+		UserID:    userID,
+		Addresses: addresses,
+	}
+}
--- a/debugtools/debugtools_disabled.go
+++ b/debugtools/debugtools_disabled.go
@@ -1,0 +1,20 @@
+//go:build !debug_tools
+// +build !debug_tools
+
+// debugtools_disabled.go is compiled whenever the `debug_tools` build tag is not set.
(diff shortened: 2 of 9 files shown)


bennycio (approved):
Love it!

Robinnnnn (approved):

kaitoo1 (approved):

PR #359, opened 2022-04-23
Fixup - various feedbot bugs

jarrel-b:
**Changes**
* Fixes the issue where tasks are sent to the backend rather than the feedbot
* Fixes nil pointer error that's caused when there is no previous collection token added event
* Fixes errors in retrieving NFT events because of type mismatch

**Testing**
* Tested changes locally and also deployed backend to sandbox

### DIFF
--- a/feedbot/collection.go
+++ b/feedbot/collection.go
@@ -130,17 +130,19 @@ func handleCollectionTokensAdded(ctx context.Context, userRepo persist.UserRepos
 	}
 
 	newTokensAdded := false
-	for _, i := range event.Data.NFTs {
-		contains := false
-		for _, j := range eventBefore.Data.NFTs {
-			if i == j {
-				contains = true
+	if eventBefore != nil {
+		for _, i := range event.Data.NFTs {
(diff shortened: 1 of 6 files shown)


radazen (commented):
Looks awesome, Jarrel! One small request: since the "feedbot" and "tasks/feed-event" parameters are reused in several places, could we make a wrapper method like "createTaskForFeedbot" that calls createTaskForService with the correct parameters? Or is there a case where we'd expect the feedbot to potentially pass in different parameters?

jarrel-b:
> Looks awesome, Jarrel! One small request: since the "feedbot" and "tasks/feed-event" parameters are reused in several places, could we make a wrapper method like "createTaskForFeedbot" that calls createTaskForService with the correct parameters? Or is there a case where we'd expect the feedbot to potentially pass in different parameters?

Great idea! Added the wrapper since I don't really see the feedbot needing more routes.

radazen (approved):
Looks great! I can't wait to see these start popping up on Discord.

PR #360, opened 2022-04-26
Feedbot Fixups

jarrel-b:
**Changes**
* Fixes bug where a username changed is considered to a new user, now it checks if that event had occurred before.
* Increases delay to 5 minutes in prod (kept to 10 seconds on dev)


radazen (approved):
🔥

PR #361, opened 2022-04-26
OpenSea refresh mutation now returns a catch-all error on failure

radazen:
Since OpenSea refreshing can fail in myriad ways, we don't want/need to handle every case with an explicit GraphQL error type, but it _is_ nice to return a GraphQL error so the client can handle sync failures with the same fragment-based approach it uses for other errors.


Robinnnnn (approved):
gracias!

PR #365, opened 2022-05-03
GAL-129 previews

bennycio:
Changes:

- **Removed** `CommunityOwner` and `MembershipOwner` types to be `TokenHolder` so the membership page and community pages use the same types for users.
- **Added** functionality for grabbing and resolving 3 thumbnails for images in the collection that a given `TokenHolder` owns for a community


graphql/schema/schema.graphql
```diff
@@ -262,7 +262,7 @@ type Gallery implements Node {
     collections: [Collection] @goField(forceResolver: true)
 }
 
-type MembershipOwner {
+type TokenHolder {
```
kaitoo1, line 265: 🙌

kaitoo1:
thanks benny! excited to take this for a spin.
just fyi I added the linear ticket number to the PR title

kaitoo1:
also this'll be a breaking change because of the type change right? we can coordinate deploying this + front end when it's ready

graphql/resolver/schema.resolvers.go
```diff
@@ -413,14 +409,18 @@ func (r *queryResolver) CollectionNftByID(ctx context.Context, nftID persist.DBI
...
+func (r *queryResolver) CommunityByAddress(ctx context.Context, contractAddress persist.Address, forceRefresh *bool) (model.CommunityByAddressOrError, error) {
+	return resolveCommunityByContractAddressWithRefresh(ctx, contractAddress, *forceRefresh)
 }
 
 func (r *queryResolver) GeneralAllowlist(ctx context.Context) ([]persist.Address, error) {
 	return publicapi.For(ctx).Misc.GetGeneralAllowlist(ctx)
 }
 
+func (r *tokenHolderResolver) User(ctx context.Context, obj *model.TokenHolder) (*model.GalleryUser, error) {
+	panic(fmt.Errorf("not implemented"))
```
kaitoo1, line 421: are there plans to implement this? I think we need to be able to resolve the user to get the username to display on the page 👀

bennycio: yes, will implement this!

bennycio:
> also this'll be a breaking change for the membership directory and community pages because of the type change right? we can coordinate deploying this + front end when it's ready

Totally, will need to coordinate.

radazen (commented):
Looks awesome, Benny! Expertly GraphQL'd. A few questions/thoughts:

- When would we expect the frontend to use the `forceRefresh` parameter?

- This might be a pain, but instead of giving a `TokenHolder` an `address` field, can we change it to `addresses` and include all of that user's token-containing addresses? E.g. for someone who has Crypto Coven NFTs spread across several wallets, could we return a single `TokenHolder` object with multiple `addresses` listed, one for each wallet that has witches in it? I'm asking because this is actually something I wanted to implement in the future! When we launched communities, we were returning one `CommunityOwner` per NFT-owning address, so we would include the same user multiple times if they owned community NFTs in multiple wallets. I added a de-duplication step to fix that, but we only end up returning one owner object (and consequently, just one of their community addresses). I don't know if we'll have a need for all of their community addresses in the future, but it feels more correct than de-duplicating and returning just one of their community-NFT-containing addresses. But also, if it's a pain, I don't think it matters that much right now!

radazen:
> Looks awesome, Benny! Expertly GraphQL'd. A few questions/thoughts:
> 
> * When would we expect the frontend to use the `forceRefresh` parameter?
> * This might be a pain, but instead of giving a `TokenHolder` an `address` field, can we change it to `addresses` and include all of that user's token-containing addresses? E.g. for someone who has Crypto Coven NFTs spread across several wallets, could we return a single `TokenHolder` object with multiple `addresses` listed, one for each wallet that has witches in it? I'm asking because this is actually something I wanted to implement in the future! When we launched communities, we were returning one `CommunityOwner` per NFT-owning address, so we would include the same user multiple times if they owned community NFTs in multiple wallets. I added a de-duplication step to fix that, but we only end up returning one owner object (and consequently, just one of their community addresses). I don't know if we'll have a need for all of their community addresses in the future, but it feels more correct than de-duplicating and returning just one of their community-NFT-containing addresses. But also, if it's a pain, I don't think it matters that much right now!

Actually, it just occurred to me: rolling all addresses into a single user would help us get multiple preview images, right? In the current implementation, if I have 3 witches across 3 wallets, would we return 3 `TokenHolder`s with one preview URL each?

db/migrations/000007_create_owner_address_index_on_nfts.up.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX IF NOT EXISTS nfts_owner_address_idx ON nfts(owner_address);
```
radazen, line 1: Getting preview images for Crypto Coven on my local machine took about 15 seconds prior to this index, and about 1 second after.

radazen:
Additional changes:

- Updated `TokenHolder` to contain a list of addresses instead of a single address
  - Each distinct address on a `TokenHolder` owns the token and belongs to the `TokenHolder`'s user
  - This is useful for users who own multiple tokens spread across wallets (e.g. for community members, we have all of their qualifying addresses available to pull preview images from)

- Removed `Dbid` field from `TokenHolder`
  - Now that `TokenHolder` references a `GalleryUser`, we don't need a user ID in the response -- we can just get it from `TokenHolder.User.Dbid` if we need it

- Set an index on the `nfts.owner_address` field to speed up preview image queries

kaitoo1 (approved):
Great catch for the case where a user owns multiple tokens spread across wallets @radazen !
The changes worked smoothly for the front end.
Thanks for taking over and seeing this through to the finish line 🔥

kaitoo1:
Just a heads up that since this updates the schema, this PR will have to be deployed to prod at the same time as these front end changes: https://github.com/gallery-so/gallery/pull/746
otherwise the queries on the Community Page and Membership Directories will be broken.

the plan is to get the front end PR reviewed today so should be merged soon. 

Not as important to get them on dev at the same time, but when we deploy to prod we should make sure to coordinate backend+frontend.

radazen:
> Just a heads up that since this updates the schema, this PR will have to be deployed to prod at the same time as these front end changes: [gallery-so/gallery#746](https://github.com/gallery-so/gallery/pull/746) otherwise the queries on the Community Page and Membership Directories will be broken.
> 
> the plan is to get the front end PR reviewed today so should be merged soon.
> 
> Not as important to get them on dev at the same time, but when we deploy to prod we should make sure to coordinate backend+frontend.

Sounds good! I'm going to run the "add column index" migration on prod now since it can be done ahead of time. Will do a bit more testing of this PR today, but it should be ready to deploy when the frontend is!

PR #367, opened 2022-05-05
[GAL-144] Return opensea IDs with NFTs

Robinnnnn:
Allows us to dedupe by opensea ID on frontend


radazen (approved):
Looks perfectly reasonable to me! Way to GraphQL!

PR #368, opened 2022-05-05
[GAL-133] Tracing and Logging

radazen:
# What's new?

_**Note to reviewers: this PR touches a lot of files, but most of those files only have trivial changes from `logrus.X` to `logger.For(ctx).X`**_

## Logging

- Created `logger` package as context-aware wrapper for logrus

  - Additional logging fields can be added to context via `logger.NewContextWithFields` or helpers like `logger.NewContextWithSpan`
  
  - Use `logger.For(ctx)` instead of `logrus` to get a logger that will include all fields added to the context
  
  - Use `logger.For(nil)` when there's no context available
    - This was originally `logger.NoCtx()`, but `logger.For(nil)` keeps the syntax more uniform and hints to maintainers that they could just drop a context in the `nil` spot in the future
  
  - Example: add span info to a context with `logger.NewContextWithSpan`, and then all subsequent logging with that context will automatically include the span and trace info

- GraphQL requests and responses are logged
  - Requests were already being logged, but they weren't very helpful because they didn't include the `variables` array sent with the request
    - The `variables` array is now scrubbed and included in logging
   
  - Google Cloud Logging will truncate log entries at 256k, so large responses will still be missing data, but pagination should help this in the future
 
  - Requests and response logs include IDs to make them easier to identify:
    - `gqlRequestId` connects a request and response
    - `requestLocatorId` and `responseLocatorId` uniquely identify a single log entry; useful because Sentry can include the ID to make searching for a particular logged request/response easy

- logrus now defaults to JSON formatting in deployed environments, which makes structured logging fields easier to search

## Tracing

- Tracing is now enabled via Sentry! We're tracing:
  - GraphQL (requests, responses, resolvers)
  - Outgoing HTTP requests (e.g. fetching from OpenSea)
  - Database queries
  - All three of these are automatic and will apply to future GraphQL/HTTP/database work without additional steps

- By using Sentry to trace both frontend and backend services, we can see the entire lifecycle of a request

- Tracing also enhances our Sentry error reports (e.g. "the error occurred during this part of the trace")

## Sentry Refactoring

To support tracing, some Sentry refactoring was required

- `sentry` methods (like `sentry.StartSpan`) don't work with `sentrygin` by default, so I've wrapped `sentrygin` with some helper middleware

- Our `sentry` helper package has been renamed to `sentryutil` to avoid naming collisions

- GraphQL errors and remapped errors are now reported _as they occur_ instead of being queued for later reporting
  - Errors need to be reported _during_ a trace in order to be associated with it
  - Didn't realize we'd be able to do this easily via GraphQL middleware, but it turns out we can!

- `resolver/error.go` has been removed
  - No longer need most of this functionality since we aren't collecting GraphQL errors anymore
  - Any remaining functionality has been moved to `sentryutil` (e.g. `sentryutil.ReportRemappedError`)

- Moved JWT cookie scrubbing into its own method (tracing requires that we use it in an `EventProcessor` instead of a `BeforeSend` hook)

- Set up event fingerprinting to separate `errors.New()` errors based on their error strings. Otherwise, every error created with `errors.New()` gets lumped into a single Sentry issue

- We no longer report an error to Sentry when a user isn't logged in (previously reported as the `NoJWT` error, but this is an expected thing for a user who isn't logged in, and we've been getting tens of thousands of these error reports)


service/rpc/rpc.go
```diff
@@ -26,20 +28,19 @@ import (
...
 	Timeout: time.Second * 30,
-	Transport: &http.Transport{
+	Transport: sentryutil.NewTracingTransport(&http.Transport{
 		Dial: (&net.Dialer{
 			KeepAlive: keepAliveTimeout,
 		}).Dial,
 		MaxIdleConns:        100,
 		MaxIdleConnsPerHost: 100,
-	},
+	}, true),
```
radazen, line 43: `true` here means that we'll only trace the HTTP request if there's already a parent trace in progress. Tracing is only enabled on the `server` service right now, and I thought it'd be weird if other services in `go-gallery` started tracing and reporting all of their HTTP requests with no additional context. If/when we set up tracing in those services, the HTTP requests will be added to the trace automatically.

service/sentry/sentry.go
```diff
@@ -11,40 +15,104 @@ import (
...
+	if hub == nil {
+		logger.For(ctx).Warnln("could not report error to Sentry because hub is nil")
+		return
+	}
+
+	// Use a new scope so our error context and tag don't persist beyond this error
+	hub.WithScope(func(scope *sentry.Scope) {
+		if remappedErr != nil {
+			SetErrorContext(scope, true, fmt.Sprintf("%T", remappedErr))
+			scope.SetTag("remappedError", "true")
```
radazen, line 45: Added a tag here so we can exclude remapped errors from Sentry dashboards/searches if we want to

service/sentry/sentry.go
```diff
@@ -54,20 +122,68 @@ func NewSentryHubContext(ctx context.Context, hub *sentry.Hub) context.Context {
...
 		return hub
 	}
 
-	hub, ok := ctx.Value(SentryHubContextKey).(*sentry.Hub)
-	if !ok {
-		return nil
+	return nil
+}
+
+type tracingTransport struct {
```
radazen, line 148: I'm not entirely convinced that this HTTP tracer belongs here and not in its own `tracing` package, but I put here for now since we use Sentry for tracing.

_encrypted_deploy
```diff
@@ -1,7 +1,7 @@
...
... JtBia2olc0o3gZO K9HAr1g05DAamx2 RzKoS2iLiiq5w7K Hr6JCQru2G7ToKC ay2CBsFjTUzF4QB 7jphyTy1j3okqbS xBDdoIx0jZGaH9C DWORLmUR4PGbgRD DeqVIr7AHGwk2os Cx21Ldbwu9QHMBY t45p8993b0ZFCp7 CJd4YZk7hKh9RQz whWJ9pcRS6wCtsF LWUW2L30aderwMv GUtQPonyjEBJ68T SqmWjTeRUqrMvDy SAcNbJMz9tLybJz 5odUXxivdxL5Hq5 DSTrTNt7EI6g250 HVIWcPU5pcaIuPZ nVxFzkkzDNBWJ59 jsDT3MwXnvIeNx4 dAJpOiqFBH52xYq pxfRQ9TSg2Z58EN G5qQYxWwFmpPOJ3 NbkwXKojLs9MIBo DcHde5o3FtCNRRN Z2tHuLDeZpTyun1 bfomyWTMwYaGESb cJeC2s0dBNhguoz Djh24DIOaPj4uO3 uv0YoycnTbLeMJt Z1wr54bUIJCTqta IRdDyyW3ubndBgA iIhp82z5AeS3zA1 lASLSEFoTsoqJ6c TE8syUxU3qcS2xp vfIRZOPHqh0jo9Z 8jcWDitw8QHXtbP fo5hveCi5aXlixn IeZrFomviyjQaHP BWLJMN7vHtwtq9s 0kSPN1QM0ZovO0d i7pia7uQzPIpjzG tNH20nCJVsoyodP VTYQ0LDddOMro5m zzJQpxm3GBbf8lJ YXORr4wwCl77UsY Mh7R4B6SVtlx9hQ CTGUOgypAihKj6g 5OClWWbgYbFKbVe AfTdH9vrReokHPE tLAl4ZeMUk3f127 6lJiOBAQqfnEkNz TpAgJbXjpOB6UaT hlDaftPmEJxH7Y0 9VXKKgABP3ZzPd2 TbjHHk3MBTJmG8s S7ITYvFBUKTbUbT KdFsRZjbZ8nn0Bb 2oaK2UcjDenOkVi l5LWLoTYXi6q0mX oxaAdNpADRSoFln 8GQdqcEjsPLq5FN HuLnj8UNbKOl6Ao U2BWvxqKliMxBFV FuZ6N6pBY0zAkHS buP6GKKL7KdH6fo qnA8SpQHFCimREB L5PVg3vv8mcnunb x9N7EijmQMCkExm Bgj16vmEZsvW7qb HcsmAl3QSR2cVFX nPNUesG4Sa7TPYO bqtsZ776Cf0iXpT
```
radazen, line 1: Changes in here:
- dev environments set `SENTRY_SAMPLE_RATE` to 1.0 so they sample everything
- prod sets `SENTRY_SAMPLE_RATE` to 0.1 so it samples 1/10th of requests

These are misleading, though, since sample rates only apply if the request we get from the frontend _doesn't_ contain a `sentry-trace` header. If it does (and it pretty much always should!), Sentry will use the sampling decision embedded that header.

graphql/resolver/handlers.go
```diff
@@ -141,46 +177,285 @@ func RestrictEnvironmentDirectiveHandler() func(ctx context.Context, obj interfa
...
+
+			// Unique ID to make finding this particular log entry easy
+			responseLocatorId = ksuid.New().String()
+
+			gc := util.GinContextFromContext(ctx)
+			userId := auth.GetUserIDFromCtx(gc)
+
+			// Fields are logged in alphabetical order, so scrubbedQuery is prefixed with a zzz_ to make sure
+			// it's last. In cases where a log entry is too large and gets truncated (e.g. Google Cloud Logging
+			// limit is 256kb per entry), we want to make sure all of our fields are visible.
```
Robinnnnn, line 247: amazing

middleware/middleware.go
```diff
@@ -215,6 +219,55 @@ func GinContextToContext() gin.HandlerFunc {
...
+		if reportGinErrors {
+			for _, err := range c.Errors {
+				sentryutil.ReportError(c, err)
+			}
+		}
+	}
+}
+
+func Tracing() gin.HandlerFunc {
+	// Trace outgoing HTTP requests
```
Robinnnnn, line 250: 😍

service/sentry/sentry.go
```diff
@@ -54,20 +122,68 @@ func NewSentryHubContext(ctx context.Context, hub *sentry.Hub) context.Context {
...
+	http.RoundTripper
+
+	continueOnly bool
+}
+
+// NewTracingTransport creates an http transport that will trace requests via Sentry. If continueOnly is true,
+// traces will only be generated if they'd contribute to an existing parent trace (e.g. if a trace is not in progress,
+// no new trace would be started).
```
Robinnnnn, line 156: 🔥

Robinnnnn (approved):
holy shit ezra, this is magnificent 👏 👏 👏 

this will go an enormous way towards tracking down issues. thank you so much for your dedication to this project!

graphql/resolver/handlers.go
```diff
@@ -141,46 +177,285 @@ func RestrictEnvironmentDirectiveHandler() func(ctx context.Context, obj interfa
...
+
+	placeholder := fmt.Sprintf("%s omitted because it is too large (%d bytes)", name, length)
+	if locatorId != "" {
+		placeholder += fmt.Sprintf(", but it should be accessible by searching logs for this unique ID: %s", locatorId)
+	}
+
+	return placeholder
+}
+
+func scrubVariable(variableDefinition *ast.VariableDefinition, schema *ast.Schema, allQueryVariables map[string]interface{}, scrubbedOutput map[string]interface{}) {
```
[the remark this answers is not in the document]
radazen: Haha, funny timing. Just pushed a fix for a potential nil pointer issue here.

Testing this is probably a _little_ tricky because it depends on schema directives, though I suppose we could supply a schema at runtime with various things flagged as being scrubbable. Otherwise, if we're using our real schema, the only thing set to be scrubbed right now is an auth nonce signature (which is mostly just a proof of concept for scrubbing since signatures are actually not very sensitive).

Definitely a good call. This will be a good one to think about a bit when setting up tests for all of these GraphQL queries/mutations.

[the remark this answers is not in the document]
Robinnnnn: nice!

service/persist/postgres/postgres.go
```diff
@@ -60,6 +73,71 @@ func NewPgxClient() *pgxpool.Pool {
 	return db
 }
 
+type pgxTracer struct{}
+
+func (l *pgxTracer) Log(ctx context.Context, level pgx.LogLevel, msg string, data map[string]interface{}) {
```
jarrel-b, line 78: 🔥

graphql/resolver/handlers.go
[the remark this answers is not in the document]
kaitoo1: great point @Robinnnnn  and thanks for the explanation @radazen

server/handler.go
```diff
@@ -54,49 +53,38 @@ func graphqlHandler(repos *persist.Repositories, queries *sqlc.Queries, ethClien
 
 	schema := generated.NewExecutableSchema(config)
 	h := handler.NewDefaultServer(schema)
-	h.AroundOperations(graphql.ScrubbedRequestLogger(schema.Schema()))
-	h.AroundFields(graphql.RemapErrors)
-	h.AroundResponses(graphql.AddErrorsToGin)
+
+	// Request/response logging is spammy in a local environment and can typically be better handled via browser debug tools.
```
kaitoo1, line 57: love all the detailed comments 👏

service/persist/postgres/postgres.go
```diff
@@ -60,6 +73,71 @@ func NewPgxClient() *pgxpool.Pool {
...
+	} else if strings.EqualFold(msg, "exec") {
+		operation = "exec"
+	}
+
+	description := msg
+
+	sqlStr, ok := data["sql"].(string)
+	if ok {
+		// If a SQL statement was supplied, use that as the default description
+		description = sqlStr
```
jarrel-b, line 101: Is `sqlStr` the result after args are bound to the template query? Wondering if this matters for nonces?

service/sentry/sentry.go
```diff
@@ -11,40 +15,104 @@ import (
...
+}
+
+func ScrubEventCookies(event *sentry.Event, hint *sentry.EventHint) *sentry.Event {
+	if event == nil || event.Request == nil {
+		return event
+	}
+
+	var scrubbed []string
+	for _, c := range strings.Split(event.Request.Cookies, "; ") {
+		if !strings.HasPrefix(c, auth.JWTCookieKey) {
```
jarrel-b, line 65: realized that maybe instead of removing it should be replaced with  `[filtered]` so we know the request contained it?

kaitoo1 (approved):
Awesome stuff @radazen ! +1 to what robin said and thanks for taking charge to make these improvements 🔥 can't wait to see this in action on prod

middleware/middleware.go
```diff
@@ -215,6 +219,55 @@ func GinContextToContext() gin.HandlerFunc {
 	}
 }
 
+func Sentry(reportGinErrors bool) gin.HandlerFunc {
+	handler := sentrygin.New(sentrygin.Options{Repanic: true})
+
+	return func(c *gin.Context) {
```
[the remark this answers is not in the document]
radazen: Yup! Since we're just wrapping the `sentrygin` handler, it'll call `c.Next()`.

Thanks for pointing this out! I'll leave a comment about it, because you're right -- the absence of `c.Next()` is odd.

service/persist/postgres/postgres.go
radazen: `sqlStr` is just the query without args bound, but I _am_ including args in the "sql args" section below.

I originally left args out because I like to err on the side of caution when it comes to security...but we don't actually work with any sensitive data right now -- even our nonces aren't actually sensitive and it would be okay if a malicious actor had access to them.

While I'd _hope_ we don't end up in a position where we'd have sensitive info in a SQL query, it's certainly possible, and something we should keep in mind if/when the situation arises. It's the kind of thing that could absolutely become an attack vector: person B added query X that they assumed was safe because no one sees our SQL queries; person C got hacked and an attacker gained access to Sentry; attacker now has access to a bunch of our database args. We can, of course, scrub those args...we just have to remember to do it.

That being said, it just seems _so incredibly useful_ to see args for slow queries and errors, so I'm hoping the benefits outweigh the risks on this one.

service/sentry/sentry.go
radazen: Good idea! I'll add that.

service/persist/postgres/postgres.go
jarrel-b: Makes sense to me, and agreed!

Robinnnnn (approved):

PR #369, opened 2022-05-09
[GAL-149] Add following

jarrel-b:
This PR adds `followers` and `following` fields to the `User` model and mutations to follow and unfollow a user. The server will also log when a user has been followed. I'll follow up to add it to the feedbot, thinking I'd like to first refactor the bot a bit so that it uses the publicapi rather than using the db directly so that its easier to write messages with better context.

Some test queries are below. To test following a user, I first inserted inserting a dummy user into the `users` table.
```
query UserByUsername {
  userByUsername(username: "jarrel") {
    ...on GalleryUser {
      id
      dbid
      followers {
        id
        dbid
      }
      following {
        id
        dbid
        following {
          id
        }
        followers {
          id
        }
      }
    }
  }
}

mutation FollowUser {
  followUser(followee: "28uRB7u8GjWH52xkhgLem732812") {
    ...on FollowUserPayload {
      userId
    }
    ...on ErrInvalidInput {
      __typename
      message
    }
  }
}

mutation UnfollowUser {
  unfollowUser(followee: "28uRB7u8GjWH52xkhgLem732812") {
    ...on UnfollowUserPayload {
      userId
    }
    ...on Error {
      __typename
      message
    }
  }
}
```


kaitoo1:
👀👀👀

Robinnnnn (approved):
wooooo!! thanks for knocking this out! 🔥🔥🔥

graphql/dataloader/dataloaders.go
```diff
@@ -120,6 +125,18 @@ func NewLoaders(ctx context.Context, q *sqlc.Queries) *Loaders {
 		fetch:    loadMembershipByMembershipId(ctx, loaders, q),
 	}
 
+	loaders.FollowersByUserId = FollowersLoaderById{
+		maxBatch: defaultMaxBatchOne,
```
radazen, line 129: I 100% did not document this at all, but could you make this `defaultMaxBatchMany`? Still experimenting with batch sizes, but in an attempt to keep them from getting too big, `defaultMaxBatchOne` is a higher limit for queries that can only return one row and `defaultMaxBatchMany` is a lower limit for queries that return many rows (like this one).

graphql/dataloader/dataloaders.go
```diff
@@ -191,6 +209,16 @@ func (l *Loaders) ClearMembershipCaches() {
 	l.MembershipByMembershipId.mu.Unlock()
 }
 
+func (l *Loaders) ClearFollowCaches() {
```
radazen, line 212: Thank you for following the pattern here!

graphql/dataloader/dataloaders.go
```diff
@@ -120,6 +125,18 @@ func NewLoaders(ctx context.Context, q *sqlc.Queries) *Loaders {
...
 	}
 
+	loaders.FollowersByUserId = FollowersLoaderById{
+		maxBatch: defaultMaxBatchOne,
+		wait:     defaultWaitTime,
+		fetch:    loadFollowersByUserId(ctx, loaders, q),
+	}
+
+	loaders.FollowingByUserId = FollowingLoaderById{
+		maxBatch: defaultMaxBatchOne,
```
radazen, line 135: Same as above, re: batch size

graphql/dataloader/dataloaders.go
```diff
@@ -493,3 +521,51 @@ func loadMembershipByMembershipId(ctx context.Context, loaders *Loaders, q *sqlc
...
+		defer b.Close()
+
+		b.Query(func(i int, u []sqlc.User, err error) {
+			following[i] = u
+			errors[i] = err
+
+			// Add results to other loaders' caches
+			if err == nil {
+				for _, user := range following[i] {
+					loaders.UserByUsername.Prime(user.Username.String, user)
```
radazen, line 564: Same as above re: priming UserByUserId

graphql/schema/schema.graphql
```diff
@@ -56,6 +56,8 @@ type GalleryUser implements Node {
     wallets: [Wallet]
     galleries: [Gallery] @goField(forceResolver: true)
     isAuthenticatedUser: Boolean
+    followers: [GalleryUser] @goField(forceResolver: true)
+    following: [GalleryUser] @goField(forceResolver: true)
```
radazen, line 60: Perfect use case for generating new resolvers!

graphql/schema/schema.graphql
```diff
@@ -628,4 +650,7 @@ type Mutation {
     createUser(authMechanism: AuthMechanism!): CreateUserPayloadOrError
     login(authMechanism: AuthMechanism!): LoginPayloadOrError
     logout: LogoutPayload
+
+    followUser(followee: DBID!): FollowUserPayloadOrError @authRequired
```
radazen, line 654: Nitpick, but maybe `userId` instead of `followee` here? Seeing `followee` in this context makes me stop and say "Wait, is it my ID or their ID? I'm the follower, they're the followee...ok, it's their ID,"  but only because follower/followee feels like a context clue that I should think about who's who. That might just be a me thing, though.

graphql/schema/schema.graphql
```diff
@@ -601,6 +603,26 @@ type CreateUserPayload {
...
+    | ErrInvalidInput
+
+union UnfollowUserPayloadOrError =
+    UnfollowUserPayload
+    | ErrAuthenticationFailed
+    | ErrUserNotFound
+    | ErrInvalidInput
+
+type FollowUserPayload {
+    userId: DBID
```
radazen, line 619: Could we return `viewer: Viewer` here instead? Same for `UnfollowUserPayload` below. That'd let callers combine the "add this person and fetch the updated list" flow into a single query, e.g.

```
followUser(someId) {
    ...on FollowUserPayload {
        viewer {
            user {
                // "following" now contains the user we just added
            }
        }
    }
}
```

publicapi/user.go
```diff
@@ -206,6 +207,88 @@ func (api UserAPI) GetCommunityByContractAddress(ctx context.Context, contractAd
 	return &community, nil
 }
 
+func (api UserAPI) GetFollowersById(ctx context.Context, userID persist.DBID) ([]sqlc.User, error) {
```
radazen, line 210: This is totally also a nitpick, but could we call it `GetFollowersByUserId`? My exists-only-in-my-head, absolutely-not-documented, really-need-to-write-this-down mental model of `ById` functions in the public API is that they represent looking up an entity by its DBID, e.g. `GetGalleryById` -> Gallery by its DBID, which leaves room for any other lookup method to specify a different kind of ID (`ByUserId`, `ByAddress`, etc)

graphql/resolver/schema.resolvers.helpers.go
```diff
@@ -315,11 +345,16 @@ func userToModel(ctx context.Context, user sqlc.User) *model.GalleryUser {
...
-		Username:            &user.Username.String,
-		Bio:                 &user.Bio.String,
-		Wallets:             wallets,
-		Galleries:           nil, // handled by dedicated resolver
+		Dbid:     user.ID,
+		Username: &user.Username.String,
+		Bio:      &user.Bio.String,
+		Wallets:  wallets,
+
+		// each handeled by dedicated resolver
```
radazen, line 353: 🙌

radazen (commented):
This is _so good_. I left a few comments about little things that are (a) nitpicky and/or (b) entirely my fault because I didn't document them well. But I _really_ appreciate just how well you stuck to the patterns set in the GraphQL/sqlc/public API/dataloader packages, especially because they are undocumented patterns that exist largely in code.

graphql/dataloader/dataloaders.go
```diff
@@ -493,3 +521,51 @@ func loadMembershipByMembershipId(ctx context.Context, loaders *Loaders, q *sqlc
...
+		defer b.Close()
+
+		b.Query(func(i int, u []sqlc.User, err error) {
+			followers[i] = u
+			errors[i] = err
+
+			// Add results to other loaders' caches
+			if err == nil {
+				for _, user := range followers[i] {
+					loaders.UserByUsername.Prime(user.Username.String, user)
```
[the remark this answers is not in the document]
jarrel-b: Good call! missed that one

graphql/schema/schema.graphql
jarrel-b: yup was actually considering renaming it as well can see how that's a little confusing for the frontend

graphql/schema/schema.graphql
jarrel-b: very neat, will change 👍

Robinnnnn (approved):
so spicy. amazing reviews @radazen 🌶🌶🌶

graphql/schema/schema.graphql
kaitoo1: ooh love this will make life easier on the front end thanks guys!

graphql/schema/schema.graphql
kaitoo1: much appreciated

kaitoo1:
awesome work @jarrel-b that was so fast 😮 
can't wait to try it out

kaitoo1 (approved):

radazen (approved):
🙌🙌🙌
🙌🙌🙌
🙌🙌🙌

radazen:
> 🙌🙌🙌
> 🙌🙌🙌
> 🙌🙌🙌

(and also, thanks for handling all of my questions/comments so quickly!)

PR #370, opened 2022-05-09
[GAL-131] REQUIRE_NFTS speedup / Redis tracing / Tracing tweaks

radazen:
# What's new?

## REQUIRE_NFTS
- We require a Gallery membership card to sign up or log in, but we _also_ require a membership card for any action that requires membership. In the GraphQL era, we're checking a user's auth status as part of any logged-in request, which means we run the REQUIRE_NFTS check every time a logged-in user does anything on our site.

- REQUIRE_NFTS fires off a call to Alchemy to see whether a user has one of our membership NFTs. This means that every time a logged-in user does anything, they have to wait for a round trip HTTP request from our servers to Alchemy. This often takes some ~400-500ms to complete.

- This PR addresses the issue by caching the result of that Alchemy call in Redis with a 1-hour expiration time. As long as a user has one of our membership cards when we check with Alchemy, they can keep using the site for the next hour without the need for another Alchemy round-trip. The 1-hour window was my guess at a good starting value; we can make this shorter or longer, but one round trip per hour seems okay to me.

## Redis Tracing
- Given the tracing PR I just finished, it should come as no surprise that I implemented Redis tracing while working on this REQUIRE_NFTS improvement. Nothing too crazy here: I took the OpenTelemetry tracing hook from the `go-redis` repository and customized it for our purposes.

## Tracing Tweaks
- I didn't want to make a `tracing` package in my original tracing PR because it didn't seem necessary, but with the addition of the Redis tracing hook, it made sense to create a new package. HTTP and Redis tracing implementations both live in the `tracing` package now, and I added a few utility methods (like `StartSpan` and `FinishSpan`) to make tracing easier.

## Auth Tweaks
- Also worth noting that I did some very minor refactoring in the `auth` package to make the "does user have required NFT" checking more reusable.

### DIFF
--- a/graphql/resolver/handlers.go
+++ b/graphql/resolver/handlers.go
@@ -8,11 +8,14 @@ import (
 	gqlgen "github.com/99designs/gqlgen/graphql"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/getsentry/sentry-go"
+	goredis "github.com/go-redis/redis/v8"
 	"github.com/mikeydub/go-gallery/graphql/model"
 	"github.com/mikeydub/go-gallery/publicapi"
 	"github.com/mikeydub/go-gallery/service/auth"
-	"github.com/mikeydub/go-gallery/service/eth"
 	"github.com/mikeydub/go-gallery/service/logger"
+	"github.com/mikeydub/go-gallery/service/memstore/redis"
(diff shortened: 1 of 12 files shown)


Robinnnnn (approved):
holy. shit.

you ser are a gentleman and a scholar!!! this is gonna lowkey speed up the app a ton 🔥🔥

PR #371, opened 2022-05-11
Add missing SQL schema items to docker init file

radazen:
Small PR: these items are present in our migrations, but not our local docker image bootstrap init file.

Longer term thought: might be worthwhile to get rid of the bootstrap init file entirely, and just run the migrations when spinning up a new local database image. That'd keep local databases more in sync with dev/prod ones!

### DIFF
--- a/docker/postgres/03_init.sql
+++ b/docker/postgres/03_init.sql
@@ -247,4 +247,24 @@ CREATE TABLE IF NOT EXISTS collection_events (
     DATA JSONB
 );
 
+CREATE TABLE IF NOT EXISTS follows (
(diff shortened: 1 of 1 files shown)


Robinnnnn (approved):
1. thank you for doing this!! 🔥
2. i do like your longer-term thought. what's the path to getting there?
3. darn unit tests failing, my bad, i'll dig into those today

radazen:
> * i do like your longer-term thought. what's the path to getting there?

Probably a one-line change to the makefile. The only downside is that we'd require devs to install the `go-migrate` tool as a prerequisite, but that shouldn't be a big deal.

PR #372, opened 2022-05-13
Initial support for Imgix resizing

radazen:
# What's new?

## Imgix
- GraphQL resolvers can now use the `imgix` SDK to generate URLs for resized versions of image and preview image URLs
- The behavior is currently **disabled** by default (with an `imgixEnabled` const in `schema.resolvers.helpers.go`), so if you'd like to test this PR locally, set `imgixEnabled` to true.

## _internal
- _internal now contains an `app-local-backend.yaml` config file for secrets required when running locally
- The only secret in `app-local-backend.yaml` right now is `IMGIX_SECRET`, which will be required in all environments -- including local -- once we switch over to imgix. Without `IMGIX_SECRET` set, the backend won't be able to return resized image URLs at all.
- Robin and I discussed the options here (if `IMGIX_SECRET` is missing, do we just return full-res images everywhere?) but decided that having the backend secretly broken and returning oversized images is not ideal. If something's going to fail, let's have it fail loudly.
- _**NOTE: THIS MEANS YOU WILL NEED TO DECRYPT THE _internal DIRECTORY AFTER THIS PR IS MERGED INTO MAIN, OR ELSE THE BACKEND WILL PANIC WHEN YOU TRY TO RUN IT.**_

## Considerations
- There are still details to be ironed out here. I picked some parameters for the resized images as a starting point, but we'll want to experiment and see what works best for us. Imgix has [an absurd number of parameters we can set](https://docs.imgix.com/apis/rendering#all-parameters), though I suspect we'll only want to use a few of them. My initial settings look like this:

```
small:  w=100,  auto=format,compress
medium: w=500,  auto=format,compress
large:  w=1000, auto=format,compress
srcSet: <not setting any parameters here, letting Imgix return its default>
```


jarrel-b (approved):
nice!

Robinnnnn (approved):
🙌🙌🙌

PR #373, opened 2022-05-13
[GAL-151] Fix flaky backups test

radazen:
Some of the test scenario times were too exact and didn't leave wiggle room (e.g. testing if two events are exactly 24 hours apart returns true if they're 24h apart but false if they're 23h 59m 59.9999s apart). Fixed by changing some of the test times so there wouldn't be any ambiguity when comparing them, and by storing a single `time.Now()` call so test scenario times are more deterministic.


Robinnnnn (approved):
🙏🙏🙏

PR #374, opened 2022-05-16
Use username for follow input

jarrel-b:
**Changes**
This PR updates the follow mutations to use `username` instead of the `userID` as the input type. I noticed that the `userByUsername` query uses `username` and thought it'd be good to keep the API consistent. 

This is also helpful for the feedbot, since it can use the `username` added to the event context to make the followee URL without needing to fetch more data.

### DIFF
--- a/publicapi/user.go
+++ b/publicapi/user.go
@@ -247,36 +247,42 @@ func (api UserAPI) GetFollowingByUserId(ctx context.Context, userID persist.DBID
 	return following, nil
 }
 
-func (api UserAPI) FollowUser(ctx context.Context, userID persist.DBID) error {
-	// Validate
+func (api UserAPI) FollowUser(ctx context.Context, username string) error {
 	curUserID, err := getAuthenticatedUser(ctx)
 	if err != nil {
 		return err
 	}
 
+	curUser, err := api.GetUserById(ctx, curUserID)
+	if err != nil {
+		return err
+	}
+
+	// Validate
 	if err := validateFields(api.validator, validationMap{
-		"userID": {userID, fmt.Sprintf("required,ne=%s", curUserID)},
+		"username": {username, fmt.Sprintf("required,ne=%s", curUser.UsernameIdempotent.String)},
 	}); err != nil {
 		return err
 	}
 
-	if _, err := api.GetUserById(ctx, userID); err != nil {
+	followee, err := api.GetUserByUsername(ctx, username)
+	if err != nil {
 		return err
 	}
 
-	err = api.repos.UserRepository.AddFollower(ctx, curUserID, userID)
+	err = api.repos.UserRepository.AddFollower(ctx, curUserID, followee.ID)
 
 	// Send event
-	userData := persist.UserEvent{FolloweeID: userID}
-	dispatchUserEvent(ctx, persist.UserFollowedEvent, userID, userData)
+	userData := persist.UserEvent{Username: curUser.UsernameIdempotent.String, FolloweeName: username}
+	dispatchUserEvent(ctx, persist.UserFollowedEvent, curUserID, userData)
 
 	return err
 }
 
-func (api UserAPI) UnfollowUser(ctx context.Context, userID persist.DBID) error {
+func (api UserAPI) UnfollowUser(ctx context.Context, username string) error {
 	// Validate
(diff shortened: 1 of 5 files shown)


publicapi/user.go
```diff
@@ -286,7 +292,12 @@ func (api UserAPI) UnfollowUser(ctx context.Context, userID persist.DBID) error
 		return err
 	}
 
-	return api.repos.UserRepository.RemoveFollower(ctx, curUserID, userID)
+	followee, err := api.GetUserByUsername(ctx, username)
+	if err != nil {
+		return err
```
jarrel-b, line 297: Previously the unfollowing operation was idempotent, we could also return `nil`, but I think this could be useful info to the caller?

radazen (commented):
This is an interesting one! I honestly lean toward using `userID` instead of `username` where possible, since `userID` will always uniquely identify the user in question, but `username` is fluid and could change.

Contrived but possible example:
- someone looks at your gallery and walks away from their computer
- you change your username
- that person comes back to their computer later and decides to follow you before navigating away from your gallery
- following fails since your username changed

I do agree that it'd make the API feel friendlier for sure, but the trade-off (IMO) is that our locally cached data also becomes a bit less reliable.

jarrel-b:
Sounds good! I think I totally misunderstood the meaning of `username_idempotent`. Was thinking that although a username could change, `username_idempotent` (thinking this is the arg that mutations get called with) is static and is another way to uniquely look up a user.

I like the idea of adding the username to the request, but to @radazen's point, if a username can change in between following and sending the event, it's probably best to re-fetch the username anyway. 

I have this somewhat arbitrary line drawn that the feedbot doesn't use the non-event tables directly, and asks the server for info instead. Was thinking if it makes sense to add a `userByUserId` query, but now that I think of it more it's probably fine to query the non-event tables since it's only for reads.

radazen:
> Sounds good! I think I totally misunderstood the meaning of username_idempotent. Was thinking that although a username could change, username_idempotent (thinking this is the arg that mutations get called with) is static and is another way to uniquely look up a user.

Ah! That makes sense. I was also confused by the `username_idempotent` column when I first encountered it. We might honestly be better off renaming it to `username_lowercase`, or even better: now that we're using Postgres instead of Mongo, get rid of it and just add a `lower` index to the username column: https://stackoverflow.com/questions/3980050/how-do-i-create-an-index-in-postgresql-based-on-lowercase-only

> I have this somewhat arbitrary line drawn that the feedbot doesn't use the non-event tables directly, and asks the server for info instead. Was thinking if it makes sense to add a userByUserId query, but now that I think of it more it's probably fine to query the non-event tables since it's only for reads.

I'm with you on this one. On one hand, it feels cleaner for the feedbot to stick to event tables only, but on the other hand, putting a username into the event data leaves the possibility that our event link is broken by the time the feedbot posts it. I actually really like the idea of a `userByUserId` query -- I think it'd be useful for stuff in the future, too. The functionality is already there, we'd just need to add it to the schema.

jarrel-b:
> > Sounds good! I think I totally misunderstood the meaning of username_idempotent. Was thinking that although a username could change, username_idempotent (thinking this is the arg that mutations get called with) is static and is another way to uniquely look up a user.
> 
> Ah! That makes sense. I was also confused by the `username_idempotent` column when I first encountered it. We might honestly be better off renaming it to `username_lowercase`, or even better: now that we're using Postgres instead of Mongo, get rid of it and just add a `lower` index to the username column: https://stackoverflow.com/questions/3980050/how-do-i-create-an-index-in-postgresql-based-on-lowercase-only
> 
> > I have this somewhat arbitrary line drawn that the feedbot doesn't use the non-event tables directly, and asks the server for info instead. Was thinking if it makes sense to add a userByUserId query, but now that I think of it more it's probably fine to query the non-event tables since it's only for reads.
> 
> I'm with you on this one. On one hand, it feels cleaner for the feedbot to stick to event tables only, but on the other hand, putting a username into the event data leaves the possibility that our event link is broken by the time the feedbot posts it. I actually really like the idea of a `userByUserId` query -- I think it'd be useful for stuff in the future, too. The functionality is already there, we'd just need to add it to the schema.

Sounds good! Will close this PR and follow up with that addition

PR #376, opened 2022-05-18
Add user field to follower mutations

jarrel-b:
**Changes**
* Adds `user` field to the following mutations for the user that was unfollowed/followed


kaitoo1 (approved):
thank you ser!!

PR #379, opened 2022-05-25
prehandle refreshes for specific tokens

bennycio:
Changes:

- **Changed** update media to update all tokens with token identifiers if specified in the call rather than manually updating each token. Saves a lot of time for ERC-1155 tokens.
- **Changed** some components that commonly cancelled due to context cancellations

### DIFF
--- a/indexer/nfts.go
+++ b/indexer/nfts.go
@@ -405,12 +405,27 @@ func updateMedia(tokenRepository persist.TokenRepository, ethClient *ethclient.C
 
 // UpdateMedia will find all of the media content for an addresses NFTs and possibly cache it in a storage bucket
 func UpdateMedia(c context.Context, input UpdateMediaInput, tokenRepository persist.TokenRepository, ethClient *ethclient.Client, ipfsClient *shell.Shell, arweaveClient *goar.Client, storageClient *storage.Client) error {
+	if input.TokenID != "" && input.ContractAddress != "" {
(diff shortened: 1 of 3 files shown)


indexer/nfts.go
```diff
@@ -405,12 +405,27 @@ func updateMedia(tokenRepository persist.TokenRepository, ethClient *ethclient.C
 
 // UpdateMedia will find all of the media content for an addresses NFTs and possibly cache it in a storage bucket
 func UpdateMedia(c context.Context, input UpdateMediaInput, tokenRepository persist.TokenRepository, ethClient *ethclient.Client, ipfsClient *shell.Shell, arweaveClient *goar.Client, storageClient *storage.Client) error {
+	if input.TokenID != "" && input.ContractAddress != "" {
+		logrus.Infof("updating media for token %s-%s", input.TokenID, input.ContractAddress)
+		tokens, err := tokenRepository.GetByTokenIdentifiers(c, input.TokenID, input.ContractAddress, 1, 0)
+		if err != nil {
+			return err
+		}
+		token := tokens[0]
```
radazen, line 414: Is this guaranteed to have a length >= 1 if there's no error returned from the previous call?

radazen (approved):
Left one comment re: slice length, but looks good to me!

indexer/nfts.go
bennycio: Yep!

PR #383, opened 2022-05-31
Feedbot Fixups

jarrel-b:
**Changes**
* Fixes a small bug where nft events were logged to stderr
* Sets up the feedbot with the newly added sentry and tracing middleware
    * Updated the tracing middleware to also set `http.DefaultClient` so that calls made using the default client gets tracing for free
    * Small bugfix where gin errors weren't being reported to Sentry
* Some small re-organization so that things are easier to find


Robinnnnn (approved):
🔥🔥🔥🔥

PR #389, opened 2022-06-02
Unescape html symbols

jarrel-b:
**Fixes**
* Some strings include html entities, this PR fixes it so that they are unescaped:
`'"'&craz'y name <> '%'*` instead of ` '&#39;&#34;craz&amp;y name &lt;&gt;&#39;'`


Robinnnnn (approved):
spicy

kaitoo1 (approved):

PR #390, opened 2022-06-02
Benny/quantity addition

bennycio:
Changes:

- **Changed** multichain update tokens to account for the total quantity between multiple tokens of the same contract address and token ID


radazen (approved):
🙌

PR #391, opened 2022-06-02
Use indexed query when looking up tokens by wallet ID

radazen:
Create a GIN index on `tokens.owned_by_wallets` and use array operations to query for tokens by wallet ID (because using non-array operations like `ANY` means we won't benefit from the index). Prior to this change, looking up tokens for a wallet took about 300ms. After indexing, it takes about 10ms.


bennycio (approved):

Robinnnnn (commented):
n o i c e

PR #392, opened 2022-06-02
Benny/remove nft table

bennycio:
Changes:

- **Update** docker files to reflect migrations
- **Removed** NFT table and removed any reference to the NFT table in persist.


db/migrations/000011_nft_removal.down..sql
```diff
@@ -0,0 +1,30 @@
+
```
radazen, line 1: Looks like there's an extra dot in this filename!

radazen (approved):
Left two comments, but LGTM aside from a quick filename fix! 🔥

service/persist/token.go
```diff
@@ -203,6 +203,65 @@ type Media struct {
 	MediaType    MediaType  `json:"media_type"`
 }
 
+// NFT represents an old nft throughout the application
+type NFT struct {
```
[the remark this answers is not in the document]
bennycio: It is being used in the migration so will just keep it around until we are merged over 👍🏻

db/migrations/000011_nft_removal.down..sql
bennycio: lol fixing that now

PR #394, opened 2022-06-03
Use collections table instead of collections_v2

radazen:
Most `tokens` queries were using the `collections_v2` table, but we're not actually using that for anything, so this commit drops it and uses `collections` in queries instead. Fixes all sorts of issues we were having with collections (since they were being created in one table and queried from another).

### DIFF
--- a/service/persist/postgres/gallery_token.go
+++ b/service/persist/postgres/gallery_token.go
@@ -57,7 +57,7 @@ func NewGalleryTokenRepository(db *sql.DB, gCache memstore.Cache) *GalleryTokenR
 	c.ID,c.OWNER_USER_ID,c.NAME,c.VERSION,c.DELETED,c.COLLECTORS_NOTE,c.LAYOUT,c.HIDDEN,c.CREATED_AT,c.LAST_UPDATED,
 	n.ID,n.OWNER_USER_ID,n.CHAIN,n.NAME,n.DESCRIPTION,n.TOKEN_TYPE,n.TOKEN_URI,n.TOKEN_ID,n.MEDIA,n.TOKEN_METADATA,n.CONTRACT_ADDRESS,n.CREATED_AT 
 	FROM galleries g, unnest(g.COLLECTIONS) WITH ORDINALITY AS u(coll, coll_ord)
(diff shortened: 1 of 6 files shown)


service/persist/postgres/gallery_token.go
```diff
@@ -77,19 +77,19 @@ func NewGalleryTokenRepository(db *sql.DB, gCache memstore.Cache) *GalleryTokenR
...
+	countCollsStmt, err := db.PrepareContext(ctx, `SELECT COUNT(c.ID) FROM galleries g, unnest(g.COLLECTIONS) WITH ORDINALITY AS u(coll, coll_ord) LEFT JOIN collections c ON c.ID = coll WHERE g.ID = $1 AND c.DELETED = false and g.DELETED = false;`)
 	checkNoErr(err)
 
-	getCollectionsStmt, err := db.PrepareContext(ctx, `SELECT ID FROM collections_v2 WHERE OWNER_USER_ID = $1 AND DELETED = false;`)
+	getCollectionsStmt, err := db.PrepareContext(ctx, `SELECT ID FROM collections WHERE OWNER_USER_ID = $1 AND DELETED = false;`)
 	checkNoErr(err)
 
-	getGalleryCollectionsStmt, err := db.PrepareContext(ctx, `SELECT array_agg(c.ID) FROM galleries g, unnest(g.COLLECTIONS) WITH ORDINALITY AS u(coll, coll_ord) LEFT JOIN collections_v2 c ON c.ID = coll WHERE g.ID = $1 AND c.DELETED = false and g.DELETED = false GROUP BY coll_ord ORDER BY coll_ord;`)
+	getGalleryCollectionsStmt, err := db.PrepareContext(ctx, `SELECT array_agg(c.ID) FROM galleries g, unnest(g.COLLECTIONS) WITH ORDINALITY AS u(coll, coll_ord) LEFT JOIN collections c ON c.ID = coll WHERE g.ID = $1 AND c.DELETED = false and g.DELETED = false GROUP BY coll_ord ORDER BY coll_ord;`)
```
Robinnnnn, line 92: sqlccccc

Robinnnnn (approved):

PR #397, opened 2022-06-05
Fix for updating token collector's notes

radazen:
SQL queries for updating tokens expect `owner_user_id`, but the code paths using these queries were passing an array of wallets. Now they're passing a user ID, and updating collector's notes works again.


Robinnnnn (approved):
amazing!!

PR #399, opened 2022-06-06
delete tokens no longer owned

bennycio:
Changes:

- **Changed** multichain update tokens step to delete tokens that are no longer owned by the user
- **Added** `DeleteByID` to the `tokens_gallery` repo
- **Changed** `deleteByIdentifiersStmt` to update the tokens deleted field to true rather than delete from the database. For the indexer it makes sense to delete the tokens in most cases from the database if they truly do not exist anymore for a given wallet because there is no user inputted data. In the case of the backend, we should do what we do for other database entities and just update the deleted field. This means all other statements that query for tokens also need to adhere to the sql WHERE statement `DELETED = false`. 
- **Added** migration to convert the indexes on `TOKEN_ID, CONTRACT_ADDRESS, OWNER_USER_ID` and `TOKEN_ID, CONTRACT_ADDRESS` to `TOKEN_ID, CONTRACT_ADDRESS, CHAIN, OWNER_USER_ID` and `TOKEN_ID, CONTRACT_ADDRESS, CHAIN` respectively. 

Considerations:

- Because the upsert statement for tokens upserts by token ID, contract address, and user ID, and this is a unique index, when a token is deleted by get's re-added to a user's wallet, the deleted token will become un-deleted in the upsert. Is this alright? Because the upsert statement changes every field, there is no chance that an old field from the token that was deleted will remain. It will always be replaced.


radazen (approved):
One question re: index/upserting, but LGTM otherwise!

PR #400, opened 2022-06-06
contract refreshing in multichain

bennycio:
Changes:

- **Added** step to migration that will get a community for each NFT and upsert the community into the contracts table
- **Added** step to the refreshing of tokens in the `multichain` package that will get the contract for each NFT and update it as well.


radazen (approved):
📜 📄 📝

PR #401, opened 2022-06-07
Ensure added wallets are always new database entities

radazen:
## What's new?

- Adding a wallet to a user's account will always add a new wallet entity now. It should never reuse an existing wallet with the same chain address
- Adding/removing a wallet happens in a transaction now to ensure all steps succeed
- Querying for a user's tokens will no longer return tokens that aren't owned by one of the user's wallets (e.g. a token in a wallet that was removed but not yet synced)
- All database `version` columns now default to 0 so we don't have to manually set a version (which is ideal -- the version of a newly created entity should always be the "current" version by default)


Robinnnnn (approved):
🔥🔥🔥

PR #402, opened 2022-06-08
contract by id and refresh logic

bennycio:
Changes:

- **Changed** multichain provider functions to return contracts with tokens
- **Changed** refresh tokens to retrieve all tokens from all chains before running any updates
- **Changed** token struct and sql schema to use contract DBID instead of contract address
- **Changed** migration to grab from the contracts table
- **Added** `NormalizeAddress` function to chain that will normalize an address as string depending on what chain it is. Currently only supports ETH.

### DIFF
--- a/service/multichain/multichain.go
+++ b/service/multichain/multichain.go
@@ -75,6 +75,12 @@ type chainContracts struct {
 	contracts []ChainAgnosticContract
 }
 
+type tokenIdentifiers struct {
+	chain    persist.Chain
+	tokenID  persist.TokenID
+	contract persist.DBID
+}
+
 // ChainProvider is an interface for retrieving data from a chain
 type ChainProvider interface {
 	GetBlockchainInfo(context.Context) (BlockchainInfo, error)
@@ -174,12 +180,21 @@ func (d *Provider) UpdateTokensForUser(ctx context.Context, userID persist.DBID)
 		return err
 	}
 	addressesToContracts := map[string]persist.DBID{}
+	contractsForChain := map[persist.Chain][]persist.Address{}
 	for _, c := range newContracts {
-		newContract, err := d.ContractRepo.GetByAddress(ctx, c.Address, c.Chain)
+		if _, ok := contractsForChain[c.Chain]; !ok {
+			contractsForChain[c.Chain] = []persist.Address{}
+		}
+		contractsForChain[c.Chain] = append(contractsForChain[c.Chain], c.Address)
+	}
+	for chain, addresses := range contractsForChain {
+		newContracts, err := d.ContractRepo.GetByAddresses(ctx, addresses, chain)
 		if err != nil {
 			return err
 		}
-		addressesToContracts[c.Chain.NormalizeAddress(newContract.Address)] = newContract.ID
+		for _, c := range newContracts {
+			addressesToContracts[c.Chain.NormalizeAddress(c.Address)] = c.ID
+		}
 	}
 
 	newTokens, err := tokensToTokens(ctx, allTokens, addressesToContracts, user)
@@ -188,11 +203,9 @@ func (d *Provider) UpdateTokensForUser(ctx context.Context, userID persist.DBID)
 	}
 
 	// ensure all old tokens are deleted
-	ownedTokens := make(map[string]bool)
+	ownedTokens := make(map[tokenIdentifiers]bool)
 	for _, t := range newTokens {
-
-		ownedTokens[createUniqueTokenString(t.Chain, t.TokenID, t.Contract)] = true
-
+		ownedTokens[tokenIdentifiers{chain: t.Chain, tokenID: t.TokenID, contract: t.Contract}] = true
 	}
 
 	allUsersNFTs, err := d.TokenRepo.GetByUserID(ctx, userID, 0, 0)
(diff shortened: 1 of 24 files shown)


bennycio:
- **Added** contract type to graphQL schema
- **Added** `ContractAPI` to public api
- **Added** resolver for contract

Considerations:

- We now have a community type and a contract type that kind of accomplish the same thing except one is a database entity and one is a conglomeration of database data. Should we consolidate into one? Maybe add the description field of the community onto the contract type and trash the community type in exchange for contract? Is it alright to assume that a contract should have a list of owners like the community currently does? Or should we keep them both because we don't need to have the contract type aware of its owners. Maybe still worth it to add the description column to contract regardless of consolidation.

service/multichain/multichain.go
```diff
@@ -216,6 +211,10 @@ func (d *Provider) UpdateTokensForUser(ctx context.Context, userID persist.DBID)
 	return nil
 }
 
+func createUniqueTokenString(chain persist.Chain, tokenID persist.TokenID, contract persist.DBID) string {
```
radazen, line 214: This may sound like a nitpick, but I think it's actually faster and more memory efficient to create a custom struct as a key here. String formatting is slower and allocates more memory compared to reusing existing string values in a struct, and golang's built-in struct equality seems to work well for map keys.

service/multichain/multichain.go
```diff
@@ -225,74 +224,75 @@ func (d *Provider) VerifySignature(ctx context.Context, pSig string, pNonce stri
...
+			}
 
-		ti := persist.NewTokenIdentifiers(token.ContractAddress, token.TokenID, chain)
+			ti := persist.NewTokenIdentifiers(token.ContractAddress, token.TokenID, chainToken.chain)
 
-		if w, ok := addressToWallets[strings.ToLower(token.OwnerAddress.String())]; ok {
-			if it, ok := seenWallets[ti]; ok {
-				it = append(it, w)
-				seenWallets[ti] = it
+			if w, ok := addressToWallets[chainToken.chain.NormalizeAddress(token.OwnerAddress)]; ok {
```
radazen, line 246: I _think_ this can be simplified because you can append to a zero-value slice, so it could just be

```
if w, ok := addressToWallets[chainToken.chain.NormalizeAddress(token.OwnerAddress)]; ok {
    seenWallets[ti] = append(seenWallets[ti], w)
}
```

Sadly not the case for quantities below, since adding a hex string to an empty string isn't a good idea.

service/persist/postgres/token_gallery.go
```diff
@@ -336,21 +336,17 @@ func (t *TokenGalleryRepository) BulkUpsert(pCtx context.Context, pTokens []pers
...
+			if err := t.deleteTokenUnsafe(pCtx, token.TokenID, token.Contract, token.OwnerUserID, token.Chain); err != nil {
 				return err
 			}
-			if len(pTokens) < i+1 {
-				pTokens = pTokens[:i]
-			} else {
-				pTokens = append(pTokens[:i], pTokens[i+1:]...)
-			}
 		}
+		newTokens[i] = token
```
radazen, line 346: Is there a `continue` missing above this? It looks like `newTokens` will end up with all items from `pTokens` right now.

service/persist/postgres/token_gallery.go
```diff
@@ -513,15 +509,15 @@ func (t *TokenGalleryRepository) DeleteByID(ctx context.Context, id persist.DBID
...
 	_, err := t.deleteByIdentifiersStmt.ExecContext(pCtx, pTokenID, pContractAddress, pOwnerUserID, pChain)
 	return err
 }
 
-func (t *TokenGalleryRepository) dedupTokens(pTokens []persist.TokenGallery) []persist.TokenGallery {
+func (t *TokenGalleryRepository) dedupeTokens(pTokens []persist.TokenGallery) []persist.TokenGallery {
 	seen := map[string]persist.TokenGallery{}
 	for _, token := range pTokens {
-		key := token.ContractAddress.String() + "-" + token.TokenID.String() + "-" + token.OwnerUserID.String()
+		key := token.Contract.String() + "-" + token.TokenID.String() + "-" + token.OwnerUserID.String()
```
radazen, line 520: Same as above re: custom struct vs formatted string for a multi-part key.

radazen (commented):
Looks awesome! Thanks for making all of these changes. Left a few comments, but generally feeling very good about this!

service/multichain/multichain.go
```diff
@@ -225,74 +224,75 @@ func (d *Provider) VerifySignature(ctx context.Context, pSig string, pNonce stri
...
-	res := make([]persist.ContractGallery, len(contracts))
-	for i, contract := range contracts {
-		res[i] = persist.ContractGallery{
-			Chain:          chain,
-			Address:        contract.Address,
-			Symbol:         persist.NullString(contract.Symbol),
-			Name:           persist.NullString(contract.Name),
-			CreatorAddress: contract.CreatorAddress,
+func contractsToContracts(ctx context.Context, contracts []chainContracts) ([]persist.ContractGallery, error) {
+	res := make([]persist.ContractGallery, 0, len(contracts))
```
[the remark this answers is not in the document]
bennycio: definitely! Will add back the dedupe step.

service/multichain/multichain.go
```diff
@@ -126,86 +136,71 @@ func (d *Provider) UpdateTokensForUser(ctx context.Context, userID persist.DBID)
...
+			return err
+		}
+	}
+	newContracts, err := contractsToContracts(ctx, allContracts)
+	if err := d.ContractRepo.BulkUpsert(ctx, newContracts); err != nil {
+		return err
+	}
+	addressesToContracts := map[string]persist.DBID{}
+	for _, c := range newContracts {
+		newContract, err := d.ContractRepo.GetByAddress(ctx, c.Address, c.Chain)
```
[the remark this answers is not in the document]
bennycio: Great idea!

service/multichain/multichain.go
bennycio: Good to know!

PR #403, opened 2022-06-09
Support early access allowlist

radazen:
## What's new?

- Added an `early_access` database table to hold allowlisted EA addresses
- `HasAllowlistNFT()` is now `HasGalleryAccess()` and checks both membership NFTs and the `early_access` database
- We only check membership NFTs and early access addresses when _creating_ an account. We no longer confirm membership for continued access to the site (i.e. the `REQUIRE_NFTS` check was removed)

## What's left to do?

- At some point in the future, we still need to remove the REQUIRE_NFTS entry from the various environment config files in the encrypted deploy directory, but I'd rather hold off on that and avoid the possibility of merge conflicts with the multichain branch.

### DIFF
--- a/db/migrations/000010_create_early_access_table.down.sql
+++ b/db/migrations/000010_create_early_access_table.down.sql
@@ -1,2 +1,1 @@
-DROP INDEX IF EXISTS lowercase_address;
 DROP TABLE IF EXISTS early_access;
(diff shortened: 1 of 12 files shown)


Robinnnnn (approved):
CLEAN

kaitoo1:
WOOO

PR #406, opened 2022-06-09
Sort by newest first when returning tokens for a user or wallet

radazen:
## What's new?

In the pre-multichain world, we returned a list of NFTs per wallet without any explicit sorting, but the returned NFTs seemed to be mostly sorted anyway (probably due to the way Postgres stores rows). This is no longer the case in the post-multichain world, so this PR adds explicit sorting when querying for a wallet's tokens or a user's tokens.


kaitoo1 (approved):

kaitoo1:
nice!

kaitoo1:
oops fat fingered and closed by accident, reopened

PR #411, opened 2022-06-16
[GAL-132] Imgix support

radazen:
_**Note: Using imgix requires an API key, even for local development. This PR includes a new "local secrets" file in the _internal directory. Make sure you decrypt the new _internal directory before running a build with these changes!**_

## What's new?

- Use imgix to generate URLs for all preview images (in thumbnail, small, medium, and large variants)
- Remove size variants for `ImageMedia` types, since we only ever want the highest resolution for the actual image (as opposed to previews, which will have size variants)

## What's missing?

- We still need to add srcSet preview URLs for high-resolution/retina devices, but that's not a requirement for getting us onto imgix initially.


Robinnnnn (approved):
🔥🔥🔥

PR #414, opened 2022-06-20
Always update token entries, regardless of block number

radazen:
The SQL statement for bulk upserting tokens was only updating existing tokens if the incoming token had a higher block number, but that logic was actually meant for the indexer. For our purposes with updating tokens, we always want to do the update -- otherwise we'd miss cases like adding tokens with lower block numbers when connecting a wallet.


jarrel-b (approved):
nice catch!

PR #415, opened 2022-06-20
Disable GraphQL API during maintenance

radazen:
This will disable GraphQL access during tonight's maintenance so we won't have to worry about anyone accidentally modifying data during migration.


jarrel-b (commented):
👍

PR #416, opened 2022-06-20
no autogenerated IDs to prevent empty ID edge cases

bennycio:
Change:

- **Removed** `GenerateID()` from the `(DBID).Value()` function so that it only returns the string version of the DBID rather than automatically generating a new ID when it is empty. Generating the ID when the current DBID is empty leads to some unexpected effects that are hard to track down because there is nothing explicit about the `driver.Valuer` interface that would require the DBID to not be empty.
- **Changed** token persist funcs to explicitly generate IDs instead of relying on auto-generation.

I went through each insert/create func in the persist package to make sure they all explicitly generated IDs and the only ones that I found that didn't were two of the unused token funcs in both `token` and `token_gallery`.


radazen (approved):
🚀 LGTM! Thanks for tracking this down!

PR #421, opened 2022-06-22
refresh individual tokens and contracts

bennycio:
Changes:

- **Added** endpoint to indexer for refreshing contracts manually
- **Added** two functions to the `ChainProvider` interface for refreshing individual tokens and contracts as well as implemented those functions for the opensea and indexer chain providers.

Considerations:

- We need the frontend to call these functions somewhere so they probably need to have their own graph QL endpoints, so we will follow this PR up with another PR for graph QL endpoints.
- Should the refreshing of tokens and contracts be combined into one `ChainProvider` function where the contract for a given token is refreshed when a token is refreshed. We could do this at the graphQL level as well if we wanted to and have one "refresh" endpoint that refreshes the contract for a token and the token. I also imagine that there could be a refresh button on individual NFT pages for refreshing tokens and also a button on community pages for refreshing a contract.

### DIFF
--- a/graphql/resolver/schema.resolvers.go
+++ b/graphql/resolver/schema.resolvers.go
@@ -298,13 +298,53 @@ func (r *mutationResolver) UpdateTokenInfo(ctx context.Context, input model.Upda
 func (r *mutationResolver) RefreshTokens(ctx context.Context) (model.RefreshTokensPayloadOrError, error) {
 	api := publicapi.For(ctx)
 
-	err := api.Token.RefreshTokens(ctx)
+	err := api.Token.RefreshTokensForUser(ctx)
 	if err != nil {
 		return nil, err
 	}
 
 	output := &model.RefreshTokensPayload{
 		Viewer: resolveViewer(ctx),
+	}
+
+	return output, nil
+}
+
+func (r *mutationResolver) RefreshTokenMetadata(ctx context.Context, tokenID persist.DBID) (model.RefreshTokenMetadataPayloadOrError, error) {
+	api := publicapi.For(ctx)
+
+	err := api.Token.RefreshToken(ctx, tokenID)
+	if err != nil {
+		return nil, err
+	}
+
+	token, err := resolveTokenByTokenID(ctx, tokenID)
+	if err != nil {
+		return nil, err
+	}
+
+	output := &model.RefreshTokenMetadataPayload{
+		Token: token,
+	}
+
+	return output, nil
+}
+
+func (r *mutationResolver) RefreshContractMetadata(ctx context.Context, contractID persist.DBID) (model.RefreshContractMetadataPayloadOrError, error) {
+	api := publicapi.For(ctx)
+
+	err := api.Contract.RefreshContract(ctx, contractID)
+	if err != nil {
+		return nil, err
+	}
+
+	contract, err := resolveContractByContractID(ctx, contractID)
+	if err != nil {
+		return nil, err
+	}
+
+	output := &model.RefreshContractMetadataPayload{
+		Contract: contract,
 	}
 
 	return output, nil
--- a/graphql/resolver/schema.resolvers.go
+++ b/graphql/resolver/schema.resolvers.go
@@ -295,22 +295,22 @@ func (r *mutationResolver) UpdateTokenInfo(ctx context.Context, input model.Upda
 	return output, nil
 }
 
-func (r *mutationResolver) RefreshTokens(ctx context.Context) (model.RefreshTokensPayloadOrError, error) {
+func (r *mutationResolver) SyncTokens(ctx context.Context) (model.SyncTokensPayloadOrError, error) {
 	api := publicapi.For(ctx)
 
-	err := api.Token.RefreshTokensForUser(ctx)
+	err := api.Token.SyncTokens(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	output := &model.RefreshTokensPayload{
+	output := &model.SyncTokensPayload{
 		Viewer: resolveViewer(ctx),
 	}
 
 	return output, nil
 }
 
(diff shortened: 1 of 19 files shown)


service/persist/postgres/contract.go
```diff
@@ -27,7 +28,10 @@ func NewContractRepository(db *sql.DB) *ContractRepository {
 	upsertByAddressStmt, err := db.PrepareContext(ctx, `INSERT INTO contracts (ID,VERSION,ADDRESS,SYMBOL,NAME,LATEST_BLOCK,CREATOR_ADDRESS) VALUES ($1,$2,$3,$4,$5,$6,$7) ON CONFLICT (ADDRESS) DO UPDATE SET VERSION = $2,ADDRESS = $3,SYMBOL = $4,NAME = $5,LATEST_BLOCK = $6,CREATOR_ADDRESS = $7;`)
 	checkNoErr(err)
 
-	return &ContractRepository{db: db, getByAddressStmt: getByAddressStmt, upsertByAddressStmt: upsertByAddressStmt}
+	updateByAddressStmt, err := db.PrepareContext(ctx, `UPDATE contracts SET NAME = $2, SYMBOL = $3, CREATOR_ADDRESS = $4, LATEST_BLOCK = $5 WHERE ADDRESS = $1;`)
```
jarrel-b, line 31: Maybe we can also update `LAST_UPDATED` so we can tell when the last refresh was?

service/multichain/opensea/opensea.go
```diff
@@ -568,6 +568,16 @@ func contractToContract(ctx context.Context, openseaContract Contract, ethClient
 	}, nil
 }
 
+// RefreshToken refreshes the metadata for a given token.
+func (p *Provider) RefreshToken(context.Context, multichain.ChainAgnosticIdentifiers) error {
+	return nil
+}
+
+// RefreshContract refreshses the metadata for a contract
+func (p *Provider) RefreshContract(context.Context, persist.Address) error {
```
jarrel-b, line 577: Is this bc there's not really a way to refresh a token and contract via OS or we don't want OS overriding token data?

service/multichain/eth/eth.go
```diff
@@ -185,7 +220,42 @@ func (d *Provider) UpdateMediaForWallet(ctx context.Context, wallet persist.Addr
...
+	}
+
+	res, err := d.httpClient.Do(req)
+	if err != nil {
+		return err
+	}
+
+	defer res.Body.Close()
+
+	if res.StatusCode >= 299 || res.StatusCode < 200 {
```
jarrel-b, line 235: was wondering if its enough to check that the status was only `200`, or are other `2xx` statuses the indexer may return ok?

service/multichain/eth/eth.go
```diff
@@ -185,7 +220,42 @@ func (d *Provider) UpdateMediaForWallet(ctx context.Context, wallet persist.Addr
...
+		if err != nil {
+			return err
+		}
+		return fmt.Errorf("unexpected status: %s | err: %s ", res.Status, errResp.Error)
+	}
+
+	return nil
+}
+
+// RefreshContract refreshses the metadata for a contract
```
jarrel-b, line 247: sorry super nitpicky! typo `refreshses`

jarrel-b (commented):
looks good! just had a few questions, mostly because I'm not very familiar with the indexer

indexer/contracts.go
```diff
@@ -1,8 +1,95 @@
...
+		up := persist.ContractUpdateInput{
+			Name:        persist.NullString(newMetadata.Name),
+			Symbol:      persist.NullString(newMetadata.Symbol),
+			LatestBlock: persist.BlockNumber(latestBlock),
+		}
+
+		timedContext, cancel := context.WithTimeout(c, time.Second*10)
+		defer cancel()
+
+		creator, err := rpc.GetContractCreator(timedContext, input.Address, ethClient)
```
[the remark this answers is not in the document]
bennycio: Hahah yep, after a lot of research this is the only solution I could find. A good ol getContractCreator JSON RPC method would have been nice.

service/persist/postgres/contract.go
bennycio: Will do!

service/multichain/opensea/opensea.go
bennycio: Opensea handles the refreshing on their end with the buttons on their webpages. When we call tokens from them we can already expect them to be refreshed.

service/multichain/eth/eth.go
bennycio: That is true, especially given that we know what the indexer can return because we built it lol we should just use 200. Thanks!

service/multichain/opensea/opensea.go
jarrel-b: Makes sense!

graphql/resolver/schema.resolvers.go
```diff
@@ -310,6 +310,46 @@ func (r *mutationResolver) RefreshTokens(ctx context.Context) (model.RefreshToke
 	return output, nil
 }
 
+func (r *mutationResolver) RefreshTokenMetadata(ctx context.Context, tokenID persist.DBID) (model.RefreshTokenMetadataPayloadOrError, error) {
```
jarrel-b, line 313: I believe @radazen mentioned for the public api, the names could follow a convention DoSomethingbySomething so something like RefreshTokenMetadataByTokenID because presumably there could be a RefreshTokenMetadataByTokenName, although probably not the best example lol

jarrel-b (approved):
Looks good to me!

radazen (approved):
Left a few comments about method names, but everything looks good to me from a functional standpoint! 🔥 🎉

graphql/resolver/schema.resolvers.go
bennycio: Hmmm maybe I could change it in the public API and multichain interface then because this is different than what `RefreshTokens` currently does. This specifically refreshes the metadata for a token where as `RefreshTokens` refreshes what tokens a user has.

PR #431, opened 2022-07-01
Init pgx client once

jarrel-b:
**Changes**
* Noticed half the response time for handling messages is spent creating a connection pool to the db because it is done per request, changed it so the connection is made once at startup (like we do in the backend)


radazen (approved):
🚀📈🚀

Robinnnnn (approved):

PR #435, opened 2022-07-08
Group follow events together

jarrel-b:
Adds the `UserFollowedByUsers` feed event type which groups user events that followed the same user together. 

The feed writes the new event alongside the existing `UserFollowedUsers` event temporarily so the frontend has some data to work with. Then will follow up to stop generating `UserFollowedUsers` events and instead display the new `UserFollowedByUsers` in the feed once the frontend supports it.

This PR also removes the `owner` as a required field from the `FeedEventData` interface since there's not really a notion of a single `owner` for events of this type. Maybe was a bit overzealous in thinking of the interface, but luckily the frontend queries already specify `owner` per event type so no changes should be needed on that side. 

Also removed the old event tables while I'm at it, and renamed some existing feed functions to better reflect what they filter on.

Updated the docs on Notion as well: https://www.notion.so/gallery-so/Event-Design-eaf7607654024d54894cb19741904280

### DIFF
--- a/db/queries/query.sql
+++ b/db/queries/query.sql
@@ -248,10 +248,7 @@ WITH cursors AS (
     AND event_time > (SELECT cur_after FROM cursors)
     AND event_time < (SELECT cur_before FROM cursors)
     AND fe.deleted = false AND fl.deleted = false
-
-    -- Can UNION ALL because a user can't follow themselves
     UNION ALL
-
     SELECT id FROM feed_events
     WHERE event_time > (SELECT cur_after FROM cursors)
     AND event_time < (SELECT cur_before FROM cursors)
@@ -278,10 +275,7 @@ SELECT
         WHERE event_time > (SELECT event_time FROM feed_events f WHERE f.id = $2)
         AND fe.deleted = false AND fl.deleted = false
         LIMIT 1)
-
-        -- Can UNION ALL because a user can't follow themselves
         UNION ALL
-
         (SELECT 1
         FROM feed_events
         WHERE event_time > (SELECT event_time FROM feed_events f WHERE f.id = $2)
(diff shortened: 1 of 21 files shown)


db/queries/query.sql
```diff
@@ -194,6 +215,8 @@ WITH cursors AS (
     WHERE event_time > (SELECT cur_after FROM cursors)
     AND event_time < (SELECT cur_before FROM cursors)
     AND deleted = false
+    -- Remove when frontend can support this type
+    AND action != 'UserFollowedByUsers'
```
radazen, line 219: What does the frontend currently do if it encounters an unknown/unsupported action type? Kind of feels like it should ignore unknown types to make adding new types a little smoother.

Robinnnnn: we return nothing!

Robinnnnn: https://github.com/gallery-so/gallery/blob/main/src/components/Feed/FeedEvent.tsx#L109-L110

db/queries/query.sql
```diff
@@ -272,6 +306,9 @@ SELECT * FROM feed_events WHERE id = $1 AND deleted = false;
 -- name: CreateFeedEvent :one
 INSERT INTO feed_events (id, owner_id, action, data, event_time, event_ids) VALUES ($1, $2, $3, $4, $5, $6) RETURNING *;
 
+-- name: CreateFeedEventNoOwner :one
```
radazen, line 309: Is there a practical difference between `CreateFeedEvent` called with a `null` owner and `CreateFeedEventNoOwner`, or is the latter a convenience method? IIRC you may be able to leave comments in the sqlc query file and have them show up as documentation in the generated methods, such that you could document that `null` is an okay value to pass to `CreateFeedEvent`. The clarity of the `NoOwner` version is nice, though! Would you expect any other variants like that to pop up in the future?

db/queries/query.sql
jarrel-b: Nice! Since we're writing both `UserFollowedByUsers` and `UserFollowedUsers` at the moment, the worst case is we get a page that's only follow events, which only half is supported by the frontend so decided to exclude the event for now.

db/queries/query.sql
jarrel-b: Ah this is me fighting with sqlc because it was passing an empty string instead of `nil/null` which broke the foreign key constraint on that column. I realized there is a NullString type that I think should do the trick

radazen (approved):
👍 👍 👍

bennycio (approved):
🔥

jarrel-b:
Closing this PR for now since this merits a larger discussion

PR #436, opened 2022-07-08
Add asUsername parameter to GraphQL debug auth

radazen:
When debugging user issues, I usually log in as that user locally (via debug auth) and see if I can reproduce the issue. Historically, that means looking up their userID and their addresses in the database and then supplying those to a debug authenticator.

This PR adds the `asUsername` parameter as a convenience method to log in as a given user without needing to look up their ID and addresses manually.

### DIFF
--- a/graphql/generated/generated.go
+++ b/graphql/generated/generated.go
@@ -3388,11 +3388,23 @@ input EoaAuth {
     signature: String! @scrub
 }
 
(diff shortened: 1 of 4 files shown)


jarrel-b (approved):
neat!

Robinnnnn (approved):
what happens if there are multiple users with that username? hehe... jk...

PR #437, opened 2022-07-11
[GAL-281] [GAL-271] Improved CreateUser flow

radazen:
## What's new?

- The `CreateUser` mutation now requires a username to be specified. Prior to this, we'd create an account and _then_ allow the user to pick a username, leading to all sorts of scenarios where a user can take actions on our site but doesn't have a username.

- `CreateUser` also takes an optional `bio` field, since we still ask for a bio during the initial sign-up flow. If omitted, the bio will be an empty string as usual.

- Bug fix: `CreateUser` now succeeds when using an address owned by a deleted wallet or a deleted user. Previously, user creation would fail if a previous (but now deleted) wallet or user had used the same address.

### DIFF
--- a/service/persist/postgres/user.go
+++ b/service/persist/postgres/user.go
@@ -156,33 +156,73 @@ func (u *UserRepository) UpdateByID(pCtx context.Context, pID persist.DBID, pUpd
 
 }
 
+func (u *UserRepository) createWalletWithTx(ctx context.Context, tx *sql.Tx, chainAddress persist.ChainAddress, walletType persist.WalletType) (persist.DBID, error) {
+	// Do we already have a wallet with this ChainAddress?
+	var walletID persist.DBID
+	if err := tx.StmtContext(ctx, u.getWalletIDStmt).QueryRowContext(ctx, chainAddress.Address(), chainAddress.Chain()).Scan(&walletID); err != nil && err != sql.ErrNoRows {
+		return "", err
+	}
+
+	if walletID != "" {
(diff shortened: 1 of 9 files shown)


service/persist/postgres/user.go
```diff
@@ -156,25 +156,76 @@ func (u *UserRepository) UpdateByID(pCtx context.Context, pID persist.DBID, pUpd
...
+
+		// If the wallet exists but doesn't belong to anyone, it should be deleted.
+		if _, err := tx.StmtContext(ctx, u.deleteWalletStmt).ExecContext(ctx, walletID); err != nil {
+			return "", err
+		}
+	}
+
+	newWalletID := persist.GenerateID()
+	// At this point, we know there's no existing wallet in the database with this ChainAddress, so let's make a new one!
+	if _, err := tx.StmtContext(ctx, u.createWalletStmt).ExecContext(ctx, newWalletID, chainAddress.Address(), chainAddress.Chain(), walletType); err != nil {
```
jarrel-b, line 187: Is there a race condition where in between checking if an address exists and then writing the new address that another user could write to the db the same address and we end up with two of the same address? I guess it's pretty unlikely, particularly with addresses but wanted to bring it up!

I think there is a similar issue with usernames but you may have brought that up already,

radazen: There's a unique index on wallet chain+address, so the worst case should be that we'd fail to insert and the whole transaction would be aborted/rolled back. There's still definitely a possibility that we return an unhelpful/unhandled error (unique index failure) vs a helpful one (user exists with this address), though.

PR #439, opened 2022-07-11
Move user created trigger

jarrel-b:
**Changes**
* Moves UserCreated trigger from `UpdateUserInfo` to `CreateInfo`


radazen (approved):
🙌 Thanks!

PR #442, opened 2022-07-14
Add figure31 integration

jarrel-b:
**Changes**
* Adds a goroutine that manages the Figure31 Marks collection
    * Listens for transfer events on the contract and will replace transferred tokens with whitespace 
    * Also add purchasers to the early access list (it doesn't check if they're already on Gallery, but think thats ok)

**Next steps**
* Need to update `_deploy` to include the Figure31 details once the collection is created
* May add a command to refresh the collection manually when we need to


bennycio (approved):
Love the combination of scheduled and subscribed, although that subscription is going to cost us buku bucks haha

Looks good to me other than that `SubscribeFilterLogs` call.

Robinnnnn (approved):
🔥

PR #444, opened 2022-07-15
Jarrel/add feed block list

jarrel-b:
**Changes**
* Adds blocklist to block events from specific user actions


Robinnnnn (approved):
sickkkk

PR #448, opened 2022-07-20
Reference correct nonce field for Gnosis Safe auth

radazen:
Gnosis auth was mistakenly using the Eoa.Nonce field instead of the GnosisSafe.Nonce field, causing a nil reference error that was reported via Sentry.


jarrel-b (approved):
nice catch!

Robinnnnn (approved):

PR #450, opened 2022-07-20
Remove figure31 routine

jarrel-b:
**Changes**
* Removes figure31 specific env vars from prod deploy
* Remove figure31 routine


Robinnnnn (approved):
that was fun!

PR #451, opened 2022-07-21
begin tezos provider

bennycio:
Changes:

- **Added** tezos multichain provider with functions for getting tokens and contracts

TODO:

- [x] Signature verification


service/multichain/opensea/opensea.go
```diff
@@ -215,6 +215,133 @@ func (p *Provider) ValidateTokensForWallet(context.Context, persist.Address, boo
 	return nil
 }
 
+// RefreshToken refreshes the metadata for a given token.
```
bennycio, line 218: disregard this, just moved this all up a bit for consistency across multichain provider go files

service/multichain/tezos/tezos.go
```diff
@@ -0,0 +1,550 @@
...
+		storageClient:  storageClient,
+		tokenBucket:    tokenBucket,
+	}
+}
+
+// GetBlockchainInfo retrieves blockchain info for ETH
+func (d *Provider) GetBlockchainInfo(ctx context.Context) (multichain.BlockchainInfo, error) {
+	return multichain.BlockchainInfo{
+		Chain:   persist.ChainTezos,
+		ChainID: 0,
```
radazen, line 164: What does `ChainID` represent?

service/multichain/tezos/tezos.go
```diff
@@ -0,0 +1,550 @@
...
+	if err != nil {
+		return false, err
+	}
+	nonce := tezosNoncePrepend + auth.NewNoncePrepend + pNonce
+	asBytes := []byte(nonce)
+	asHex := hex.EncodeToString(asBytes)
+	lenHexBytes := []byte(fmt.Sprintf("%d", len(asHex)))
+
+	asBytes = append(lenHexBytes, asBytes...)
+	asBytes = append([]byte{0x05, 0x01, 0x00}, asBytes...)
```
radazen, line 310: What's the significant of these bytes? It might be helpful to put in some comments explaining exactly what's going on in the verification flow here!

service/multichain/tezos/tezos.go
```diff
@@ -0,0 +1,550 @@
...
+		seen[id.String()] = t
+	}
+	for _, t := range seen {
+		result = append(result, t)
+	}
+	return result
+}
+
+func (d *Provider) getPublicKeyFromAddress(ctx context.Context, address string) (persist.Address, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("%s/v1/accounts/%s", d.apiURL, address), nil)
```
radazen, line 483: Huh, this is interesting. What's the relationship between a Tezos public key and a Tezos address? Is there a way to verify that the public key returned from the provider does indeed match the address we supplied?

radazen (approved):
Looks good! Left some questions/requests/comments, but nothing major. TEZOOOOOOOOOOOS!

ꜩ ꜩ ꜩ ꜩ ꜩ ꜩ ꜩ ꜩ ꜩ ꜩ

service/multichain/tezos/tezos.go
bennycio: It is impossible to go from address to public key but it is easy to verify if we have both because the address is just a hash of the public key.

Robinnnnn (approved):
yay

PR #453, opened 2022-07-25
Add Sentry error handling to the indexer

jarrel-b:
**Changes**
* Adds Sentry to the indexer
* Adds Sentry to the indexer servers
* Adds Sentry logging hook that can be enabled via `logger.AddHook(sentryutil.SentryLoggerHook)` (only enabled for the indexer atm)
    * Log levels > `INFO` get automatically reported as an error to Sentry
    * Log levels <= `INFO` get automatically reported as a breadcrumb to an error that gets reported

**Follow ups**
* Doesn't handle tracing of the indexer yet i.e. traces for writes/reads of objects to GCP or timing of RPC calls or DB writes/reads but we do get HTTP traces for server requests because of the `Tracing` middleware


Robinnnnn (approved):
🔥🔥🔥🔥🔥

indexer/nfts.go
```diff
@@ -466,7 +481,10 @@ func updateTokens(tokenRepository persist.TokenRepository, ethClient *ethclient.
 // refreshToken will find all of the media content for an addresses NFTs and possibly cache it in a storage bucket
 func refreshToken(c context.Context, input UpdateTokenMediaInput, tokenRepository persist.TokenRepository, ethClient *ethclient.Client, ipfsClient *shell.Shell, arweaveClient *goar.Client, storageClient *storage.Client, tokenBucket string) error {
 	if input.TokenID != "" && input.ContractAddress != "" {
-		logrus.Infof("updating media for token %s-%s", input.TokenID, input.ContractAddress)
+		logger.For(c).WithFields(logrus.Fields{
```
radazen, line 484: This is an interesting case. The `logger.For(c)` pattern is mostly useful because it lets you attach fields to a context and have them printed in subsequent logging calls, too. For a case like this, you could add the fields to the context (e.g. `ctx = logger.NewContextWithFields(c, theseFields)`) and pass that to `tokenRepository.GetByTokenIdentifiers` so it includes the token ID and contract address in any logging it does.

But I'm wondering how that interacts with breadcrumb logging. If we logged these fields here _and_ attached them to the context, would a subsequent log statement that includes these fields add duplicate breadcrumbs?

service/logger/logger.go
```diff
@@ -21,7 +24,7 @@ func SetLoggerOptions(optionsFunc func(logger *logrus.Logger)) {
 
 func For(ctx context.Context) *logrus.Entry {
 	if ctx == nil {
-		return defaultEntry
+		return defaultEntry.WithContext(ctx)
```
radazen, line 27: This is a no-op, right? We might want to leave out the `WithContext(ctx)` part (or maybe use `WithContext(nil)` explicitly?) to make it obvious that if we've reached this line, `ctx` doesn't actually contain anything.

middleware/middleware.go
```diff
@@ -181,7 +181,9 @@ func ErrLogger() gin.HandlerFunc {
 	return func(c *gin.Context) {
 		c.Next()
 		if len(c.Errors) > 0 {
-			logger.For(c).Errorf("%s %s %s %s %s", c.Request.Method, c.Request.URL, c.ClientIP(), c.Request.Header.Get("User-Agent"), c.Errors.JSON())
+			err := logger.GinErrorLoggerErr{Context: c.Copy()}
```
radazen, line 184: What's the benefit of `logger.GinErrorLoggerErr` compared to `Errorf` here?

service/sentry/sentry.go
```diff
@@ -182,3 +229,105 @@ func SentryHubFromContext(ctx context.Context) *sentry.Hub {
...
+		return nil
+	}
+	if hub := SentryHubFromContext(entry.Context); hub != nil {
+		switch isErr := entry.Level <= logrus.ErrorLevel; isErr {
+		// Send as an error
+		case true:
+			if scope := hub.Scope(); scope == nil {
+				hub.PushScope()
+			}
+			defer hub.PopScope()
```
radazen, line 261: If `hub.Scope()` above _wasn't_ nil, wouldn't we end up popping the current/existing scope here?

bennycio (approved):
Wonderful!

radazen (approved):
Left some questions and comments, but looks great to me!

service/sentry/sentry.go
jarrel-b: doh yes it will, will fix!

middleware/middleware.go
jarrel-b: This was so we can avoid reporting an error twice to Sentry if the gin `ErrLogger ` middleware and the logger `SentryLoggerHook` are enabled. Just realized though that we'll also run into the same issue if the gin `Sentry` middleware is also enabled. 

I'll update so that the logger hook is enabled only for the indexer, and not for the indexer servers since those errors are well covered.

service/logger/logger.go
jarrel-b: Good call! will change to `WithContext(nil)` so that its explicit

indexer/nfts.go
jarrel-b: Hmm yea we would get the fields in subsequent log statements (the log messages themselves would be different, but it would be the same fields attached to the log message). Is the worry about duplicated fields or something else?

jarrel-b: Ah now I understand what you meant for this particular case, will update to use `NewContextWithFields` 👍

radazen: FWIW, I wasn't really visualizing breadcrumbs correctly when I wrote the above! For whatever reason, I was thinking of it as a key/value type thing instead of a sequential list of entries. I was wondering what would happen to breadcrumbs if we kept writing log messages that included the same fields. But the fields would just be part of the log message, which itself would be a breadcrumb, right?

middleware/middleware.go
radazen: Aah, I see! I missed the handling that skips logging based on `GinErrorLoggerErr` type checks. No objections here to doing it that way (or the current way, which also looks good to me!), just wasn't sure what we gained by using a separate type, but I see it now!

radazen:
Ship ship ship iiiiiiiiiiit 🚢 🚢 🚢

indexer/nfts.go
jarrel-b: Yup exactly!

PR #454, opened 2022-07-26
Fix CollectionToken node id

jarrel-b:


graphql/model/models.go
```diff
@@ -13,11 +13,11 @@ type GqlID string
 var Cursor connCursor
 
 func (r *CollectionToken) GetGqlIDField_TokenID() string {
-	return r.TokenId.String()
+	return r.HelperCollectionTokenData.TokenId.String()
```
jarrel-b, line 16: Added back the forwarding calls because it was confusing for me to debug haha

Robinnnnn (approved):

graphql/schema/schema.graphql
```diff
@@ -258,7 +258,7 @@ type OwnerAtBlock {
 
 type CollectionToken implements Node
 @goEmbedHelper
-@goGqlId(fields: ["tokenId", "collectionId"]) {
+@goGqlId(fields: ["collectionId", "tokenId"]) {
```
radazen, line 261: Ordering matters here (though I suppose I should document that, huh?) -- swapping these means the generated `NodeFetcher.OnCollectionToken` function is going to change the order of its arguments as well, so you'd also need to change the parameter order in the binding method for `OnCollectionToken` at the top of `schema.resolvers.helpers.go`. Something like:

```
OnCollectionToken: func(ctx context.Context, collectionId string, tokenId string) (*model.CollectionToken, error) {
    return resolveCollectionTokenByIDs(ctx, persist.DBID(tokenId), persist.DBID(collectionId))
},
```


More than anything, I guess I should document the fact that changing the ordering after methods have been generated can be dangerous. The initial ordering doesn't matter much: generated methods will have useful parameter names to make it obvious what should go where. But if the parameters have the same type (and thus avoid compiler errors when their positions are swapped), changing their ordering later on can be problematic.

radazen (changes requested):

graphql/schema/schema.graphql
jarrel-b: Ah thanks for clarifying was wondering why there were no compiler errors but that makes total sense

jarrel-b: I think I'll just keep it how it was because these are meant to be opaque ids anyway?

radazen: Sounds good to me! I think I actually went through this same process. The type was originally called `GalleryToken`, so it kind of felt right for the `token` to be the leading thing. Then we renamed it to `CollectionToken` and it felt like the collection should be first, so I started changing the ordering...and then realized I'd need to change it elsewhere and decided not to bother.

radazen (approved):
🙌🙌🙌

PR #457, opened 2022-07-27
index tokens called for but don't exist

bennycio:
Changes:

- **Updated** refresh token function to manually index a token if it is not found in the database upon refreshing.


indexer/nfts.go
```diff
@@ -636,3 +645,34 @@ func getUpdateForToken(pCtx context.Context, uniqueHandlers uniqueMetadatas, tok
...
+	}
+	e1155, err = contracts.NewIERC1155Caller(contractAddress.Address(), ec)
+	if err != nil {
+		return
+	}
+	t.TokenType = persist.TokenTypeERC721
+	owner, err := e721.OwnerOf(&bind.CallOpts{Context: pCtx}, tokenID.BigInt())
+	isERC721 := err == nil
+	if isERC721 {
+		t.OwnerAddress = persist.EthereumAddress(owner.String())
```
jarrel-b, line 665: Was wondering would `owner` be the current wallet that is signed in, or is always the current owner of the token (like their other wallet)?

service/media/media.go
```diff
@@ -300,7 +300,8 @@ func cacheRawSvgMedia(ctx context.Context, img []byte, bucket, fileName string,
 
 	// Update the object to set the metadata.
 	objectAttrsToUpdate := storage.ObjectAttrsToUpdate{
-		ContentType: "image/svg+xml",
+		ContentType:  "image/svg+xml",
+		CacheControl: "no-store",
```
jarrel-b, line 304: just to check my understanding, we want `no-store` because otherwise refreshes will show the old assets until the cache is bust?

jarrel-b (approved):

indexer/nfts.go
bennycio: Yep, there is a rare chance they no longer own this token. Since this is at the indexer server level as well, we cannot prove that the caller is indeed the owner without doing this. The caller could send any address they want technically.

service/media/media.go
bennycio: Yep!

Robinnnnn (approved):

server/server.go
```diff
@@ -72,7 +72,7 @@ func CoreInit(pqClient *sql.DB, pgx *pgxpool.Pool) *gin.Engine {
 
 	repos := newRepos(pqClient)
 	ethClient := newEthClient()
-	httpClient := &http.Client{Timeout: 10 * time.Second}
+	httpClient := &http.Client{Timeout: 10 * time.Minute}
```
jarrel-b, line 75: Would the auth stuff use this client as well? Wondering how the frontend would handle this if the provider is actually down, guess it would hang for 10 minutes?

jarrel-b (approved):

indexer/indexer.go
```diff
@@ -1057,7 +1079,15 @@ func upsertTokensAndContracts(ctx context.Context, t []persist.Token, tokenRepo
 			}
 			err := tokenRepo.BulkUpsert(ctx, t[i:end])
 			if err != nil {
-				return err
+				if strings.Contains(err.Error(), "deadlock detected (SQLSTATE 40P01)") {
+					logger.For(ctx).Errorf("Deadlock detected, retrying upsert")
+					time.Sleep(time.Second)
```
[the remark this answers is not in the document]
bennycio: 1 second, and we could wait longer sure! I'll up it to 5

server/server.go
bennycio: The multichain auth stuff will use this client, but currently none of the multichain auth code makes HTTP calls, but we can also control the timeout on a request with contexts as well. This ten minute limit I'd say is more of a fallback if the context is not going to timeout at some point.

PR #464, opened 2022-07-29
Benny/minor metadata fixes

bennycio:
Changes:

Minor fixes to the following metadata types so far:

- HTML
- GIF
- Autoglyphs
- Colorglyphs

### DIFF
--- a/service/media/media.go
+++ b/service/media/media.go
@@ -121,9 +121,7 @@ func MakePreviewsForMetadata(pCtx context.Context, metadata persist.TokenMetadat
 	switch mediaType {
 	case persist.MediaTypeImage:
 		res = getImageMedia(pCtx, name, tokenBucket, storageClient, vURL, imgURL)
-	case persist.MediaTypeGIF:
-		res = getGIFMedia(pCtx, name, tokenBucket, storageClient, vURL, imgURL)
-	case persist.MediaTypeVideo, persist.MediaTypeAudio, persist.MediaTypeHTML, persist.MediaTypeText, persist.MediaTypeAnimation:
+	case persist.MediaTypeVideo, persist.MediaTypeAudio, persist.MediaTypeHTML, persist.MediaTypeText, persist.MediaTypeAnimation, persist.MediaTypeGIF:
 		res = getAuxilaryMedia(pCtx, name, tokenBucket, storageClient, vURL, imgURL, mediaType)
 	case persist.MediaTypeSVG:
 		res = getSvgMedia(pCtx, name, tokenBucket, storageClient, vURL, imgURL)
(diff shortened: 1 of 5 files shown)


Robinnnnn (approved):
🙏🙏🙏🙏

service/media/media.go
```diff
@@ -410,13 +420,15 @@ func downloadAndCache(pCtx context.Context, url, name, ipfsPrefix string, ipfsCl
 
 outer:
 	switch mediaType {
-	case persist.MediaTypeVideo, persist.MediaTypeGIF, persist.MediaTypeUnknown, persist.MediaTypeSVG:
+	case persist.MediaTypeVideo, persist.MediaTypeUnknown, persist.MediaTypeSVG:
 		break outer
 	default:
 		switch asURI.Type() {
 		case persist.URITypeIPFS, persist.URITypeArweave:
+			logger.For(pCtx).Infof("uri for %s is of type %s: trying to cache", name, asURI.Type())
```
jarrel-b, line 428: Mind explaining this bit to me? Is it bc we don't store IPFS and arweave media to gcloud so there's nothing to delete?

service/media/media.go
```diff
@@ -348,7 +350,11 @@ func cacheRawAnimationMedia(ctx context.Context, animation []byte, bucket, fileN
 		return err
 	}
 
-	return mediamapper.PurgeImage(ctx, fmt.Sprintf("https://storage.googleapis.com/%s/%s", bucket, fileName))
+	err = mediamapper.PurgeImage(ctx, fmt.Sprintf("https://storage.googleapis.com/%s/%s", bucket, fileName))
+	if err != nil {
+		logger.For(ctx).Errorf("could not purge image %s: %s", fileName, err)
```
jarrel-b, line 355: Sorry not very familiar with imgix! `PurgeImage` looks like an API call to imgix, but we pass it a `googleapis.com` URL, does that mean we delete it from imgix's servers or from our bucket? If the call fails do users get the old media still?

jarrel-b (approved):
lgtm, just had a few questions!

service/media/media.go
bennycio: The reason why the delete only occurs in the default case, is because in every other case we will be storing something in GCP. IPFS/arweave, videos, etc. all get cached in GCP. If the media type or URI does not need caching, we need to delete what is in GCP for it so that when we are looking up URLs to fill out the media struct, we don't use an old image/video from a time when the metadata might have been a video.

service/media/media.go
bennycio: Yes, if this fails, what shows up might still be cached. It does fail every so often, need to look into why, unfortunately the imgix documentation and error returns are very weak.

PR #467, opened 2022-08-03
handle gifs separately, manually index with non-indexed tokens of owner

bennycio:
Changes:

- **Changed** gif handling to work similarly to images
- **Changed** manual index step to work for ERC-1155 tokens

### DIFF
--- a/graphql/resolver/schema.resolvers.go
+++ b/graphql/resolver/schema.resolvers.go
@@ -393,13 +393,13 @@ func (r *mutationResolver) RefreshCollection(ctx context.Context, collectionID p
 		return nil, err
 	}
 
-	token, err := resolveCollectionByCollectionID(ctx, collectionID)
+	collection, err := resolveCollectionByCollectionID(ctx, collectionID)
 	if err != nil {
 		return nil, err
 	}
 
 	output := &model.RefreshCollectionPayload{
-		Collection: token,
+		Collection: collection,
 	}
 
 	return output, nil
--- a/graphql/resolver/schema.resolvers.go
+++ b/graphql/resolver/schema.resolvers.go
@@ -385,6 +385,26 @@ func (r *mutationResolver) RefreshToken(ctx context.Context, tokenID persist.DBI
 	return output, nil
 }
 
+func (r *mutationResolver) RefreshCollection(ctx context.Context, collectionID persist.DBID) (model.RefreshCollectionPayloadOrError, error) {
+	api := publicapi.For(ctx)
+
+	err := api.Token.RefreshCollection(ctx, collectionID)
+	if err != nil {
+		return nil, err
+	}
+
+	token, err := resolveCollectionByCollectionID(ctx, collectionID)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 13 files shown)


Robinnnnn (approved):
🔥

service/media/media.go
```diff
@@ -191,6 +197,38 @@ func getAuxilaryMedia(pCtx context.Context, name, tokenBucket string, storageCli
...
+		MediaType: persist.MediaTypeGIF,
+	}
+	videoURL, err := getMediaServingURL(pCtx, tokenBucket, fmt.Sprintf("video-%s", name), storageClient)
+	if err == nil {
+		vURL = videoURL
+	}
+	imageURL, err := getMediaServingURL(pCtx, tokenBucket, fmt.Sprintf("image-%s", name), storageClient)
+	if err == nil {
+		logger.For(pCtx).Infof("found imageURL for %s: %s", name, imageURL)
+		res.MediaURL = persist.NullString(imageURL)
```
jarrel-b, line 211: Would we want to set the thumbnail URL too, or is it only useful for the case where we can't find the `image-%s` object?

jarrel-b (commented):
lgtm! There's gonna be a few conflicts with the my spam branch, but I'll merge this branch into mine since it's easier to integrate

graphql/resolver/schema.resolvers.go
```diff
@@ -385,6 +385,26 @@ func (r *mutationResolver) RefreshToken(ctx context.Context, tokenID persist.DBI
...
 
+func (r *mutationResolver) RefreshCollection(ctx context.Context, collectionID persist.DBID) (model.RefreshCollectionPayloadOrError, error) {
+	api := publicapi.For(ctx)
+
+	err := api.Token.RefreshCollection(ctx, collectionID)
+	if err != nil {
+		return nil, err
+	}
+
+	token, err := resolveCollectionByCollectionID(ctx, collectionID)
```
jarrel-b, line 396: maybe `collection` instead of `token`?

service/media/media.go
```diff
@@ -286,19 +280,18 @@ func getHTMLMedia(pCtx context.Context, name, tokenBucket string, storageClient
 		logger.For(pCtx).Infof("using imgURL for %s: %s", name, imgURL)
 		res.MediaURL = persist.NullString(imgURL)
 	}
-	imageURL, err := getMediaServingURL(pCtx, tokenBucket, fmt.Sprintf("image-%s", name), storageClient)
-	if err == nil {
+	if imageURL, err := getMediaServingURL(pCtx, tokenBucket, fmt.Sprintf("image-%s", name), storageClient); err == nil {
```
jarrel-b, line 283: Wonder if this could be helpful if pulled out into its own getThumbnailURL function?

service/media/media.go
bennycio: For things like GIFs and Images, we don't need to set the thumbnail because the image or GIF will be used as the thumbnail with imgix.

bennycio: With things like GIFs or Images we don't need to have a thumbnail because imgix will use the media URL to create the thumbnail.

PR #468, opened 2022-08-04
Add spam filter

jarrel-b:
**Changes**
* Adds a spam check in the indexer to determine if a token is spam or not
* Adds the option for a user to toggle if a token is spam or not
* [misc] - moves the "tokens owned by owner" query to the tokens repo and handles the check in `publicapi` instead
* [misc] - removes a few unused methods in the tokens and tokens gallery repo
* [misc] - fixes small bug where erc-1155 tokens with no balance wasn't getting deleted from the tokens table

**Deployment**
* [ ] Add `is_spam` column to indexer tokens table before deploying
* [ ] Run backend db migrations
* [ ] Would check Alchemy after deploying the indexer for 24hrs to make sure the spam check isn't adding too much overhead, since it requires one extra call per mint
    * IF it is too expensive, the quickest way to turn it off is to update `SpamCheck` of `indexer.go` to `return nil, nil` OR to delete the spam routine in `findFields` of `indexer.go` (also remember to decrement the wait group)
    * You could also unwind all the indexer changes, but that would also involve updating some SQL statements in `token.go`


indexer/indexer.go
```diff
@@ -1254,3 +1346,31 @@ func logEthCallRPCError(entry *logrus.Entry, err error, message string) {
...
+func SpamCheck(ctx context.Context, ethClient *ethclient.Client, transfer rpc.Transfer) (*bool, error) {
+	notSpam := util.BoolToPointer(false)
+	isSpam := util.BoolToPointer(true)
+
+	// Only transfers that are a result of minting is considered as potential spam.
+	if common.HexToAddress(string(transfer.From)) != common.HexToAddress("") {
+		return notSpam, nil
+	}
+
+	// Get the sender of the transaction. This requires only one RPC call because the sender gets cached on the first call.
```
bennycio, line 1360: So cool!

bennycio (approved):
Wonderful! Thanks for all of the cleanup as well with the unused persist functions!

PR #471, opened 2022-08-10
GAL-370 Add indexer plugins and full refresh endpoint

jarrel-b:
https://github.com/gallery-so/go-gallery/pulls should be reviewed+merged before reviewing!
### New deepRefresh graphql mutation
There's now a `deepRefresh` mutation that enqueues a refresh by the indexer. The indexer server's `/nfts/refresh` endpoint now accepts a `deepRefresh` flag to indicate that the update is a deep refresh.

### Indexer plugins
The Indexer was refactored a bit to support plugins which is a pattern that already existed in the indexer, but plugins make it a bit more concrete and should make it easier to add new features for a token. The plugins we have atm are:
* uris - gets the URI for a token
* balances - gets balance information for ERC1155s
* owners - gets ownership information for a token (prevOwners and owners was consolidated into a single plugin)
* refreshes - populates and saves the bloom filters for refreshes

### Redis synchronization primitives
Two new types were added to the `redis` package to help with buffering requests:
* `FifoQueue` - this implements a de-duped, reliable FIFO queue in redis and is used by the refresh system to enqueue requests. If there's a refresh already running with the same inputs, then the request won't be enqueued. If the server is restarted during a refresh, the message is kept in a "processing" queue that is used to re-process messages
* `Semaphore` - this implements a counting semaphore in redis. This is used to cap the number of refreshes that run at a time (capped somewhat arbitrarily at 24 atm) so that we don't hammer Alchemy and prevent the production Indexer from falling behind

### BlockFilterManager
This is a helper that pre-downloads chunks of filters at a time via a dataloader and saves the filters on disk to be read later. This helps download the filters more quickly, since we can assume if your downloading block 1001, you'll probably also need blocks 1002 to 10000 later on.

### Retrying RPC calls
All indexer calls to Alchemy use their `Retry*` variant to handle rate limits and to ensure that the production Indexer can still do its job

## Follow-ups 
**Considerations**
* The indexer server supports refreshes by: (`address`), (`address`, `contract`), (`address`, `contract`, `token`), and (`contract`, `token`), but the backend currently only sends requests by `address`. I think the other combinations of refreshes are more operational and could live in the admin UI in the future
* A full refresh (last 5 million blocks) takes about 8-15 minutes on my laptop depending on the user
* Depending on the user, a refresh takes about 100k alchemy compute units, our monthly limit is 14 billion units which we're currently well under so we should be good there - back of the envelope math is well under < $1 per refresh

**Post merge**
* [ ] Run backfill script to populate filters


indexer/indexer.go
```diff
@@ -265,7 +262,7 @@ func (i *indexer) Start(rootCtx context.Context) {
 	wp.StopWait()
 	logger.For(rootCtx).Info("Finished processing old logs, subscribing to new logs...")
 	i.lastSavedLog = i.lastSyncedBlock
-	if !i.isRPCEnabled {
+	if rpcEnabled {
```
benny-conn: needs to be `!rpcEnabled`?

indexer/nfts.go
```diff
@@ -553,6 +565,203 @@ func updateTokens(tokenRepository persist.TokenRepository, ethClient *ethclient.
...
+			}
+		}
+
+		acquired, err := refreshLock.Acquire(ctx)
+		if err != nil {
+			logger.For(ctx).WithError(err).Errorf("error occurred acquiring lock")
+			if err := refreshQueue.ReAdd(ctx, message); err != nil {
+				panic(err)
+			}
+			panic(err)
```
benny-conn, line 604: What is the purpose of adding back to the queue if this will panic regardless?

service/memstore/redis/redis.go
```diff
@@ -111,3 +112,212 @@ func (c *Cache) Close(clear bool) error {
...
+
+	removed := pipe.ZRem(ctx, s.name, s.id)
+	pipe.ZRem(ctx, s.owners, s.id)
+
+	_, err := pipe.Exec(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	return removed.Val() > 0, nil
```
benny-conn, line 283: 😵‍💫 going to trust that all of this works because I have no idea what is going on lol

benny-conn (approved):
Love it! Some really complex stuff with the filters, queues, semaphore (didn't know what that word meant) and the indexer feels a lot cleaner!

indexer/indexer.go
jarrel-b: good catch! just realized that when the test didn't finish

indexer/nfts.go
jarrel-b: ah we want to add the message back to the redis queue another worker can pick it up later

service/memstore/redis/redis.go
jarrel-b: haha this is all really so we don't run a million refreshes simultaneously

PR #475, opened 2022-08-17
badges for communities

bennycio:
Changes:

- **Added** fields to the contracts type to support user inputted and manually inputted data for profile elements and badges
- **Added** dataloaders and queries for getting all contracts that are from tokens owned by a user
- **Added** fields in the graphQL schema for querying all of a user's contracts as well as the new community and contract fields

Considerations:

- As I made this PR, I noticed some repetition now between the `Community` and `Contract` types. The community type seems to be a super-type of the contract, but if we are planning on using contracts in a "gallery way" like we do tokens (not the blockchain way), is it worth consolidating these types? 
- The major difference right now between communities and contracts is that communities are tied to users that display NFTs from a given contract where the current `getContractsForUser` is all of the contracts they own tokens from regardless of display. This leads me to ask: Should the query for contracts of a user only correspond to contracts of NFTs they have displayed or ones they own? 
- Also, right now the get community query does some expensive querying for getting all of the holders in a community with preview images. If we are going to consolidate types, we could turn this into a separate resolver. We would also need to consider if the current `description` field on a community which is derived from the `description` field on one of the NFTs in the community should be a new field on the `contracts` table as well.


radazen (approved):
Left some comments (mostly copy/paste edits), but this looks A-OK to me!

Re: consolidating communities and contracts, these are all good questions! Probably a good topic for discussion in tomorrow's standup. I'm going to keep thinking about it in the meantime!

radazen:
Also, thinking out loud: maybe we actually want a `Badges` resolver as a child of the GraphQL `User` type, instead of (or in addition to) the `Contracts` resolver?

Contracts would be "all of the contracts this user holds tokens for," but Badges would be "all of the badges this user has." At present, I think the sole distinction between SQL queries would be that `user.Badges` should only return contracts that have those badge profile URL fields filled out, instead of returning all contracts. In the future, `Badges` would also return manually-added badges, but the frontend wouldn't need to care about that detail, because the backend would just include them as part of the `Badges` resolver.

And maybe a GraphQL `Badge` type? Given that some future badges may not be contracts, maybe a `Badge` object is just the set of badge-specific URLs, and an optional `contract_id` pointing to the contract if there is one? Maybe a name, too, if badges have names?

PR #476, opened 2022-08-18
traits column

bennycio:
Changes:

- **Added** `traits` field on users
- **Added** query and resolver for getting all users with a given trait

Considerations:

- The `traits` field is a JSON object where the key is the name of the trait and the value could be anything. In the case of GOTW, the value might be a timestamp representing when they were GOTW. I added a query for getting users with a given trait but in many cases it seems like we might want a very specific query for something like "get all users who were GOTW and sort them by how recent they were GOTW" so that we can see who is currently GOTW as well as an ordered list of previous recipients. Do we need to come up for a system on the backend for things like this or is it alright if the front-end sorts when it receives this information?
- Maybe we should see how this performs with out indexes on the new JSONB column and then potentially add them if necessary?


Robinnnnn (approved):
> "get all users who were GOTW and sort them by how recent they were GOTW"

this is definitely something we'll want but i don't think it's something we need to optimize for immediately!

awesome work on this, benny!

kaitoo1 (approved):

PR #477, opened 2022-08-23
poap support

bennycio:
Changes:

- **Added** multichain provider for POAP tokens

Considerations:

- Need an API key to test whether or not they return token ID's in base 10 or base 16 like we want them
- Only one contract exists for POAP. It is an Ethereum contract. I hard-coded it to be returned with the tokens for consistency across other chains. Is this alright? We could also just return no contract because technically they do not exist on a contract until they are minted.


Robinnnnn:
> I hard-coded it to be returned with the tokens for consistency across other chains. Is this alright?

sounds fine with me! do you mind just making sure to add a comment in the code about it?

Robinnnnn:
also, incredibly excited about how "easy" it was to integrate a new chain for our new provider 😍

Robinnnnn (approved):

Robinnnnn (approved):

service/multichain/multichain.go
```diff
@@ -19,6 +19,8 @@ type Provider struct {
 	ContractRepo persist.ContractGalleryRepository
 	UserRepo     persist.UserRepository
 	Chains       map[persist.Chain][]ChainProvider
+	// some chains use the addresses of other chains, this will map of chain we want tokens from => chain that's address will be used for lookup
+	ChainAddressOverrides map[persist.Chain]persist.Chain
```
jarrel-b, line 23: Are POAPs only transferrable to eth, could they get transferred to polygon or somewhere else? I was wondering too if whenever a user adds an Ethereum wallet we add an extra entry but with POAP as the chain. Would that be weird? We wouldn't need the override concept, but not sure if that has side effects elsewhere.

bennycio: Yeah, they are currently only transferrable to ETH, but the primary purpose of this is that even when they are not on any chain, they use the ETH address as their "address". If POAP decided to let people sign up with Tezos addresses as well, we could change the value of this map to an array of chains instead of just one chain?

jarrel-b: makes sense!

jarrel-b (approved):

PR #478, opened 2022-08-25
syncing media

bennycio:
Changes:

- **Added** `SyncingMedia` to the graphQL schema
- **Added** MediaTypeSyncing that will be returned when an NFT is still syncing
- **Added** resolvers for this media type
- **Updated** default case for graphQL media resolving to invalid instead of unknown. Unknown is a case we want to handle and expect to occur, invalid is a case where something is going wrong.

### DIFF
--- a/graphql/generated/generated.go
+++ b/graphql/generated/generated.go
@@ -451,6 +451,13 @@ type ComplexityRoot struct {
 		Viewer func(childComplexity int) int
 	}
 
+	SyncingMedia struct {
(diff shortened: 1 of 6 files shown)


tbezman (approved):

Robinnnnn (approved):

radazen (approved):
As per 2 minutes ago, it'd be cool to give `MediaTypeSyncing` a non-empty string type, but otherwise everything LGTM! 🔥

PR #479, opened 2022-08-25
Benny/tez address

bennycio:
Changes:

- **Added** new `ChainPubKeyInput` and `ChainPubKey` types to the graphQL schema
- **Added** respective go types
- **Changed** verify signature to take in a pub key

### DIFF
--- a/service/user/user.go
+++ b/service/user/user.go
@@ -16,7 +16,6 @@ import (
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/mikeydub/go-gallery/service/auth"
 	"github.com/mikeydub/go-gallery/service/eth"
-	"github.com/mikeydub/go-gallery/service/multichain"
 	"github.com/mikeydub/go-gallery/service/persist"
 	"github.com/mikeydub/go-gallery/util"
 	"github.com/mikeydub/go-gallery/validate"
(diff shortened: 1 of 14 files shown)


service/user/user.go
```diff
@@ -329,35 +328,36 @@ func UpdateUser(pCtx context.Context, userID persist.DBID, username string, bio
...
+// 	if pInput.WalletType != persist.WalletTypeEOA {
+// 		if auth.NewNoncePrepend+nonce != pInput.Nonce && auth.NoncePrepend+nonce != pInput.Nonce {
+// 			return auth.ErrNonceMismatch
+// 		}
+// 	}
+
+// 	sigValidBool, err := multichainProvider.VerifySignature(pCtx, pInput.Signature, nonce, chainAddress, pInput.WalletType)
+// 	if err != nil {
+// 		return err
+// 	}
```
bennycio, line 352: commented out because of this part, and it is not in use, we can probably just delete this

Robinnnnn: WEN ADMIN MERGE @radazen

Robinnnnn: in all seriousness, @tbezman made a good point that tezos users may accidentally create a new account that they want to combine with their ETH one. something to think about hehe

still got a few weeks until launch!

Robinnnnn (approved):
muy fast

Robinnnnn:
actually @bennycio i think this needs to be updated as well: 
image

PR #482, opened 2022-08-29
[GAL-355] Add tokensByChain field to GraphQL GalleryUser type

radazen:
This adds a new `tokensByChain` query to the GraphQL `GalleryUser` type, which can be used to query a user's tokens on a per-chain basis. While I had originally planned to accept an array of chain inputs (e.g. `tokensByChain([Chains!]!)`, I think that starts to get ugly with pagination and cursors (not currently implemented, but likely will be at some point!).

Instead, the query accepts a single chain, but if we need to fetch tokens for multiple chains in the same GraphQL query, we can still do that:

```
...on GalleryUser {
  ethTokens: tokensByChain(chain: Ethereum) { ... }
  tezosTokens: tokensByChain(chain: Tezos) { ... }
}
```

I also added a validator for chain types.


db/queries/query.sql
```diff
@@ -159,6 +159,22 @@ SELECT tokens.* FROM tokens, users
       AND tokens.deleted = false AND users.deleted = false
     ORDER BY tokens.created_at DESC, tokens.name DESC, tokens.id DESC;
 
+-- name: GetTokensByUserIdAndChain :many
```
jarrel-b, line 162: Is this needed if there's a batched version?

jarrel-b (approved):

tbezman (approved):
+1 from the schema side of things!

db/queries/query.sql
radazen: Good call. I've been adding batched and non-batched versions of everything out of habit, but there's really no reason to have the non-batched one since it's not being used anywhere.

radazen:
> +1 from the schema side of things!
> 
> 

Doh. Sorry I didn't request a review from you in the first place!

PR #485, opened 2022-08-31
admire, comment

bennycio:
Changes:

- **Added** two new tables for storing admires and comments
- **Added** graphQL and sqlc types, types, dataloaders, queries, and mutations
- **Added** minor indexer fix for memory management

Considerations:

- Left one comment in the code on whether or not we want to include the feed event in the graphQL types for admire and feed event so we can go from admire to feed event and back recursively if we want?
- Are any of these blockers for releasing this feature: Moderation by the `feed_event`'s `actor_id` (being able to remove unwanted comments), blocking other users and not being able to see their interactions, spam protection to ensure someone doesn't comment a bajillion times? (Anything else that I could be missing???)
- Assuming we want comments and admires also to be feed events, that could probably be a quick follow up to this PR.
- We probably want to restrict what kind of feed events can be admired and commented on. First part of this is figuring out which ones we want, options below (thinking all of them except following users and user created), and how best to enforce them. We can enforce them easily through validation at the publicAPI level, but I know Ezra has been working on some fun database level requirement enforcement and was wondering if maybe Ezra had any insights to enforcing that the `FEED_EVENT_ID` references an event whose `ACTION` field is one of the valid actions.
```
        ActionUserCreated                     Action = "UserCreated"
	ActionUserFollowedUsers               Action = "UserFollowedUsers"
	ActionCollectorsNoteAddedToToken      Action = "CollectorsNoteAddedToToken"
	ActionCollectionCreated               Action = "CollectionCreated"
	ActionCollectorsNoteAddedToCollection Action = "CollectorsNoteAddedToCollection"
	ActionTokensAddedToCollection         Action = "TokensAddedToCollection"
```


Robinnnnn:
answering some of these:

> Left one comment in the code on whether or not we want to include the feed event in the graphQL types for admire and feed event so we can go from admire to feed event and back recursively if we want?

not necessary out of the gate. for Notifications, however, we'll want to go from a notification to the original Feed Event

> Are any of these blockers for releasing this feature: Moderation by the feed_event's actor_id (being able to remove unwanted comments), blocking other users and not being able to see their interactions, spam protection to ensure someone doesn't comment a bajillion times? (Anything else that I could be missing???)

not blockers for feature release. these are really great questions though, cc @kaitoo1. we'll probably need to think about what the UI looks like to actually block people, etc.

> Assuming we want comments and admires also to be feed events, that could probably be a quick follow up to this PR.

this is super interesting. i don't think we have this mocked up, but it would indeed be cool to see "50 people liked DC investor's collection" pop up in the feed. i would consider this non-blocking for the initial release. that said, we could still send reactions as feed events, and the frontend could just ignore it (similar to how the frontend ignores "X user joined gallery", even though the feed event is available to parse). cc @kaitoo1 

> We probably want to restrict what kind of feed events can be admired and commented on. First part of this is figuring out which ones we want, options below (thinking all of them except following users and user created), and how best to enforce them. We can enforce them easily through validation at the publicAPI level, but I know Ezra has been working on some fun database level requirement enforcement and was wondering if maybe Ezra had any insights to enforcing that the FEED_EVENT_ID references an event whose ACTION field is one of the valid actions.

great callout. i do not know what the lift is to do API and/or DB-level validation, but i would also consider this non-blocking for the initial rollout. if a user can figure out how to use our API to directly "like" random feed events, so be it!

furthermore, from the events you've listed, i don't want to explicitly _block_ the ability to react to them in case at some point down the line we want to enable it. for example, right now we don't have UI support for letting users "like" the fact that Benny followed Ezra, but why not open that up in the future? if we consider any activity on gallery to be a "primitive", then you could argue that anyone should be able to react to it.

kaitoo1:
great questions benny!! and thanks for answering robin

re: moderation, yup these will be great follow up improvements: (delete, report, spam protection, block) that we can design for the next version. Do these mutations have rate limits? wouldn't hurt to have some kind of rate limit in place to put a cap to any spam.
so not blockers for release, but love that we're thinking about these. if we can just anticipate these features so that it doesn't require a big refactor to implement them, that'd be enough for this release!

re: comments + admires as feed events - agree with robin this is totally a possibility but let's keep out of scope for this version.

publicapi/admire.go
```diff
@@ -0,0 +1,74 @@
...
+		"feedEventID": {feedEventID, "required"},
+		"actorID":     {actorID, "required"},
+	}); err != nil {
+		return "", err
+	}
+
+	return api.repos.AdmireRepository.CreateAdmire(ctx, feedEventID, actorID)
+}
+
+func (api AdmireAPI) UnadmireFeedEvent(ctx context.Context, admireID persist.DBID) error {
```
jarrel-b, line 65: Would we want some authorization check so that a user can only un-admire their own admire, and not a another user's admire?

publicapi/comment.go
```diff
@@ -0,0 +1,74 @@
...
+		"feedEventID": {feedEventID, "required"},
+		"actorID":     {actorID, "required"},
+	}); err != nil {
+		return "", err
+	}
+
+	return api.repos.CommentRepository.CreateComment(ctx, feedEventID, actorID, replyToID, comment)
+}
+
+func (api CommentAPI) RemoveCommentOnFeedEvent(ctx context.Context, commentID persist.DBID) error {
```
jarrel-b, line 65: Maybe an authorization check so that a user can only delete their own comments?

graphql/resolver/schema.resolvers.go
```diff
@@ -578,6 +622,100 @@ func (r *mutationResolver) UnfollowUser(ctx context.Context, userID persist.DBID
...
+	output := &model.AdmireFeedEventPayload{
+		Viewer:    resolveViewer(ctx),
+		Admire:    admire,
+		FeedEvent: feedEvent,
+	}
+	return output, nil
+}
+
+func (r *mutationResolver) RemoveAdmire(ctx context.Context, admireID persist.DBID) (model.RemoveAdmirePayloadOrError, error) {
+	admire, err := publicapi.For(ctx).Admire.GetAdmireByID(ctx, admireID)
```
jarrel-b, line 649: You think we need to check if an admire exists vs. just deleting it straight away? Similar question for comments

graphql/resolver/schema.resolvers.go
```diff
@@ -578,6 +622,100 @@ func (r *mutationResolver) UnfollowUser(ctx context.Context, userID persist.DBID
...
+	if err != nil {
+		return nil, err
+	}
+
+	commentDB, err := resolveCommentByCommentID(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+
+	feedEvent, err := resolveFeedEventByEventID(ctx, feedEventID)
```
jarrel-b, line 680: I think we'll only need to fully resolve a comment and a feedEvent if the query specifies those fields. Maybe we can return partially filled structs for those types with the `ID` field added for their own resolvers to fill in the rest?

jarrel-b (commented):
looking good! just had a few comments, super excited to see this come out

graphql/resolver/schema.resolvers.go
```diff
@@ -578,6 +622,75 @@ func (r *mutationResolver) UnfollowUser(ctx context.Context, userID persist.DBID
...
+		Viewer: resolveViewer(ctx),
+		FeedEvent: &model.FeedEvent{
+			Dbid: feedEvent, // remaining fields handled by dedicated resolver
+		},
+	}
+	return output, nil
+}
+
+func (r *mutationResolver) CommentOnFeedEvent(ctx context.Context, feedEventID persist.DBID, replyToID *persist.DBID, comment string) (model.CommentOnFeedEventPayloadOrError, error) {
+	user := publicapi.For(ctx).User.GetLoggedInUserId(ctx)
```
radazen, line 656: Could we move the `GetLoggedInUserId` part of this to the public API layer? I usually try asking myself the following: "if we had a GraphQL API _and_ a REST API, would I want this functionality in both of them? If so, it belongs in the Public API layer and not in the GraphQL layer." It's not a likely scenario, but it _is_ a good mental model for what should live where.

In this particular case, I think we'd want to make sure that the user attempting to comment on a feed event is the logged-in user, regardless of what transport mechanism they used, so any authorization checks should live in the Public API layer. Same for Admire/RemoveAdmire/RemoveComment.

sqlc.yaml
```diff
@@ -77,6 +77,20 @@ overrides:
   - column: "events.feed_window_size"
     go_type: "int"
 
+  # Admires
+  - column: "admires.actor_id"
```
radazen, line 81: No action items here, but as an FYI, the PR I've been working on includes a wildcard for `*.*_id` to be mapped to persist.DBID. Hopefully that'll make these mappings a little less involved in the future!

radazen (approved):
Left some comments/requests, but this looks great overall!

PR #487, opened 2022-08-31
Turn off spam

jarrel-b:
Turns out the spam check adds a lot of latency to indexing and is a bottleneck in processing transfers (more than the actual downloading of the logs themselves), example trace from Sentry:

https://sentry.io/organizations/usegallery/performance/indexer:219bf50da08a41da8768fc97404bbf8c/?environment=local&project=6592179&query=&showTransactions=recent&statsPeriod=24h&transaction=indexer-main%3AstartPipeline&unselectedSeries=p100%28%29

I don't think the spam check should prevent new blocks from being indexed, so I'm removing it from the main path of the indexer. It probably makes sense to check for spam when a user syncs their tokens - around the time we fetch the media content and cache it. We can also prioritize refreshing media for tokens that aren't spam, then refresh tokens that are marked as spam.

I'll work on that as a fast follow up to the deep refresh stuff!
https://linear.app/galleryso/issue/GAL-374/move-spam-check-to-sync-tokens-path

### DIFF
--- a/indexer/indexer.go
+++ b/indexer/indexer.go
@@ -115,11 +115,6 @@ type tokenMedia struct {
 	media persist.Media
 }
 
-type spamEvaluation struct {
-	ti     persist.EthereumTokenIdentifiers
-	isSpam *bool
-}
-
 // indexer is the indexer for the blockchain that uses JSON RPC to scan through logs and process them
 // into a format used by the application
 type indexer struct {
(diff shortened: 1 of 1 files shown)


Robinnnnn (approved):

radazen (approved):
LGTM!

radazen:
Also, I love seeing Sentry traces being useful.

PR #489, opened 2022-08-31
poap events and begin multichain community

bennycio:
Changes:

- **Changed** POAP multichain provider to use events as their contracts
- **Added** a `GetCommunityOwners` function to the chain providers and implemented it for POAP
- **Added** a `GetCommunityOwners` function to the multichain provider that will be used to resolve owners for a given community
- **Changed** graphQL resolvers for communities to use the `contracts` repository and multichain providers instead of community repository
- **Removed** `CommunityRepository` (although it will technically be back soon when we rename the `ContractRepository`)


bennycio:
Still need to add:

- [ ] implement `GetCommunityOwners` for ETH and Tezos

Still need to test: 

- [ ] new community owners resolving and ensure preview images are there?
- [ ] do POAPs show up as new contracts/communities/collections/whatever-the-hecks now?

service/multichain/multichain.go
```diff
@@ -489,6 +594,54 @@ func contractsToNewDedupedContracts(ctx context.Context, contracts []chainContra
...
+	}
+
+	res := make([]TokenHolder, 0, len(seen))
+	for _, t := range seen {
+		res = append(res, t)
+	}
+	return res
+}
+
+func tokenHoldersToTokenHolders(ctx context.Context, owners []persist.TokenHolder, userRepo persist.UserRepository) ([]TokenHolder, error) {
```
Robinnnnn, line 614: love this name 😂

Robinnnnn (approved):
looks incredible benny, thanks for cranking this out so quickly! LGTM but would probably be best if there were at least 1 other reviewer 🙏

could we also deploy this to Playground for testing?

service/multichain/multichain.go
```diff
@@ -74,6 +78,17 @@ type ChainAgnosticIdentifiers struct {
 	TokenID         persist.TokenID `json:"token_id"`
 }
 
+type ChainAgnosticCommunityOwner struct {
+	Address persist.Address `json:"address"`
+}
+
+type TokenHolder struct {
```
radazen, line 85: Is `TokenHolder` a Gallery user, or could it be a non-Gallery member of the community? If it's the latter, what would we expect these fields to look like for a non-Gallery TokenHolder?

service/multichain/multichain.go
```diff
@@ -489,6 +594,54 @@ func contractsToNewDedupedContracts(ctx context.Context, contracts []chainContra
...
+	return res
+}
+
+func tokenHoldersToTokenHolders(ctx context.Context, owners []persist.TokenHolder, userRepo persist.UserRepository) ([]TokenHolder, error) {
+	seen := make(map[persist.DBID]TokenHolder)
+	for _, owner := range owners {
+
+		if _, ok := seen[owner.UserID]; !ok {
+			var username string
+			user, err := userRepo.GetByID(ctx, owner.UserID)
```
radazen, line 620: I think this could benefit significantly from a `userRepo.GetByIDs` call that returns all users in a list of input IDs and doesn't require a round trip per lookup. Batching could also work, but I think it's probably more efficient to do it as a single query and cache the results in a map.

service/multichain/poap/poap.go
```diff
@@ -187,13 +334,14 @@ func (d *Provider) poapToToken(pPoap poapToken) multichain.ChainAgnosticToken {
 		Description:     pPoap.Event.Description,
 		Quantity:        "1",
 		ExternalURL:     pPoap.Event.EventURL,
-		ContractAddress: poapAddress,
+		ContractAddress: persist.Address(pPoap.Event.FancyID),
```
radazen, line 337: This made me laugh, but I fully support it!

service/persist/contract_gallery.go
```diff
@@ -16,16 +17,28 @@ type ContractGallery struct {
...
 	Name             NullString `json:"name"`
+	Description      NullString `json:"description"`
 	CreatorAddress   Address    `json:"creator_address"`
 	ProfileImageURL  NullString `json:"profile_image_url"`
 	ProfileBannerURL NullString `json:"profile_banner_url"`
 	BadgeURL         NullString `json:"badge_url"`
 }
 
+// ErrContractNotFoundByAddress is an error type for when a contract is not found by address
+type ErrGalleryContractNotFoundByAddress struct {
```
radazen, line 28: Similar to my comments on the other PR, how about `ErrGalleryContractNotFound`?

graphql/schema/schema.graphql
```diff
@@ -353,22 +354,25 @@ type MembershipTier implements Node {
     owners: [TokenHolder]
 }
 
-type Community implements Node @goGqlId(fields: ["contractAddress", "chain"]) {
-    id: ID!
+type Community implements Node
```
radazen, line 357: Since a `Community` has a DBID now (because it's a contract?), I think we should be able to use that as the default gqlID instead of using the @goGqlId directive. But we might want to hold off on changing that immediately...going to think a bit on the implications.

radazen (approved):
Left some minor comments, but looks good to me! POOOOOAAAAAAPPPPPSSSSSSSS!!!

service/multichain/multichain.go
bennycio: It could be a GalleryUser if the UserID and WalletIDs are filled out, we just want to fill out as much information as we can for each holder, in the cases they aren't a user, we just fill out the display name and try to find some previews.

service/multichain/multichain.go
bennycio: True! Will put this together real quick before merging.

graphql/schema/schema.graphql
bennycio: True, will hold off for now until we complete the change from contract to community but that makes sense to me!

PR #496, opened 2022-09-13
[GAL-422] Check if token's media is servable

jarrel-b:
### Changes
Updates the dedupe logic when combining separate providers:
* Add `IsServable` method to the `Media` type. It seems that we can generalize on the fact that a token's media is servable if it has a media type and it has a media URL. Only types of `unknown` and `""` are missing a `media_url`
```
postgres=# select media->>'media_type' media_type, media->>'media_url' != '' has_url, count(*) count from tokens group by 1, 2;
 media_type | has_url | count
------------+---------+--------
 audio      | t       |   2953
 json       | t       |      1
 svg        | t       |  16431
 animation  | t       |    901
 unknown    | t       |  31228
 html       | t       |  35120
 video      | t       |  55509
 text       | t       |    785
 unknown    |         |   5673
 gif        | t       |  34164
            |         |  58903
 image      | t       | 308467
(12 rows)
```
* Use `IsServable` to determine which provider to use for the token

### How did I test?
Pointed my local backend to the prod indexer API and ran "Refresh wallet" on `noise_dao`.

Prior to the change, noise dao's NFTs looked like:
```
postgres=# select media->>'media_type' media_type, deleted, count(*) count from tokens where owner_user_id = '2CmnkzXTHkxlJTfhOVUIkjJqHI1' group by 1,2 order by 2,1;
 media_type | deleted | count
------------+---------+-------
            | f       |   415
 audio      | f       |   325
 gif        | f       |     1
 html       | f       |    82
 image      | f       |    81
 svg        | f       |     5
 unknown    | f       |     3
 video      | f       |   144
 audio      | t       |     1
 image      | t       |    38
 svg        | t       |     1
 video      | t       |     2
(12 rows)
```

After running a sync, their NFTs looks like:
```
 media_type | deleted | count
------------+---------+-------
            | f       |     7
 audio      | f       |   430
 gif        | f       |     1
 html       | f       |   204
 image      | f       |   120
 svg        | f       |     6
 unknown    | f       |    21
 video      | f       |   301
 audio      | t       |     1
 image      | t       |    38
 svg        | t       |     1
 video      | t       |     2
(12 rows)
```

For `dcinvestor`, before:
```
postgres=# select media->>'media_type' media_type, deleted, count(*) count from tokens where owner_user_id = '3369d75dc3f482f8cf7213b03670112f' group by 1,2 order by 2,1;
 media_type | deleted | count
------------+---------+-------
 gif        | f       |    65
 html       | f       |   104
 image      | f       |   408
 svg        | f       |    12
 unknown    | f       |    23
 video      | f       |   116
 image      | t       |    28
(7 rows)
```
and after:
```
postgres=# select media->>'media_type' media_type, deleted, count(*) count from tokens where owner_user_id = '3369d75dc3f482f8cf7213b03670112f' group by 1,2 order by 2,1;
 media_type | deleted | count
------------+---------+-------
            | f       |   374
 animation  | f       |     3
 gif        | f       |    84
 html       | f       |    99
 image      | f       |   468
 svg        | f       |    50
 unknown    | f       |    42
 video      | f       |   175
 image      | t       |    30
(9 rows)
```

It looks like dcinvestor hasn't ran a sync since early July, but the extra `unknown`s look like a bunch of spam nfts. Inspecting his Gallery after the refresh though looks identical to before the refresh. Also running a refresh on my Gallery fixed tokens that were previously broken.

### DIFF
--- a/service/multichain/multichain.go
+++ b/service/multichain/multichain.go
@@ -496,33 +496,34 @@ func tokensToNewDedupedTokens(ctx context.Context, tokens []chainTokens, contrac
 
 			ti := persist.NewTokenIdentifiers(token.ContractAddress, token.TokenID, chainToken.chain)
 
-			if it, ok := seenTokens[ti]; ok {
-				if !(it.Media.MediaType == persist.MediaTypeVideo && it.Media.ThumbnailURL == "") {
-					if it.Media.MediaURL != "" && it.Name != "" {
-						continue
-					}
-				} else {
-					if it.Media.MediaURL != "" {
-						token.Media.MediaURL = it.Media.MediaURL
-					}
-				}
-				logger.For(ctx).Debugf("updating token %s because current version is invalid", ti)
+			if seen, ok := seenTokens[ti]; ok && !seen.Media.IsServable() && token.Media.IsServable() {
+				seen.Media = token.Media
+				seenTokens[ti] = seen
 			} else {
 				if w, ok := addressToWallets[chainToken.chain.NormalizeAddress(token.OwnerAddress)]; ok {
 					seenWallets[ti] = append(seenWallets[ti], w)
 				}
 				if q, ok := seenQuantities[ti]; ok {
-					if token.Name == "aspiring chad" {
-						asJSON, _ := json.MarshalIndent(token, "", "  ")
-						logger.For(ctx).Warnf("t quantity again %s | %s", token.Quantity, string(asJSON))
-					}
 					seenQuantities[ti] = q.Add(token.Quantity)
 				} else {
-					if token.Name == "aspiring chad" {
-						asJSON, _ := json.MarshalIndent(token, "", "  ")
-						logger.For(ctx).Warnf("t quantity %s | %s", token.Quantity, string(asJSON))
-					}
 					seenQuantities[ti] = token.Quantity
+				}
+				seenTokens[ti] = persist.TokenGallery{
+					Media:                token.Media,
+					TokenType:            token.TokenType,
+					Chain:                chainToken.chain,
+					Name:                 persist.NullString(token.Name),
+					Description:          persist.NullString(token.Description),
+					TokenURI:             token.TokenURI,
+					TokenID:              token.TokenID,
+					Quantity:             seenQuantities[ti],
+					OwnerUserID:          ownerUser.ID,
+					OwnedByWallets:       seenWallets[ti],
+					TokenMetadata:        token.TokenMetadata,
+					Contract:             contractAddressIDs[chainToken.chain.NormalizeAddress(token.ContractAddress)],
+					ExternalURL:          persist.NullString(token.ExternalURL),
+					BlockNumber:          token.BlockNumber,
+					IsProviderMarkedSpam: token.IsSpam,
 				}
 			}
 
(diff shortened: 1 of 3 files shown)


service/multichain/multichain.go
```diff
@@ -496,60 +496,45 @@ func tokensToNewDedupedTokens(ctx context.Context, tokens []chainTokens, contrac
...
-						continue
-					}
-				} else {
-					if it.Media.MediaURL != "" {
-						token.Media.MediaURL = it.Media.MediaURL
-					}
-				}
-				logger.For(ctx).Debugf("updating token %s because current version is invalid", ti)
+			if seen, ok := seenTokens[ti]; ok && !seen.Media.IsServable() && token.Media.IsServable() {
+				seen.Media = token.Media
```
jarrel-b, line 500: Would we want this to be all-or-nothing, as in use everything field coming from OS or use everything from the indexer, rather than a mix of both?

Robinnnnn: yes i think all-or-nothing, bc if some of the fields are broken from the indexer then we should assume the worst about every other field

Robinnnnn (commented):
thanks for doing this, jarrel!

if we realize the media is broken for a token, would we at least trigger an async refresh for the token on the indexer side?

service/multichain/multichain.go
radazen: I'd be curious to hear @bennycio's opinion on this! Since we're only using Opensea as a fallback for media-related data, I feel like it should be all-or-nothing, but restricted to "all media information." That being said, what really constitutes media information? Just the `media` field? Or should `token_uri` and `token_metadata` be filled in from Opensea, too?

service/multichain/multichain.go
```diff
@@ -496,60 +496,45 @@ func tokensToNewDedupedTokens(ctx context.Context, tokens []chainTokens, contrac
 
 			ti := persist.NewTokenIdentifiers(token.ContractAddress, token.TokenID, chainToken.chain)
 
-			if it, ok := seenTokens[ti]; ok {
-				if !(it.Media.MediaType == persist.MediaTypeVideo && it.Media.ThumbnailURL == "") {
```
radazen, line 500: Do you know what we'd be losing by removing this `persist.MediaTypeVideo` check? It's not clear to me what it's actually doing, which also makes me a little nervous about removing it.

jarrel-b: thats a good point, I kind of yolo'd this but I'll see what of our tokens meet this condition. Thinking in general that checking for a specific media type isn't needed because it seems to me if the media URL is empty we'd always want to replace it. Similar reasoning for `Name`, doesn't seem relevant unless its an empty name could be an indicator that something went wrong while indexing? @bennycio does that sound reasonable?

service/multichain/multichain.go
jarrel-b: closing the loop: want to only replace the `Media` field, only want to serve assets from OS when we need to, we shouldn't get too dependant on OS

service/multichain/multichain.go
jarrel-b: closing the loop: this check was originally for live render; videos without thumbnails would not have a proper preview image to use. It sounds like the frontend a) already handles this case because it had to do so with OS, or b) there will be no preview and a user would need to refresh their nft. Will confirm what happens now though

jarrel-b: Looks like the frontend doesn't take a static image of the video, so users would need to refresh their token if they aren't. This only affects about 50 tokens being displayed atm though:
```
select t.*
from collections c
, (
	select *
	from tokens 
	where media->>'media_type' = 'video' and (media->>'thumbnail_url' = '' or media->>'thumbnail_url' is null)
) t
where (c.token_settings ? t.id is null or c.token_settings ? t.id is false) and t.id = any(c.nfts)
```

jarrel-b:
> thanks for doing this, jarrel!
> 
> if we realize the media is broken for a token, would we at least trigger an async refresh for the token on the indexer side?

Indexer already handles refreshes for tokens with broken media on syncs, so no need to do a re-refresh

service/multichain/multichain.go
Robinnnnn: awesome, thanks for doing the research @jarrel-b!

Robinnnnn (approved):

radazen (approved):
🚀

PR #498, opened 2022-09-13
[GAL-427] svgs, base64, and thumbnails

bennycio:
Changes:

- **Changed** regression for svg's and base64 images. Caused by url unescaping every URI in the `GetDataFromURI` function. The `(TokenURI).String()` will already query unescape the tokenURIs that need it (the ones that are actually URLs with queries).
- **Changed** `getThumbnailURL` function in final step of media creation to only use the raw `imgURL` if it `IsRenderable` which is a new function on the tokenURI to check if the frontend can use it. This will prevent IPFS urls and the like from showing up in the thumbnail field.
- **Changed** internal folder to have env files for indexer server as well as some improvements to the backend env file


Robinnnnn (approved):

indexer/core.go
```diff
@@ -143,7 +143,11 @@ func setDefaults() {
 	viper.AutomaticEnv()
 
 	if viper.GetString("ENV") == "local" {
-		path, err := util.FindFile("_internal/app-local-indexer.yaml", 3)
+		filePath := "_internal/app-local-indexer.yaml"
+		if server {
+			filePath = "_internal/app-local-indexer-server.yaml"
```
Robinnnnn, line 148: i vote we rename `_internal` to `_local`

indexer/metadatas.go
```diff
@@ -418,3 +421,44 @@ func cryptopunks(ctx context.Context, turi persist.TokenURI, addr persist.Ethere
...
+	case persist.MediaTypeImage, persist.MediaTypeGIF, persist.MediaTypeSVG:
+		resultMetadata["image"] = mediaURI
+	default:
+		resultMetadata["animation_url"] = mediaURI
+		someOtherURI, ok := util.FindFirstFieldFromMap(tokenMetadata, "image", "thumbnail", "uri").(string)
+		if ok {
+			resultMetadata["image"] = someOtherURI
+		}
+	}
+	return persist.TokenURI(metadataURI), resultMetadata, nil
```
jarrel-b, line 463: you think it'd be worth it to just throw in all the other fields in there too just so we have it?

service/persist/token.go
```diff
@@ -570,6 +570,13 @@ func (uri TokenURI) Type() URIType {
 	}
 }
 
+// IsRenderable returns whether a frontend could render the given URI directly
```
jarrel-b, line 573: Is this because our frontend doesn't handle non-http uris yet, or browsers in general, don't really support things like ipfs yet?

would it make sense if this went something like:
```
func (uri TokenURI) IsHTTP() bool {}

// maybe in the future we have this method
func (uri TokenURI) IsIPFS() bool{}

func (uri TokenURI) IsRenderable() bool {
  return uri.IsHTTP()

  // maybe in the future 
  return uri.IsHTTP() || uri.IsIPFS()
}
```

indexer/metadatas.go
```diff
@@ -418,3 +421,44 @@ func cryptopunks(ctx context.Context, turi persist.TokenURI, addr persist.Ethere
 	}, nil
 
 }
+
+func zora(ctx context.Context, turi persist.TokenURI, addr persist.EthereumAddress, tid persist.TokenID, ethCl *ethclient.Client, ipfs *shell.Shell, arweave *goar.Client) (persist.TokenURI, persist.TokenMetadata, error) {
```
jarrel-b, line 425: I had this idea coming out of the OOM issues with zora last week that the indexer could save as much token info that could be gleaned from the raw transfer event, and plugins later in the pipeline can fill out the rest of a token's information (basically plugins would be responsible for updating their own respective columns in the tokens table).

I think this means for this case we need to handle zora in the "generic" URI plugin, or if we have a "zora" plugin, that we'd somehow let the "generic" URI plugin know not to handle this zora contract, like have something dispatch tokens to the appropriate plugins based on some condition. Something to think about in the future though!

indexer/metadatas.go
```diff
@@ -418,3 +421,44 @@ func cryptopunks(ctx context.Context, turi persist.TokenURI, addr persist.Ethere
...
+	}
+	contentType, ok := util.FindFirstFieldFromMap(tokenMetadata, "mimeType", "contentType", "content-type", "type").(string)
+	var mediaType persist.MediaType
+	if ok {
+		mediaType = persist.MediaFromContentType(contentType)
+	} else {
+		mediaType, err = media.PredictMediaType(ctx, mediaURI)
+	}
+	switch mediaType {
+	case persist.MediaTypeImage, persist.MediaTypeGIF, persist.MediaTypeSVG:
```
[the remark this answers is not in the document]
bennycio: forgot bmps were images 🤦‍♂️

radazen (approved):

PR #500, opened 2022-09-14
[GAL-428] Use make deploy in CircleCI

jarrel-b:
### Changes
* Running  `make deploy-<env>-<service>`, now publishes a release to Sentry as the last step
* Updates our CircleCI deployments to use `make` so our services will get deployed in the same way as local deploys
* The indexer server also has its own Sentry project now to split up issues from the main indexer and issues from refreshes/syncs


Robinnnnn (approved):
SICKKK

also HBDDD 🎉

PR #501, opened 2022-09-14
[GAL-429] Collection refresh fixes

radazen:
In theory, we should be able to fix a broken user's gallery by refreshing each of their collections. In practice, I've run into several instances of [this sqlc bug](https://github.com/kyleconroy/sqlc/pull/1794), preventing me from refreshing a collection.

The bug is triggered when batch queries return errors, which is actually normal and expected for some types of errors (like a "not found" error when looking something up by ID). The real fix for this bug has already been merged into sqlc main and should be in the next release, but in the meantime, I've worked around it by:

- Looking up tokens by their collection instead of their ID
  - This is preferable anyway, and it also stops us from trying to look up deleted tokens
- Using a hand-written contractByID batch query
  - This is something I've wanted to for performance purposes -- instead of using sqlc batching, we use a single SQL query to look up many contracts by their IDs at the same time. It also lets us return "not found" errors without triggering the sqlc batching bug.


jarrel-b (approved):
lgtm!

Robinnnnn (approved):
🔥

PR #502, opened 2022-09-14
local setups [GAL-418]

bennycio:
Changes:

- **Changed** `_internal` to `_local`
- **Added** new local files for different environments like connecting to dev backend DB locally


indexer/core.go
```diff
@@ -78,7 +79,13 @@ func coreInit() (*gin.Engine, *indexer) {
 func coreInitServer() *gin.Engine {
 	ctx := sentry.SetHubOnContext(context.Background(), sentry.CurrentHub())
 
-	setDefaults("_internal/app-local-indexer-server.yaml")
+	path := "_local/app-local-indexer-server.yaml"
+	if len(os.Args) > 0 {
```
bennycio, line 83: This would look like `go run cmd/indexer/server/main.go prod`

server/server.go
```diff
@@ -138,15 +140,25 @@ func setDefaults() {
 	viper.AutomaticEnv()
 
 	if viper.GetString("ENV") == "local" {
+
+		filePath := "_local/app-local-backend.yaml"
+		if len(os.Args) > 1 {
+			if os.Args[1] == "dev" {
```
bennycio, line 146: This would look like `go run -tags debug_tools cmd/server/main.go dev`

Robinnnnn (approved):
thanks benny!!

radazen:
Neat! How does a `_local` .yaml file differ from its `_deploy` counterpart? E.g. what would the differences be between `_local/app-prod-backend.yaml` and `_deploy/app-prod-backend.yaml`?

And on that note, would `_local/app-prod-backend.yaml` set my `ENV` variable to `production`, or would it stay `local`?

bennycio:
> Neat! How does a `_local` .yaml file differ from its `_deploy` counterpart? E.g. what would the differences be between `_local/app-prod-backend.yaml` and `_deploy/app-prod-backend.yaml`?
> 
> And on that note, would `_local/app-prod-backend.yaml` set my `ENV` variable to `production`, or would it stay `local`?

The _local files just have the necessary variables to run the server locally connected to data that is either local, prod, or dev. Your env variable will still be local, these files only handle where data is coming from.

PR #514, opened 2022-09-16
last minute tez fixes

bennycio:
Changes:

- **Changed** keyword usage to be standardized by chain so that chains can handle what keywords to use for searching for a token
- **Added** keyword handlers for hic et nunc and fxhash to support their weird metadata
- **Changed** default env to use mainnet tez api
- **Changed** sync step to use the current db token in case the media is invalid on sync


service/media/media.go
```diff
@@ -685,6 +693,40 @@ func pipeIOForCmd(c *exec.Cmd, input []byte) ([]byte, error) {
...
+	return d
+}
+
+func (i TezImageKeywords) ForToken(tokenID persist.TokenID, contract persist.Address) []string {
+
+	switch contract {
+	case "KT1RJ6PbjHpwc3M5rw5s2Nbmefwbuwbdxton":
+		return []string{"displayUri", "image", "artifactUri", "uri"}
+	case "KT1BJC12dG17CVvPKJ1VYaNnaT5mzfnUTwXv":
+		return []string{"displayUri", "image", "artifactUri", "uri"}
```
Robinnnnn, line 706: could we label these with comments

Robinnnnn (approved):

PR #516, opened 2022-09-17
[GAL-444] Paginate comment + admire + combined interactions (1/2)

tbezman:
_**NOTE: Submitting this as a "part 1 of 2" so we can get the important parts merged into main.**_

## What's new?
- CommentsByFeedEventID and AdmiresByFeedEventID now comply with the Relay pagination spec
- Added general-purpose methods for paginating queries
- Clients can query for `interactions` to get both comments and admires in the same result stream
- AdmiresAPI and CommentsAPI have been merged into a single InteractionsAPI

## What's left?
A subsequent PR will include:
- Adding appropriate indexes to speed up comment/admire/interaction pagination
- Correct parameters to filter the `interactions` query by interaction type
- A few more dataloaders (not all of my new queries are currently wrapped in dataloaders)


tbezman (commented):
schema side lgtm

db/queries/core/query.sql
```diff
@@ -329,26 +332,59 @@ SELECT * FROM admires WHERE actor_id = $1 AND deleted = false ORDER BY created_a
...
 -- name: GetCommentsByActorIDBatch :batchmany
 SELECT * FROM comments WHERE actor_id = $1 AND deleted = false ORDER BY created_at DESC;
 
--- name: GetCommentsByFeedEventID :many
-SELECT * FROM comments WHERE feed_event_id = $1 AND deleted = false ORDER BY created_at DESC;
+-- name: PaginateInteractionsByFeedEventID :many
+SELECT interactions.created_At, interactions.id, interactions.tag FROM (
+    SELECT t.created_at, t.id, @admire_tag::int as tag FROM admires t WHERE @admire_tag != 0 AND t.feed_event_id = $1 AND t.deleted = false
+        AND (t.created_at, t.id) < (@cur_before_time, @cur_before_id) AND (t.created_at, t.id) > (@cur_after_time, @cur_after_id)
+                                                                    UNION
```
radazen, line 375: Is putting this `UNION` keyword all the way in the middle of the line a good idea? I don't know, but IntelliJ did it automatically at some point, and it looked better than any other formatting I tried.

db/queries/core/query.sql
```diff
@@ -329,26 +332,59 @@ SELECT * FROM admires WHERE actor_id = $1 AND deleted = false ORDER BY created_a
...
 -- name: GetCommentsByActorID :many
 SELECT * FROM comments WHERE actor_id = $1 AND deleted = false ORDER BY created_at DESC;
 
 -- name: GetCommentsByActorIDBatch :batchmany
 SELECT * FROM comments WHERE actor_id = $1 AND deleted = false ORDER BY created_at DESC;
 
--- name: GetCommentsByFeedEventID :many
-SELECT * FROM comments WHERE feed_event_id = $1 AND deleted = false ORDER BY created_at DESC;
+-- name: PaginateInteractionsByFeedEventID :many
+SELECT interactions.created_At, interactions.id, interactions.tag FROM (
```
radazen, line 372: It may not be obvious, but the `@admire_tag` and `@comment_tag` parameters serve two purposes: if non-zero, they tag the output rows with the supplied value so we know which rows are which interaction types. If zero, the corresponding `WHERE` clause will evaluate to `false` and the table won't be queried at all, which is a nice optimization that also allows us to filter which interactions we receive.

graphql/resolver/schema.resolvers.go
```diff
@@ -137,12 +138,102 @@ func (r *feedEventResolver) EventData(ctx context.Context, obj *model.FeedEvent)
...
+	if err != nil {
+		return nil, err
+	}
+
+	var edges []*model.FeedEventAdmireEdge
+	for _, admire := range admires {
+		edges = append(edges, &model.FeedEventAdmireEdge{
+			Node:   admireToModel(ctx, admire),
+			Event:  obj,
+			Cursor: "", // TODO: Cursors should all be forced resolvers here, since we typically won't need to create them
```
benny-conn, line 152: Could I get a little clarification on when we create them and when we don't just to help me understand personally how the cursor system works?

benny-conn (approved):
Looks fantastic! I just need some clarification on the cursors with regards to graph QL I guess. I love how the new pagination helpers handle them!

graphql/resolver/schema.resolvers.go
radazen: Ah, good question! The `Cursor` field here is part of the official spec, but it's a historical part of the spec that isn't used by current Relay. See: [this part of the Relay pagination spec](https://relay.dev/graphql/connections.htm#note-4ca64). The only cursors we actually create are the `PageInfo.StartCursor` and `PageInfo.EndCursor` fields.

I was actually just talking to Terence about this today. Relay will still fail to compile if we omit the `Cursor` field on edges entirely, but it looks like we can make it optional instead of mandatory, at which point we can just leave it null, and that sounds like a good compromise to me. We can assign a cursor there if it ever makes sense to do so, but for the time being, setting up a resolver for it would just be wasted effort.

I'll make it optional in my follow-up PR!

PR #519, opened 2022-09-19
[GAL-423] Add Alchemy RPC Tracing to Sentry

jarrel-b:
**Changes**
* Adds functions to create geth clients using either HTTP and websockets that have request tracing enabled
* Uses the new HTTP client for the indexer and indexer-api
* Now we can trace the duration of RPC calls to Alchemy and also have them available in Sentry

Sample JSON RPC tracing with the indexer: [sentry sample tracing](https://sentry.io/organizations/usegallery/performance/summary/?environment=local&project=6592179&query=&statsPeriod=1h&transaction=gethRPC&unselectedSeries=p100%28%29)

With these traces, we can set up alerts on Sentry based on whether we're making too many calls in some timeframe (like transactions per 15 minutes) and also alerts if Alchemy's response times increase.

Future improvements:
* Would be great to get the name of the RPC method called like `eth_filterlogs`
* Figure out error and success rates for monitoring


service/rpc/rpc.go
```diff
@@ -87,6 +90,62 @@ func NewEthClient() *ethclient.Client {
 
 }
 
+// NewEthHTTPClient returns a new http client with request tracing enabled
```
Robinnnnn, line 93: so gorgeous

service/rpc/rpc.go
```diff
@@ -87,6 +90,62 @@ func NewEthClient() *ethclient.Client {
...
+// Log sends trace information to Sentry.
+// Geth logs the duration of each RPC call: https://github.com/ethereum/go-ethereum/blob/master/rpc/handler.go#L242
+// We take advantage of this by configuring the root logger with a custom handler that parses records
+// that have useful metrics data attached such as the duration and the request id.
+func (h metricsHandler) Log(r *log.Record) error {
+	reqID := valFromSlice(r.Ctx, "reqid")
+
+	// A useful context isn't passed to the log record, so we use the background context here.
+	ctx := context.Background()
+	span, _ := tracing.StartSpan(ctx, "geth.wss", "rpcCall", sentryutil.TransactionNameSafe("gethRPC"))
```
Robinnnnn, line 136: any reason we're calling this span `geth.wss`? is this method specific to websockets?

Robinnnnn (approved):
🔥🔥🔥🔥🔥

service/rpc/rpc.go
jarrel-b: Yup this path is only used if the schema is wss://

PR #525, opened 2022-09-20
Notifications GraphQL Schema

tbezman:


graphql/schema/schema.graphql
```diff
@@ -977,6 +1013,90 @@ type RemoveCommentPayload {
...
+interface GroupedNotification implements Notification & Node {
+    id: ID!
+    seen: Boolean
+    creationTime: Time
+    updatedTime: Time
+
+    count: Int
+}
+
+type GroupNotificationUserEdge {
```
radazen, line 1032: Nit: probably want to stick with either `GroupedNotification` or `GroupNotification`

graphql/schema/schema.graphql
```diff
@@ -977,6 +1013,90 @@ type RemoveCommentPayload {
...
+    seen: Boolean
+    creationTime: Time
+    updatedTime: Time
+    count: Int
+
+    feedEvent: FeedEvent
+    admirers: GroupNotificationUsersConnection
+}
+
+type SomeoneCommentedOnYourFeedEventNotification implements Notification & Node{
```
radazen, line 1072: Should this implement `GroupedNotification`, since it has a `count` field?

graphql/schema/schema.graphql
```diff
@@ -396,11 +396,46 @@ type ViewerGallery {
...
+    """
+    Returns a list of notifications in reverse chronological order.
+    Seen notifications come after unseen notifications
+    """
+    notifications(before: String, after: String, first: Int, last: Int): NotificationsConnection
+
+    notificationSettings: NotificationSettings
 }
+
+type NotificationSettings implements Node {
```
radazen, line 423: This will need an explicit `gqlId` directive since it implements `Node` but doesn't have a `dbid` field. If we end up storing `NotificationSettings` in a separate DB table, we can just add the `dbid` field and use the DBID from the table. If not, we probably want to use a `goEmbedHelper` directive to store the parent user's ID here, and we can use that as the backing for the `Node` ID field. Happy to discuss implementation details further; just depends on how we actually store this on the backend!

graphql/schema/schema.graphql
```diff
@@ -977,6 +1013,90 @@ type RemoveCommentPayload {
   feedEvent: FeedEvent @goField(forceResolver: true)
 }
 
+interface Notification implements Node {
+    id: ID!
```
radazen, line 1017: Similar to the comment I left on `NotificationSettings`: this implements `Node` but doesn't have a `dbid` field, so it'll have to use `gqlId` to explicitly bind the parameters that will make up its `ID`. If we intend to have a `notifications` table in the database, we can just put DBIDs on all of our notification types and everything will be A-OK.

graphql/schema/schema.graphql
```diff
@@ -977,6 +1013,90 @@ type RemoveCommentPayload {
...
+    seen: Boolean
+    creationTime: Time
+    updatedTime: Time
+    count: Int
+
+    viewers: [GalleryUser]
+    gallery: Gallery
+}
+
+type ClearAllNotificationsPayload {
```
radazen, line 1096: This is potentially a question for the product side of things too, but how does a notification get cleared? Are they cleared just by virtue of seeing them, or do you have to click or hover or take some other action before an individual notification is cleared? Asking because it doesn't look like there's a mutation to clear an individual notification, and I'm wondering if clearing all notifications is sufficient for our purposes. cc: @kaitoo1

graphql/schema/schema.graphql
```diff
@@ -1014,4 +1134,13 @@ type Mutation {
...
     removeComment(commentId: DBID!): RemoveCommentPayloadOrError @authRequired
+
+    clearAllNotifications: ClearAllNotificationsPayload
+
+    updateNotificationSettings(settings: NotificationSettingsInput): NotificationSettings
+}
+
+type Subscription {
+    newNotification: Notification
+    notificationUpdated: Notification
```
radazen, line 1145: What's it mean for a notification to be updated?

graphql/schema/schema.graphql
```diff
@@ -977,6 +1013,90 @@ type RemoveCommentPayload {
...
 }
 
+interface Notification implements Node {
+    id: ID!
+    seen: Boolean
+    creationTime: Time
+    updatedTime: Time
+}
+
+interface GroupedNotification implements Notification & Node {
```
radazen, line 1023: Seems like most (or possibly all) of our current notifications are `GroupedNotifications` too. What kinds of notifications do we think won't be groupable?

graphql/schema/schema.graphql
tbezman: I think we want these to be grouped but we're only going to show the latest comment in the UI. Let's follow up in eng gm

graphql/schema/schema.graphql
tbezman: Afaik we're just updating the count of the notification.

PR #527, opened 2022-09-21
community page 1.5

benny-conn:
Changes:

- **Changed** `mediaprocessing` -> `tokenprocessing` so that this service can handle more async jobs with regards to tokens (such as finding all tokens in a collection and creating universal profiles where necessary)
- **Added** universal profile field to users
- **Added** a mutlichain function for refreshing tokens in collections that will go out and create universal users and grab tokens from a given collection for all of those users, saving them to the database
- **Changed** multichain `GetCommunityOwners` to also create universal profiles for users found that are not gallery users

Questions:
- will "other members" in the List View have image previews associated with them? not a hard requirement but seems like the data will be available

TODO:

- [x] pagination and sorting of owners
- [x] `multichain.GetDisplayNameByAddress`
- [x] testing
- [x] If user with chainAddress already exists and is a universal profile, it gets marked as deleted and the new account is created
- [x] initial page landing will kick off potentially a heavy process; this should be cached by every few hours
- [x] test with 10K sized collection


graphql/resolver/schema.resolvers.helpers.go
```diff
@@ -362,6 +366,65 @@ func resolveTokensByWalletID(ctx context.Context, walletID persist.DBID) ([]*mod
...
+		return err
+	}
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		logger.For(ctx).Errorf("failed to refresh tokens in collection: %s", util.GetErrFromResp(resp))
+	}
```
benny-conn, line 424: note to self: turn this into a task and have the token-processing service handle this sync

service/multichain/eth/eth.go
```diff
@@ -51,8 +52,12 @@ func (d *Provider) GetBlockchainInfo(ctx context.Context) (multichain.Blockchain
 }
 
 // GetTokensByWalletAddress retrieves tokens for a wallet address on the Ethereum Blockchain
-func (d *Provider) GetTokensByWalletAddress(ctx context.Context, addr persist.Address) ([]multichain.ChainAgnosticToken, []multichain.ChainAgnosticContract, error) {
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("%s/nfts/get?address=%s", d.indexerBaseURL, addr), nil)
+func (d *Provider) GetTokensByWalletAddress(ctx context.Context, addr persist.Address, limit, offset int) ([]multichain.ChainAgnosticToken, []multichain.ChainAgnosticContract, error) {
+	page := 0
+	if limit > 0 {
+		page = offset / limit
+	}
```
benny-conn, line 59: I don't like this so I am going to change it on the indexer side of things to just use offset. Pages suck

service/persist/postgres/user.go
```diff
@@ -274,13 +274,17 @@ func (u *UserRepository) GetByIDs(pCtx context.Context, pIDs []persist.DBID) ([]
...
-		rows.Scan(&user.ID, &user.Deleted, &user.Version, &user.Username, &user.UsernameIdempotent, &user.Bio, &user.Traits, pq.Array(&walletIDs), &user.CreationTime, &user.LastUpdated)
+		rows.Scan(&user.ID, &user.Deleted, &user.Version, &user.Username, &user.UsernameIdempotent, &user.Bio, &user.Traits, pq.Array(&walletIDs), &user.Universal, &user.CreationTime, &user.LastUpdated)
 		wallets := make([]persist.Wallet, len(walletIDs))
 
 		for i, walletID := range walletIDs {
 			wallet := persist.Wallet{ID: walletID}
 			err = u.getWalletStmt.QueryRowContext(pCtx, walletID).Scan(&wallet.Address, &wallet.Chain, &wallet.WalletType, &wallet.Version, &wallet.CreationTime, &wallet.LastUpdated)
 			if err != nil {
+				if err == sql.ErrNoRows {
+					i--
```
jarrel-b, line 285: Is this because it's needed that `wallets` is the same length as the input `walletIDs`? Wondering if wallets are missing, we might return an array of wallets with default wallet values at the end of the array.

service/multichain/multichain.go
```diff
@@ -232,14 +220,12 @@ func (p *Provider) SyncTokens(ctx context.Context, userID persist.DBID, chains [
 				for i, p := range providers {
 					go func(provider ChainProvider, priority int) {
 						defer subWg.Done()
-						tokens, contracts, err := provider.GetTokensByWalletAddress(ctx, addr)
+						tokens, contracts, err := provider.GetTokensByWalletAddress(ctx, addr, defaultLimit, 0)
```
jarrel-b, line 223: Is the limit because we're seeing requests timeout for really large collections? Would this be something we eventually want to paginate or run a bunch of smaller requests so that all tokens eventually get synced?

service/multichain/multichain.go
```diff
@@ -567,7 +475,291 @@ func (d *Provider) RefreshContract(ctx context.Context, ci persist.ContractIdent
...
+	return newTokens, nil
+}
+
+func (d *Provider) upsertContracts(ctx context.Context, allContracts []chainContracts) (map[string]persist.DBID, error) {
+	newContracts, err := contractsToNewDedupedContracts(ctx, allContracts)
+	if err != nil {
+		return nil, err
+	}
+	if err := d.Repos.ContractRepository.BulkUpsert(ctx, newContracts); err != nil {
+		return nil, fmt.Errorf("error upserting tokens: %s", err)
```
jarrel-b, line 739: Maybe `contracts` instead?

jarrel-b (commented):
Still need to read through `multichain.go`, but here are my comments so far 👍

service/multichain/multichain.go
benny-conn: This being here is arbitrary at the moment, could be there could not. I'll remove it for now, since we have not seen any issues with the size of a sync.

db/migrations/core/000028_universal_profile.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE users ADD COLUMN universal boolean NOT NULL DEFAULT false;
+DROP INDEX IF EXISTS users_username_idempotent_idx;
+CREATE UNIQUE INDEX users_username_idempotent_idx ON users (username_idempotent) WHERE deleted = false AND universal = false;
```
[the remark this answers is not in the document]
benny-conn: We could add this to the queries if we are positive that we will not be using the queries to retrieve universal profiles. For now, I guess I imagined this index is more for the purpose of enforcing uniqueness than for the queries.

graphql/schema/schema.graphql
```diff
@@ -55,6 +55,7 @@ type GalleryUser implements Node {
     username: String
     bio: String
     traits: String
+    universal: Boolean
```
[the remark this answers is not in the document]
benny-conn: This is addressed in one of your further comments where the user is being deleted during nonce creation 👍🏻

db/migrations/core/000028_universal_profile.up.sql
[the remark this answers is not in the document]
benny-conn: Never mind, some queries need it because we manually check things like is the username available. Thanks for pointing this out!

service/auth/auth.go
```diff
@@ -272,6 +272,11 @@ func GetAuthNonce(pCtx context.Context, pChainAddress persist.ChainAddress, user
 		logger.For(pCtx).WithError(err).Error("error retrieving user by address to get login nonce")
 	}
 
+	if user.Universal == true {
+		userRepo.Delete(pCtx, user.ID)
```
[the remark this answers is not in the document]
benny-conn: Moved this to where it makes more sense in the create user step 👍🏻

service/multichain/multichain.go
```diff
@@ -567,7 +474,291 @@ func (d *Provider) RefreshContract(ctx context.Context, ci persist.ContractIdent
 	return nil
 }
 
-func tokensToNewDedupedTokens(ctx context.Context, tokens []chainTokens, contractAddressIDs map[string]persist.DBID, ownerUser persist.User) ([]persist.TokenGallery, error) {
+// RefreshTokensForCollection refreshes all tokens in a given collection
+func (p *Provider) RefreshTokensForCollection(ctx context.Context, ci persist.ContractIdentifiers) error {
```
jarrel-b, line 478: I wonder if there's a different word we can use to distinguish an nft collection vs a gallery collection. Maybe even just `RefreshTokensForTokenCollection`?

service/multichain/multichain.go
```diff
@@ -567,7 +474,291 @@ func (d *Provider) RefreshContract(ctx context.Context, ci persist.ContractIdent
...
+			tid := tokenUniqueIdentifiers{chain: chainToken.chain, contract: agnosticToken.ContractAddress, tokenID: agnosticToken.TokenID, ownerAddress: agnosticToken.OwnerAddress}
+			if seenTokens[tid] {
+				continue
+			}
+			seenTokens[tid] = true
+			ct := chainToken
+			t := agnosticToken
+			wp.Submit(func() {
+				user, err := p.Repos.UserRepository.GetByChainAddress(ctx, persist.NewChainAddress(t.OwnerAddress, ct.chain))
+				if err != nil || user.ID == "" {
```
jarrel-b, line 614: Should this be an `&&`, or so long as we have some user ID we can drop the `err` check?

service/multichain/multichain.go
```diff
@@ -567,7 +474,291 @@ func (d *Provider) RefreshContract(ctx context.Context, ci persist.ContractIdent
...
+	for {
+		select {
+		case user := <-userChan:
+			logger.For(ctx).Debugf("got user %s", user.Username)
+			users[user.ID] = user
+			if userTokens[user.ID] == nil {
+				userTokens[user.ID] = map[int]chainTokens{}
+			}
+		case token := <-tokensForUserChan:
+			chainTokensForUser := userTokens[token.userID]
```
jarrel-b, line 691: Just double checking, is this guaranteed to always exist? Could there be a race condition we receive the token before we've saved the user?

service/multichain/multichain.go
benny-conn: if the user does exist, we can not worry about the error 👍🏻

service/multichain/multichain.go
benny-conn: Because the token gets sent on a chan after the user for that token gets sent, it would be impossible for this to ever get run before the first channel accepts the user 👍🏻

PR #528, opened 2022-09-22
Add more indexer tests

jarrel-b:
**Changes**
* Added more tests (contracts, erc1155s, checking specific fields in db) and also refactored the tests a bit
* Refactored the indexer to make it easier to test, namely moving `rpcEnabled` to a global var and splitting the indexer `Start` method into `catchUp` and `waitForBlocks`  methods


Robinnnnn (approved):
omfg this is huuuuge

PR #537, opened 2022-09-26
Add collectionsByIds query

radazen:
## What's new?

This PR adds a new `collectionsByIds` query to fetch multiple collections at a time. While putting this together, I realized that the automatic handling we use to map golang <-> GraphQL errors doesn't really work for resolvers that return a list of `SomethingOrError` types, so I've handled it manually here (and added comments about it).

It also adds `ErrInvalidInput` as an error type for `CollectionByIdOrError`, which was always a possible return type, but now it'll be handled appropriately.

### DIFF
--- a/publicapi/collection.go
+++ b/publicapi/collection.go
@@ -42,6 +42,44 @@ func (api CollectionAPI) GetCollectionById(ctx context.Context, collectionID per
 	return &collection, nil
 }
 
+func (api CollectionAPI) GetCollectionsByIds(ctx context.Context, collectionIDs []persist.DBID) ([]*db.Collection, []error) {
+	collectionThunk := func(collectionID persist.DBID) func() (db.Collection, error) {
+		// Validate
+		if err := validateFields(api.validator, validationMap{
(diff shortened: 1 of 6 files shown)


publicapi/collection.go
```diff
@@ -42,6 +42,44 @@ func (api CollectionAPI) GetCollectionById(ctx context.Context, collectionID per
...
+
+		return api.loaders.CollectionByCollectionId.LoadThunk(collectionID)
+	}
+
+	// A "thunk" will add this request to a batch, and then return a function that will block to fetch
+	// data when called. By creating all of the thunks first (without invoking the functions they return),
+	// we're setting up a batch that will eventually fetch all of these requests at the same time when
+	// their functions are invoked. "LoadAll" would accomplish something similar, but wouldn't let us
+	// validate each collectionID parameter first.
+	thunks := make([]func() (db.Collection, error), len(collectionIDs))
```
Robinnnnn, line 62: super cool!

Robinnnnn (approved):

PR #538, opened 2022-09-26
Add Sentry to mediaprocessing

jarrel-b:
**Changes**
* Adds sentry error handling and tracing to mediaprocessing
* Figured out a way to patch GCP HTTP calls in Sentry so now we can also trace reads/writes from storage too
* Uses this gcp client in the indexer as well

Sample trace: https://sentry.io/organizations/usegallery/performance/mediaprocessing:7de37786f4054995a263339a5b12ca64/?environment=local&project=6771759&query=&statsPeriod=24h&transaction=processMedia&unselectedSeries=p100%28%29

### DIFF
--- a/indexer/core.go
+++ b/indexer/core.go
@@ -8,11 +8,11 @@ import (
 	"strconv"
 	"time"
 
-	"cloud.google.com/go/storage"
 	"github.com/getsentry/sentry-go"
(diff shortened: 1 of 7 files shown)


radazen (approved):
TRACE ALL THE THIIIIINGS

PR #540, opened 2022-09-27
Add deploy diff

jarrel-b:
**Changes**
* You can run `make diff` to compare your version of `_encrypted_deploy` to what's checked into  `main`
* Running `make encrypt` runs the diff and also requires user input to confirm the changes are up to date


Robinnnnn (approved):
image

PR #546, opened 2022-09-30
[GAL-472] Add ffmpeg to mediaprocessing image

jarrel-b:
**Changes**
* Adds `ffmpeg` to mediaprocessing image
* Updates `_encrypted_deploy` with cloud run deployments

**Misc changes**
* Previously, a new cloud task client was created per feed event. Now a single client is initialized and gets re-used across requests
* Added a function `util.LoadEnvFile` which logs and loads the local yaml file gets used (I was getting confused which file I was using at times)
* Moved the functions to submit mediaprocessing messages from `multichain` to the `task` package as `CreateTaskForMediaProcessing` so we can get sentry tracing and re-use some of the helpers in that package


README.md
```diff
@@ -146,8 +146,13 @@ See [targeting services](https://github.com/aertje/cloud-tasks-emulator#targetin
...
                                             +-------------->  feedbot   |
-                                                           |  (:4123)   |
-                                                           +------------+
+                                            |              |  (:4123)   |
+                                            |              +------------+
+                                            |
+                                            |              +-------------------+
+                                            +-------------->  mediaprocessing  |
+                                                           |      (:6500)      |
+                                                           +-------------------+
```
Robinnnnn, line 155: thank you for doing this!

docker-compose.yml
```diff
@@ -45,3 +45,16 @@ services:
...
+  #     context: "."
+  #     dockerfile: "docker/mediaprocessing/Dockerfile"
+  #   ports:
+  #     - "6500:6500"
+  #   volumes:
+  #     - ${PWD}/_local:/app/_local
+  #     - ${PWD}/_deploy:/app/_deploy
+  #   depends_on:
+  #     - postgres
+  #     - redis
```
Robinnnnn, line 60: should the `postgres_indexer` block be optional as well?

feed/feed.go
```diff
@@ -35,7 +37,7 @@ func coreInit(pgx *pgxpool.Pool) *gin.Engine {
 		gin.SetMode(gin.DebugMode)
 	}
 
-	return handlersInit(router, pgx)
+	return handlersInit(router, pgx, task.NewClient(context.Background(), "./_deploy/service-key-dev.json"))
```
Robinnnnn, line 40: will this always use `service-key-dev`?

Robinnnnn (commented):
love the new config / quality of live improvements!

docker-compose.yml
jarrel-b: yea generally people don't run the indexer locally so will comment out

jarrel-b: doh actually we need this for our integration tests

server/server.go
```diff
@@ -259,10 +234,20 @@ func initSentry() {
 	}
 }
 
-func newMultichainProvider(repos *persist.Repositories, cache memstore.Cache, ethClient *ethclient.Client, httpClient *http.Client, ipfsClient *shell.Shell, arweaveClient *goar.Client, storageClient *storage.Client, tokenBucket string) *multichain.Provider {
+func newMultichainProvider(repos *persist.Repositories, cache memstore.Cache, ethClient *ethclient.Client, httpClient *http.Client, ipfsClient *shell.Shell, arweaveClient *goar.Client, storageClient *storage.Client, tokenBucket string, taskClient *cloudtasks.Client) *multichain.Provider {
 	ethChain := persist.ChainETH
-
-	return multichain.NewMultiChainDataRetriever(context.Background(), repos, cache, newTasksClient(), multichain.ChainOverrideMap{persist.ChainPOAP: &ethChain}, eth.NewProvider(viper.GetString("INDEXER_HOST"), httpClient, ethClient), opensea.NewProvider(ethClient, httpClient), tezos.NewProvider(viper.GetString("TEZOS_API_URL"), viper.GetString("MEDIA_PROCESSING_URL"), viper.GetString("IPFS_URL"), httpClient, ipfsClient, arweaveClient, storageClient, tokenBucket), poap.NewProvider(httpClient, viper.GetString("POAP_API_KEY"), viper.GetString("POAP_AUTH_TOKEN")))
```
radazen, line 265: You are a hero

feed/feed.go
radazen: I had the same reaction as Robin! Without jumping into the `task.NewClient` function, it's hard to know that this value is local-only, and I think it'd confuse me if I were just skimming through code. Since it's hardcoded anyway, can we move it into the `task.NewClient` function entirely so it's more obvious how it's being used? Or maybe change the parameter name to "localEnvKeyPath" or something?

radazen (approved):
Left one small request in a comment. Looks great to me! I love all of the refactoring, too.

docker/mediaprocessing/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+# Install deps
+WORKDIR /app
+COPY go.mod go.sum /app/
+RUN go mod download
+
+COPY . /app
+RUN go build -o ./bin/mediaprocessing ./cmd/mediaprocessing/main.go
+
+EXPOSE 6500
```
benny-conn, line 17: Should this expose 6500 or 8080? I believe cloud run will use 8080?? Unless you have got it working on 6500 already.

benny-conn (approved):
Looks great! One little comment, that is maybe actionable??

docker/mediaprocessing/Dockerfile
jarrel-b: got it working with port `6500`, thanks for checking!

PR #551, opened 2022-10-03
fixes for dave corp [GAL-485]

benny-conn:
Changes:

- **Changed** media grabbing to use the correctly mapped ipfs URLs for gateways so that we don't run into broken gateways. For example, `ipfs.infura.io` does not work anymore but some NFTs still use it. We should not rely on those API's being up when we can use the IPFS path with our own gateway and better guarantee success


jarrel-b (approved):
🚢

PR #552, opened 2022-10-03
[GAL-467] Fix tezos wallet refresh sync

jarrel-b:
**Changes**
* Fixes tezos wallet sync. Before, if we already had a token with media in the db, we would replace the provider token with the token we had in the db. This was an issue because the token still references wallets that have since been deleted. Now we only replace the `media`.

### DIFF
--- a/service/multichain/multichain.go
+++ b/service/multichain/multichain.go
@@ -311,10 +311,12 @@ outer:
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


Robinnnnn (approved):
😍 thank you!!

benny-conn (approved):

kaitoo1:
Lfg!! thanks for this

PR #555, opened 2022-10-04
[GAL-484] Update tokenprocessing base to bullseye

jarrel-b:
There is a bug in `ffmpeg v4.1.x`, which is the latest available version on the buster distro, upgrading to bullseye fixed the issue.


Robinnnnn (approved):
muy classico

PR #565, opened 2022-10-10
Indexer Fixups

jarrel-b:
**Changes**
* Changes indexer and indexer-api rpc transport from http to wss
* Before, a plugin would handle each incoming transfer one at a time. Now each plugin keeps a pool of workers to handle incoming transfers
* Removes `batchTransfers` because each plugin now handles their own buffering of messages


benny-conn (approved):
Love the async

PR #567, opened 2022-10-10
hasViewerAdmiredEvent property on FeedEvent

tbezman:
A little sweaty, but I separated the types for logged in / logged out so we know contextually what properties should / shouldn't be on a type.


graphql/schema/schema.graphql
```diff
@@ -467,14 +468,33 @@ interface FeedEventData {
     action: Action
 }
 
-type FeedEvent implements Node {
+interface FeedEvent {
+    id: ID!
+    dbid: DBID!
+    eventData: FeedEventData @goField(forceResolver: true)
+    admires: [Admire!] @goField(forceResolver: true)
+    comments: [Comment!] @goField(forceResolver: true)
```
jarrel-b, line 476: wonder if this could be simplified by having `hasViewerAdmiredEvent` live as an optional field on the FeedEvent type? if there's no viewer it'll be false or null?

tbezman: yeah there's some tradeoffs to be had with either. usually every "authed" field goes under the `viewer` type, but i was trying to avoid doubly fetching the "authed" and "unauthed" version of the types which is hard to do without a waterfall.

so the tradeoff is between
- type safe under the viewer field
- doubly fetching feeds

the annoying bug I'm thinking about is the following
1. user logs in
2. user admires event
3. user logs out
4. user sees that they've still admired a post

by keeping the type under the `Viewer` type, you ensure that when they log out, all that data disappears.

I think I'm more open to doubly fetching the data assuming that gzip will help with the bigger payload as most of the content will be the same

tbezman: Let's discuss this in eng gm

tbezman:
btw @jarrel-b , @radazen and I caught up and made the decision to do things the simpler way for now until it becomes a serious issues. I'll try to recap the issues in the eng gm tomorrow morning.

PR #572, opened 2022-10-12
[GAL-444] Paginate comment + admire + combined interactions (2/2) 

radazen:
### Clean-up items, including...
- Using dataloaders for all comment/admire/interaction queries
- Added indexes for comment/admire keyset pagination
- `pageInfo.Total` will now be present on subsequent connection queries if it was present on the initial one
- Since we don't actually use the `Cursor` field on comment/admire/interaction edges, it's optional now
- Interactions queried by feed event ID can now be filtered via `typeFilter` parameter
- Default to descending order within a page of results, but allow implementors to choose ascending where appropriate


jarrel-b (approved):
Looks great, will add these changes!

PR #573, opened 2022-10-12
Sentry error reporting tweaks

jarrel-b:
**Changes**
* Updates the sentry logger to properly handle the generic `errors.errorString` so that they are grouped separately


radazen (approved):
👍 👍 👍

PR #574, opened 2022-10-12
[GAL-511] Add hasViewerAdmiredEvent field to FeedEvent

radazen:
## What's new?

- The GraphQL `FeedEvent` type now has a field that lets clients know whether the current user has admired the event. Returns true if a user is logged in and has admired the event, false otherwise.


jarrel-b (approved):
🚢

PR #576, opened 2022-10-12
[GAL-380] Update feed schema to handle deleted collections and tokens

jarrel-b:
This is a proposal of a schema to handle the cases where a token is no longer owned by a user or a collection has been deleted. The idea is that these types may or may not be `deleted`. In the case where they are `deleted`, we won't link to it if it's a deleted collection. If it is a token, I'm thinking we can either :
* not resolve to anything
* resolve to showing stale data (the token as it was before it got deleted)
* resolve to the current owner in Gallery (there could be multiple owners if users haven't synced and if it was transferred frequently, but we can decided based on the block number)
* longterm - refresh that token and resolve to the universal page

Considerations:
* May need to do something similar for when we have multi-gallery support when users can delete galleries if that is something we choose to display on the feed

### DIFF
--- a/graphql/schema/schema.graphql
+++ b/graphql/schema/schema.graphql
@@ -525,6 +525,35 @@ type FeedEvent implements Node {
     hasViewerAdmiredEvent: Boolean @goField(forceResolver: true)
 }
 
+# FeedCollection represents a snapshot of a collection at a point in time.
+# Therefore, it's properties and its assets (i.e. its tokens) in the collection
+# may be stale and could no longer be owned by the user.
+type FeedCollection implements Node {
+    id: ID!
+    dbid: DBID!
+    deleted: Boolean
+    version: Int
+    name: String
+    collectorsNote: String
+    gallery: Gallery @goField(forceResolver: true)
+    layout: CollectionLayout
+    hidden: Boolean
+    tokens: [FeedToken] @goField(forceResolver: true)
+}
+
+# FeedToken represents a snapshot of a token at a point in time.
+# It's properties may be out of date and are no longer accurate.
+type FeedToken implements Node
+@goEmbedHelper
+@goGqlId(fields: ["tokenId", "collectionId"]) {
+    id: ID!
+    dbid: DBID!
+    deleted: Boolean
+    token: Token
+    collection: FeedCollection
+    tokenSettings: CollectionTokenSettings @goField(forceResolver: true)
+}
+
 type UserCreatedFeedEventData implements FeedEventData {
     eventTime: Time
     owner: GalleryUser @goField(forceResolver: true)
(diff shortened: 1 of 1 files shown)


Robinnnnn (commented):
so freaking excited about this, thank you based ser!

re: stale token resolution, the question is: do we want to be able to preserve history such that we can visually see what tokens were added to a collection, even after the collection/token get deleted?

i believe the answer is a definite yes. the act of curating and showcasing a token should ultimately be part of the token's provenance. it'd be really cool to see how a token moves around and gets curated throughout its lifetime.

to "guarantee" this, it seems like the easiest thing would be to store the token's contract address + ID (or some way to retrieve it) and reveal it behind a universal profile. so that should be the north star.

as an intermediary step, if it's not too hard, it'd be great to resolve to the current gallery owner. i'm pretty sure there can't be multiple owners of a token in our current setup – if user A syncs their wallet and owns some tokens that belong to user B, i'm prettyyy sure the tokens actually get removed from user B's gallery.

the more i think about this, the more i feel like tokens should never get `deleted` from our DB. even if a user disconnects their wallet, it'd be useful to have it around for cases like this, and/or universal token profiles.

jarrel-b:
> so freaking excited about this, thank you based ser!
> 
> re: stale token resolution, the question is: do we want to be able to preserve history such that we can visually see what tokens were added to a collection, even after the collection/token get deleted?
> 
> i believe the answer is a definite yes. the act of curating and showcasing a token should ultimately be part of the token's provenance. it'd be really cool to see how a token moves around and gets curated throughout its lifetime.
> 
> to "guarantee" this, it seems like the easiest thing would be to store the token's contract address + ID (or some way to retrieve it) and reveal it behind a universal profile. so that should be the north star.
> 
> as an intermediary step, if it's not too hard, it'd be great to resolve to the current gallery owner. i'm pretty sure there can't be multiple owners of a token in our current setup – if user A syncs their wallet and owns some tokens that belong to user B, i'm prettyyy sure the tokens actually get removed from user B's gallery.
> 
> the more i think about this, the more i feel like tokens should never get `deleted` from our DB. even if a user disconnects their wallet, it'd be useful to have it around for cases like this, and/or universal token profiles.

@Robinnnnn confirmed that syncs only affect the current user's gallery so, in your example, User B's tokens would not get removed from their gallery when User A syncs.

Also, by `deleted`, do you mean soft deletes or hard deletes? We only ever soft delete tokens from the db (and for everything else actually), which is analogous to meaning that the user no longer owns that token (as a result of them syncing), but we still retain all the data we have for that token.

Robinnnnn:
> confirmed that syncs only affect the current user's gallery so
gotcha, thanks @jarrel-b !

> Also, by deleted, do you mean soft deletes or hard deletes
oh yeah i figured that we never hard delete, so i was wondering if we should even soft-delete. in other words, maybe we can keep it as `deleted:false` and just drop the owner ID or something, so that it's still queryable (if not by ID, at least by contract address / token ID) in the case of a stale collection / token

PR #577, opened 2022-10-13
Sanitize/validate comments

radazen:
## What's new?
- Comments are now sanitized (the same way we sanitize other user-generated content)
- Empty string comments are no longer valid

Follow-up question: how should we handle empty space or trailing whitespace in comments? Should a comment that consists entirely of whitespace be rejected? Should trailing whitespace on comments be trimmed? Should leading whitespace be trimmed? cc: @kaitoo1 @Robinnnnn

### DIFF
--- a/publicapi/interaction.go
+++ b/publicapi/interaction.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


jarrel-b (approved):
amazing

PR #580, opened 2022-10-14
[GAL-524] [GAL-525] Fix pagination ordering + remove admire by feed event ID

radazen:
## What's new?
- Reverted a change to pagination that allowed for different sort orders within a page of results. This actually violates the Relay connection spec and makes things weird on the frontend
- Changed `removeAdmire` to accept a feedEventID instead of an admireID
- Schema update: `removeAdmire` now returns the ID of the removed admire, too
- Refactored `hasUserAdmiredEvent` to use the same underlying query as `removeAdmire`, since both need to look up admires by user and feed event


tbezman (approved):
schema changes lgtm!

Robinnnnn (approved):

PR #583, opened 2022-10-15
More comment/admire tweaks

radazen:
## What's new?
- Replaced `hasViewerAdmiredEvent` with `viewerAdmire`, the latter of which returns an `Admire` type instead of a bool. As it turns out, having the `admire` is more useful than just knowing it exists.
- `hasViewerAdmiredEvent` is still present -- for now -- so we don't break existing clients. It always returns false and should be removed once a new frontend without it has been deployed for a while.
- Adding/removing comments and admires will now return appropriate error types for handled errors: `ErrAdmireNotFound`, `ErrAdmireAlreadyExists`, and `ErrCommentNotFound`. These previously returned unhandled errors that weren't in our schema.
- The `removeAdmire`mutation takes an admireID again, instead of a userID + feedEventID (which I switched it to briefly before we settled on this latest design)

### DIFF
--- a/graphql/generated/generated.go
+++ b/graphql/generated/generated.go
@@ -254,11 +254,23 @@ type ComplexityRoot struct {
 		Message func(childComplexity int) int
 	}
 
+	ErrAdmireAlreadyExists struct {
+		Message func(childComplexity int) int
+	}
+
+	ErrAdmireNotFound struct {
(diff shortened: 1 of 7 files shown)


Robinnnnn (approved):
😍😍😍

jarrel-b (approved):

PR #586, opened 2022-10-17
only return displayed badges

benny-conn:
Changes:

- **Changed** the badges resolver to use a new dataloader for resolving the contracts that a user owns tokens from. Rather than using `ContractsByUserID` it uses `ContractsDisplayedByUserID` which uses this sql query under the hood:

```sql
SELECT DISTINCT ON (contracts.id) contracts.* FROM contracts, tokens
    JOIN collections c ON c.NFTS && ARRAY[tokens.id]
    WHERE tokens.owner_user_id = $1 AND tokens.contract = contracts.id
    AND tokens.deleted = false AND contracts.deleted = false;
```

This query is tested and works well, for my user `benny` it ran in 1 secs 685 msec time. This is definitely a more expensive query than just querying for contracts that are attached to tokens owned by a user.


Robinnnnn:
@benny-conn thank you so much for this!

FWIW ~1.7 seconds seems like it's quite expensive and could easily be the bottleneck for querying a user's profile. how hard would it be to cache this value, and update it every few minutes?

benny-conn:
> FWIW ~1.7 seconds seems like it's quite expensive and could easily be the bottleneck for querying a user's profile. how hard would it be to cache this value, and update it every few minutes?

The dataloaders should help to cache this value, this was just the raw time for an uncached query. If I find that the dataloaders caching is not caching enough for long enough through some more testing then I will see about using redis here 👍🏻

radazen:
> FWIW ~1.7 seconds seems like it's quite expensive and could easily be the bottleneck for querying a user's profile. how hard would it be to cache this value, and update it every few minutes?

I'd love to pair on this tomorrow if you've got the time, @benny-conn! Feels like we should be able to get that time down somehow.

db/queries/core/query.sql
```diff
@@ -125,6 +125,18 @@ SELECT DISTINCT ON (contracts.id) contracts.* FROM contracts, tokens
     WHERE tokens.owner_user_id = $1 AND tokens.contract = contracts.id
     AND tokens.deleted = false AND contracts.deleted = false;
 
+-- name: GetContractsDisplayedByUserID :many
```
radazen, line 128: I originally thought it was a good idea to make batched and unbatched versions of all our queries, but Jarrel made an excellent point a while back: we typically only use one version, so there's not much reason to add both until we actually need both. In this case, I think you could drop the non-batched version since we only use the batched version.

PR #587, opened 2022-10-18
emails

benny-conn:
This PR adds a new service called `emails` that can run in cloud run and will be responsible with shooting out emails to users whether that be verification emails or notification emails. This service should be set up in a way that is not accessible to traffic outside of GCP (this is the default for new services) and will be communicated to by both our backend as well as cloud scheduler jobs for automated emails. This service also has access to the backend database (write and read, although we can make it read only if we decide to handle updating verification statuses on the backend instead of on the email service).

## TLDR completion points
**note**: this is all still early so even complete points might not be 100% complete, but are functional right now

- ✅ verification process
- ½ notification emails (sending of emails good, triggering of emails not done)
- ✅ adding emails to users on signup
- ✅ subscribing users on sendgrid for single marketing email sends

## TLDR main TODO points for future PRs

- updating emails
- updating notification setting and unsubscribing
- triggering of notification emails on a schedule

# Summary of infrastructure plan for emails

## Verification process
When a user creates their account on the backend, the backend service would call the email service's verification email endpoint and the email service will respond with whether it successfully sent the verification email to the user. 

When the user verifies their email using the link in the email (the link will be to a new frontend page with some query parameter that will be a JWT signed by the email service and probably just immediately redirect them to a success or failure page on that request's result) that frontend page will call a backend endpoint with the JWT and the backend will then call the email service to verify that the JWT is valid.

The email service will update the user's verification status to true if the JWT is valid or return an error if not (the backend could also do the update, and then we can make the email service read only, good with either).

## Automated emails
On GCP we can create scheduled jobs that can call out to the email service using http ([docs for scheduler](https://cloud.google.com/scheduler/docs/schedule-run-cron-job) (these docs use pub/sub in the examples but we can use http as well to directly call the endpoints)). The endpoint responding will iteratively paginate through every user with the right unsubscribe statuses (are they unsubscribed to all emails or the specific type of email we are trying to send?) and concurrently send each user an email using go templating to fill out the data for that user.

## Single Emails

Single emails can be sent directly from SendGrid using the "[Single Email](https://mc.sendgrid.com/single-sends?view=raw)" feature on their app webpage. Anyone with access can create their own templates using a templating dashboard or html/plain text directly. The recipients for the single email will have to be consistent with the emails signed up on gallery so at the end of the verification process, we also ensure that a user's email is added to a SendGrid Contact List (we can create multiple lists for different types of emails but I created one default list that is supposed to just be all of our user's emails) which can then be used to send emails to. We also want to make sure that when user's update their unsubscriptions, they are added to the appropriate "Unsubscribe group" on SendGrid or removed from the appropriate Contact List.

## Considerations

- Verification emails use JWTs for authentication. These JWTs are signed by a different service (with a different signing secret) than the backend so a user couldn't just send their cookie along and have it work. I have not found too much online as far as practices go for this but I do need to more research here. If anyone already has experience with or knowledge of a better way to best verify emails in a secure way, or believe this way is totally secure enough, let me know!
- Should the email verification service handle updating the user verification status or should the backend handle this? Currently the email service goes to the database to read information so that for example we can get the email directly from a user ID during the verification process rather than the service being sent an email as an input and the service just handling the email. Is it then alright that the service can write to the DB as well?

## TODO's after merge

- [ ] Migration
- [ ] Schema changes (not breaking)


emails/send.go
```diff
@@ -0,0 +1,163 @@
...
+		err := runForUsersWithNotificationsOnForEmailType(c, persist.EmailTypeNotifications, queries, func(u coredb.User) error {
+
+			from := mail.NewEmail("Gallery", viper.GetString("FROM_EMAIL"))
+			to := mail.NewEmail(u.Username.String, u.Email.String())
+			m := mail.NewV3Mail()
+			m.SetFrom(from)
+			p := mail.NewPersonalization()
+			m.SetTemplateID(viper.GetString("SENDGRID_NOTIFICATIONS_TEMPLATE_ID"))
+			p.DynamicTemplateData = map[string]interface{}{
+				"notifications": []string{"notification 1", "notification 2"},
```
benny-conn, line 100: test data for now

graphql/schema/schema.graphql
```diff
@@ -1295,6 +1339,8 @@ type Mutation {
     clearAllNotifications: ClearAllNotificationsPayload @authRequired
 
     updateNotificationSettings(settings: NotificationSettingsInput): NotificationSettings
+
+    verifyEmail(token: String!): VerifyEmailPayloadOrError
```
jarrel-b, line 1343: Would we want this to accept an email they intended to verify rather than the jwt itself? I don't think we're storing anything sensitive in the jwt, but seems a bit odd to pass it in as a param. Maybe the jwt is included in the headers and verified with the `authRequired` directive?

graphql/schema/schema.graphql
```diff
@@ -446,35 +446,58 @@ type NotificationsConnection @goEmbedHelper {
...
+  someoneAdmiredYourUpdate: Boolean
+  someoneCommentedOnYourUpdate: Boolean
+  someoneViewedYourGallery: Boolean
 }
 
+enum EmailVerificationStatus {
+    Unverified
+    Verified
+    Failed
+    Admin
```
jarrel-b, line 487: Would this just be for while this feature is being developed, or what would authorization would `Admin` confer? It doesn't seem to be an email status per se, more like a role on the user itself?

publicapi/auth.go
```diff
@@ -55,12 +55,13 @@ func (api AuthAPI) NewDebugAuthenticator(ctx context.Context, debugParams model.
 			userID = *debugParams.UserID
 		}
 
-		user, err := api.repos.UserRepository.GetByID(ctx, userID)
-		if err != nil {
-			return nil, fmt.Errorf("debug auth failed: %w", err)
```
jarrel-b, line 60: Do we still want to exit here if we failed to fetch the user, or its fine to continue?

publicapi/user.go
```diff
@@ -213,7 +214,7 @@ func (api UserAPI) RemoveWalletsFromUser(ctx context.Context, walletIDs []persis
 	return nil
 }
 
-func (api UserAPI) CreateUser(ctx context.Context, authenticator auth.Authenticator, username string, bio string) (userID persist.DBID, galleryID persist.DBID, err error) {
+func (api UserAPI) CreateUser(ctx context.Context, authenticator auth.Authenticator, username string, email string, bio string) (userID persist.DBID, galleryID persist.DBID, err error) {
 	// Validate
 	if err := validateFields(api.validator, validationMap{
 		"username": {username, "required,username"},
```
jarrel-b, line 220: Does the validation lib we use have some sort of email validator we can use? would be nice to verify it before saving it on the user.

emails/send.go
```diff
@@ -0,0 +1,163 @@
...
+			util.ErrResponse(c, http.StatusBadRequest, err)
+			return
+		}
+
+		if user.Email == "" {
+			util.ErrResponse(c, http.StatusBadRequest, fmt.Errorf("user %s has no email", input.UserID))
+			return
+		}
+
+		j, err := auth.JWTGeneratePipeline(c, input.UserID)
```
jarrel-b, line 59: Just curious are we setting a ttl for how long the jwt is valid?

graphql/resolver/schema.resolvers.helpers.go
```diff
@@ -903,6 +927,30 @@ func resolveCommentByCommentID(ctx context.Context, commentID persist.DBID) (*mo
 	return commentToModel(ctx, *comment), nil
 }
 
+func verifyEmail(ctx context.Context, token string) (*model.VerifyEmailPayload, error) {
+	output, err := emails.VerifyEmail(ctx, token)
```
jarrel-b, line 931: Thinking we could check that the email is already verified before sending a verification email?

service/persist/email.go
```diff
@@ -0,0 +1,92 @@
...
+
+func (e EmailVerificationStatus) String() string {
+	return emailVerificationStatuses[e]
+}
+
+func (e EmailVerificationStatus) MarshalGQL(w io.Writer) {
+	w.Write([]byte(fmt.Sprintf(`"%s"`, e.String())))
+}
+
+func (e *EmailVerificationStatus) UnmarshalGQL(v interface{}) error {
```
jarrel-b, line 78: Woah do you need all of this for `EmailVerificationStatus` even if the underlying type is just a scalar? Seems like we didn't need this for the `Chain` type?

kaitoo1:
Nice @benny-conn ! 
+1 to what @jarrel-b  said, would you be able to diagram the system for documentation? I'm sure it'll help reviewers understand this PR better too!

db/queries/core/query.sql
```diff
@@ -531,5 +531,31 @@ SELECT count(*), @comment_tag::int as tag FROM comments t WHERE @comment_tag !=
 -- name: GetAdmireByActorIDAndFeedEventID :batchone
 SELECT * FROM admires WHERE actor_id = $1 AND feed_event_id = $2 AND deleted = false;
 
+
+-- for some reason this query will not allow me to use @tags for $1
+-- verified is commented out for testing purposes so I don't have to verify an email to send stuff to it
+-- name: GetUsersWithNotificationsOnForEmailType :many
+SELECT * FROM users WHERE (email_unsubscriptions->>'all' = 'false' OR email_unsubscriptions->>'all' IS NULL) AND (email_unsubscriptions->>$1::varchar = 'false' OR email_unsubscriptions->>$1::varchar IS NULL) AND deleted = false AND email IS NOT NULL -- AND email_verified = true
```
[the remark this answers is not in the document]
benny-conn: You would think! I tried this and sqlc just skips the tag and does not include it as a parameter. Maybe something we can suggest to them if they haven't already started implementing a fix for whatever problem this is.

db/queries/core/query.sql
```diff
@@ -531,5 +531,31 @@ SELECT count(*), @comment_tag::int as tag FROM comments t WHERE @comment_tag !=
...
+    AND (created_at, id) < (@cur_before_time, @cur_before_id)
+    AND (created_at, id) > (@cur_after_time, @cur_after_id)
+    ORDER BY CASE WHEN @paging_forward::bool THEN (created_at, id) END ASC,
+             CASE WHEN NOT @paging_forward::bool THEN (created_at, id) END DESC
+    LIMIT $2;
+
+-- verified is commented out for testing purposes so I don't have to verify an email to send stuff to it
+-- name: GetUsersWithNotificationsOn :many
+SELECT * FROM users WHERE (email_unsubscriptions->>'all' = 'false' OR email_unsubscriptions->>'all' IS NULL) AND deleted = false AND email IS NOT NULL -- AND email_verified = true
```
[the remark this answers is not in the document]
benny-conn: I imagine that `all` will be a sort of override so that when checking for if they are subscribed, we pass in the specific field for the email type, for example `notification_emails`, and if `all` for that user is set to `true` then we don't even check what status they have for `notification_emails`. If `all` is `false` then we just check if `notification_emails` is false as well. This is why the first check in the `GetUsersWithEmailNotificationsOnForEmailType` query is `WHERE (email_unsubscriptions->>'all' = 'false' OR email_unsubscriptions->>'all' IS NULL) AND ...`.

Updating those settings will be in a future PR btw.

emails/emails.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	if viper.GetString("ENV") != "local" {
+		util.EnvVarMustExist("SENTRY_DSN", "")
+		util.EnvVarMustExist("VERSION", "")
+	}
+}
+
+func newThrottler() *throttle.Locker {
+	return throttle.NewThrottleLocker(redis.NewCache(redis.TokenProcessingThrottleDB), time.Minute*5)
```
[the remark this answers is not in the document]
benny-conn: Yes indeed!

graphql/resolver/schema.resolvers.go
```diff
@@ -715,7 +715,12 @@ func (r *mutationResolver) CreateUser(ctx context.Context, authMechanism model.A
 		bioStr = *bio
 	}
 
-	userID, galleryID, err := publicapi.For(ctx).User.CreateUser(ctx, authenticator, username, bioStr)
+	emailStr := ""
+	if email != nil {
+		emailStr = *email
+	}
+
+	userID, galleryID, err := publicapi.For(ctx).User.CreateUser(ctx, authenticator, username, emailStr, bioStr)
```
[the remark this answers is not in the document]
benny-conn: We would be writing empty strings just the same way we do for the bio. Let me see if we can get around this.

graphql/schema/schema.graphql
benny-conn: This token is the string that is included at the end of the link in the verification email. It is not able to be forged because it is signed by the email server and it is also not going to be stored in any cookie or auth header because we want the user to be able to verify from their phone for example even if they are signing in on their computer. If we just passed along the email, anyone could verify any email without actually proving they have had some sort of access to that email address's inbox.

To be clear, this email does not trigger the verification email, this is used to verify the token in the verification email.

graphql/schema/schema.graphql
benny-conn: After talking with Robin, I thought this might be useful mainly for testing in development. Basically this would be a version of `Verified` that is used in place of `Verified` on dev and local so we can shoot emails to our admin emails.

publicapi/auth.go
benny-conn: It's fine to continue because we could be creating a user in the debug auth, in which case there would be no user yet.

publicapi/user.go
benny-conn: Yes indeed, we are actually going to be using a sendgrid validation service to validate the emails up front, but will be saving that for a future PR. Put in a TODO where it is going to be.

emails/send.go
benny-conn: Yes indeed, and that will be set by `JWT_TTL` env var

service/persist/email.go
benny-conn: We actually do do it this way currently for the `Chain` type as well, unfortunately hahah

benny-conn: I ran into an issue with this before where when I added a chain, it wasn't working because I forgot to include the marshaling into the chain type's marshal GQL func. Will add this to fix it week because it is stinky!

graphql/schema/schema.graphql
jarrel-b: Makes sense, thanks for explaining!

PR #588, opened 2022-10-19
quickfix

benny-conn:
Changes:

- **Changed** SQL query to use matching order by
- **Changed** local file to be up to date with what I use locally to run everything


radazen (approved):
👍 👍 Thanks for being so quick on this!

PR #591, opened 2022-10-20
[GAL-415] Add Feed Captioning

jarrel-b:
**Changes**
* Adds an optional `caption` input to CreateCollection and AddTokensToCollection mutations
* Adds a type, `CapturedAction`, which returns a `FeedEvent` if an event was captured in the handling of the above mutations
* Refactors the `eventDispatcher` to include `immediateHandlers` which returns the result of handling an event (@benny-conn can help with merging your changes in the notifications PR)

**Post Merge**
* [ ] Run migrations to add `caption` to `events` and `feed_events` table


graphql/schema/schema.graphql
```diff
@@ -561,6 +562,14 @@ type FeedEvent implements Node {
     hasViewerAdmiredEvent: Boolean @goField(forceResolver: true)
 }
 
+type ErrActionNotCapturable implements Error {
+  message: String!
+}
+
+union CapturedAction = 
+  FeedEvent
+  | ErrActionNotCapturable
```
jarrel-b, line 571: Is this weird as an Error? It's not an application error, but something clients should handle.  Using `Error` was convenient, but should this be its own type?

event/event.go
```diff
@@ -42,43 +46,94 @@ func DispatchEventToFeed(ctx context.Context, event db.Event) error {
...
+func newFeedHandler(queries *db.Queries, taskClient *cloudtasks.Client) feedHandler {
+	return feedHandler{
+		eventRepo:    postgres.EventRepository{Queries: queries},
+		eventBuilder: feed.NewEventBuilder(queries, true),
+		tc:           taskClient,
+	}
+}
 
+// Submit creates a delayed task for the Feed service to handle later.
+func (h feedHandler) Submit(ctx context.Context, event db.Event) error {
```
radazen, line 122: Are `Submit`/`Call` standard terms for these sorts of operations? If so, disregard my comment here! If not, could we give them more descriptive names? The doc comments above them are awesome, but I feel like the method names could make their functionality more obvious.

feed/feed.go
```diff
@@ -37,7 +38,7 @@ func coreInit(pgx *pgxpool.Pool) *gin.Engine {
 		gin.SetMode(gin.DebugMode)
 	}
 
-	return handlersInit(router, pgx, task.NewClient(context.Background()))
+	return handlersInit(router, db.New(pgx), task.NewClient(context.Background()))
```
radazen, line 41: 🔥

graphql/resolver/schema.resolvers.helpers.go
```diff
@@ -668,6 +668,27 @@ func feedEventToDataModel(event *db.FeedEvent) (model.FeedEventData, error) {
 	}
 }
 
+func eventToModel(event *db.FeedEvent) (*model.FeedEvent, error) {
+	data, err := feedEventToDataModel(event)
+	if err != nil {
+		return nil, err
+	}
+
+	// Value always returns a nil error so we can safely ignore it.
```
radazen, line 677: I could not love this comment more

graphql/schema/schema.graphql
jarrel-b: I tried this originally and found I needed to add a bunch of guard clauses everywhere to check if the feed ID was null, but now I realize I was doing something silly so will try to go this way!

radazen (approved):
This is awesome! I love the design. I left some nitpicky method naming questions, because that's who I am as a person, but everything looks 👌 to me!

event/event.go
```diff
@@ -42,43 +46,94 @@ func DispatchEventToFeed(ctx context.Context, event db.Event) error {
...
-type eventDispatcher struct {
-	handlers map[persist.Action]eventHandler
+func newEventDispatcher() eventDispatcher {
+	return eventDispatcher{
+		handlers:          map[persist.Action]eventHandler{},
+		immediateHandlers: map[persist.Action]immediateHandler{},
+	}
 }
 
 func (d *eventDispatcher) AddHandler(action persist.Action, handler eventHandler) {
```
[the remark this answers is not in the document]
jarrel-b: makes sense!

graphql/plugin/gqlidgen/code_generator.gotpl
```diff
@@ -21,6 +21,10 @@ func (r *{{$node.Name}}) ID() GqlID {
         {{- end }}
     //-----------------------------------------------------------------------------------------------
     {{- end }}
+    // It's safe to return the same ID since the node itself will resolve to nil and the ID never gets used.
+    if r == nil {
```
jarrel-b, line 25: @radazen had to add this line but this should be ok, lmk what you think!

radazen: Ah, interesting. I ran into something similar with making the `viewerAdmire` field nullable (w.r.t. trace span logging), but I assumed that was a special case because we check for the GqlID interface and call the `ID()` method directly there. Where are you seeing nil IDs causing trouble?

jarrel-b: Without the addition, I'd run into pointer dereference errors for queries like below, where if a caption was provided, it wants the event returned too.

```
# $input does not contain a caption

mutation useCreateCollectionMutation {
  createCollection(input: $input) {
    ... on CreateCollectionPayload {
      feedEvent {
        __typename
      }
    }
  }
}
```

Also happy to pair on this next week too

radazen: Gotcha. Yeah, let's follow up next week! I've been under the impression that gqlgen (and GraphQL in general) wouldn't attempt to resolve child fields on something that comes back `null`, but maybe it still attempts to resolve the ID?

My only concern with making GqlID _not_ throw an error is that we might silently ignore some real errors (referencing the ID of a thing that shouldn't exist, getting the empty string, treating it as a valid ID in the Relay cache or something, etc).

jarrel-b: Sounds good!

jarrel-b: I thought so too, seems like gqlgen chooses to always run extensions even if a field is null?

radazen: Ah! Are you specifically hitting the tracing error in the FieldReporter extension? Or is it somewhere else?

jarrel-b: doh yes pulling in `main` fixed it!

PR #597, opened 2022-10-24
use different time marshalling

benny-conn:
Changes:

- **Changed** time marshaling to use JSON instead of binary (binary was causing some bad utf8 characters to appear but JSON will always be valid utf8)
- **Changed** token processing to also update name and description because previously the tokens would just keep the name and description they have when being updated which is often empty.

### DIFF
--- a/publicapi/pagination.go
+++ b/publicapi/pagination.go
@@ -3,8 +3,10 @@ package publicapi
 import (
 	"encoding/base64"
 	"time"
+	"unicode/utf8"
 
 	"github.com/go-playground/validator/v10"
+	"github.com/mikeydub/go-gallery/service/logger"
 	"github.com/mikeydub/go-gallery/service/persist"
 	"github.com/mikeydub/go-gallery/validate"
 )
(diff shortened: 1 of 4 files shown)


radazen:
Hmm, something about this feels off to me. I don't understand the connection between binary marshaling and UTF-8.

`time.Time.MarshalBinary` serializes to 15 or 16 bytes, and then base64 encoding turns those bytes into base64 runes that are UTF-8 compatible (and even ASCII compatible). `MarshalJSON` will create bytes that get base64-encoded in the same way -- there will just be more of them, because we're doing text -> bytes -> base64 instead of just turning bytes -> base64.

Where are we seeing an error related to UTF-8?

radazen:
I did some digging and found an answer! There's a bug in the decode step where we're accidentally reading an extra leading byte for the ID (it's actually the last byte of the timestamp getting reused). Depending on what that byte is, the resulting ID might not contain valid UTF-8 characters.

Changing the return statement in `parseTime` from this: `return t, len(timeSlice[1 : timeLen+1]), nil`

to this: `return t, len(timeSlice[1 : timeLen+1]) + 1, nil`

fixes the issue.

benny-conn:
Thank you ezra for finding that!

PR #598, opened 2022-10-25
only gallery users field

benny-conn:
Changes:

- **Added** `onlyGalleryUsers` field to the `owners` query for the `Community` graphQL type. This field will filter out non-gallery users and is an optional field


radazen:
Idea: update the existing query to conditionally restrict the result set to Gallery users only:

```
-- name: GetOwnersByContractIdBatchPaginate :batchmany
SELECT DISTINCT ON (result.id) result.* FROM (SELECT users.* FROM users, tokens
    WHERE tokens.contract = $1 AND tokens.owner_user_id = users.id
    AND (NOT @gallery_users_only OR users.universal = false) <-- this line!
    AND tokens.deleted = false AND users.deleted = false
    AND (users.universal,users.created_at,users.id) < (@cur_before_universal, @cur_before_time::timestamptz, @cur_before_id)
    AND (users.universal,users.created_at,users.id) > (@cur_after_universal, @cur_after_time::timestamptz, @cur_after_id)
    ORDER BY CASE WHEN @paging_forward::bool THEN (users.universal,users.created_at,users.id) END ASC,
        CASE WHEN NOT @paging_forward::bool THEN (users.universal,users.created_at,users.id) END DESC) AS result LIMIT $2;
```

That way, you don't have to change _any_ of your pagination stuff. All you need to do is pass the additional parameter in your sqlc query, and everything else just works (even the original boolTimeID cursor -- it's totally fine for the bool component to always be false when `@gallery_users_only` is set).

Same thing for counting owners: instead of adding a new query to count only the Gallery users, just update the existing query to conditionally filter out non-Gallery users.

publicapi/contract.go
```diff
@@ -180,25 +187,30 @@ func (api ContractAPI) GetCommunityOwnersByContractAddress(ctx context.Context,
 	}
 
 	countFunc := func() (int, error) {
-		total, err := api.queries.CountOwnersByContractId(ctx, contract.ID)
+		var total int64
+		var err error
+		if onlyGalleryUsers != nil && *onlyGalleryUsers {
+			total, err = api.queries.CountGalleryOwnersByContractId(ctx, contract.ID)
+		} else {
+			total, err = api.queries.CountOwnersByContractId(ctx, contract.ID)
```
radazen, line 195: I might be missing it, but I don't see this query defined in the `query.sql` file (though I do see a generated version in `query.sql.go`, which makes me wonder if running `go generate ./...` might remove the generated version).

Thoughts on parameterizing this as well? It'd complement the other query nicely.

```
-- name: CountOwnersByContractId :one
SELECT count(DISTINCT users.id) FROM users, tokens
    WHERE tokens.contract = $1 AND tokens.owner_user_id = users.id
    AND tokens.deleted = false AND users.deleted = false
    AND (NOT @gallery_users_only::bool OR users.universal = false);
```

benny-conn: Totally! Just added this to my most recent commit as well and now everything is good to review!

radazen (approved):
🔥 👏 🚀

PR #603, opened 2022-10-25
get all users up front

benny-conn:
Changes:

- **Changed** create users step in refresh tokens in collection step to get all users for wallets up front to avoid expensive queries


radazen (approved):
👍 👍

PR #604, opened 2022-10-25
Comment and Admire Notification Fixups

jarrel-b:
Noticed while getting the Feed Caption branch up to date that foreign key errors popped up as I was commenting and admiring stuff. This also removes the `ResourceTypeFeedEvent` type since it doesn't seem to be needed anymore.

### DIFF
--- a/service/notifications/notifications.go
+++ b/service/notifications/notifications.go
@@ -131,7 +131,6 @@ type groupedNotificationHandler struct {
 }
 
 func (h groupedNotificationHandler) Handle(ctx context.Context, notif db.Notification) error {
-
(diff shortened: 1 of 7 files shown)


service/notifications/notifications.go
```diff
@@ -329,7 +328,7 @@ func updateAndPublishNotif(ctx context.Context, notif db.Notification, mostRecen
 	if err != nil {
 		return fmt.Errorf("error updating notification: %w", err)
 	}
-	updatedNotif, err := queries.GetNotificationByID(ctx, mostRecentNotif.ID)
+	updatedNotif, err := queries.GetNotificationByID(ctx, notif.ID)
```
jarrel-b, line 331: I think this should be `notif` and not `mostRecentNotif` since `mostRecentNotif` isn't stored at this point?

Robinnnnn (approved):

PR #606, opened 2022-10-26
use index for wallets

benny-conn:
Changes:

- **Changed** user by wallet ID query to use new index


service/persist/postgres/user.go
```diff
@@ -59,7 +59,7 @@ func NewUserRepository(db *sql.DB) *UserRepository {
 	getByIDsStmt, err := db.PrepareContext(ctx, `SELECT ID,DELETED,VERSION,USERNAME,USERNAME_IDEMPOTENT,BIO,TRAITS,WALLETS,UNIVERSAL,CREATED_AT,LAST_UPDATED FROM users WHERE ID = ANY($1) AND DELETED = false;`)
 	checkNoErr(err)
 
-	getByWalletIDStmt, err := db.PrepareContext(ctx, `SELECT ID,DELETED,VERSION,USERNAME,USERNAME_IDEMPOTENT,WALLETS,BIO,TRAITS,UNIVERSAL,CREATED_AT,LAST_UPDATED FROM users WHERE $1 = ANY(WALLETS) AND DELETED = false;`)
+	getByWalletIDStmt, err := db.PrepareContext(ctx, `SELECT ID,DELETED,VERSION,USERNAME,USERNAME_IDEMPOTENT,WALLETS,BIO,TRAITS,UNIVERSAL,CREATED_AT,LAST_UPDATED FROM users WHERE ARRAY[$1] <@ WALLETS AND DELETED = false;`)
```
radazen, line 62: Tested locally; I think this actually needs to be `ARRAY[$1]::varchar[]` or else it throws an error!

benny-conn: wonderful! thank you for that quick test!

radazen (approved):
🚀

Issue #609, opened 2022-10-27
"make g-docker" fail


Geapefurit:
The following error appear after runing **"make g-docker"** command.

**docker build -t bcgallery/gallery-postgres:circle -f docker/postgres/circle/DOCKERFILE .**
Sending build context to Docker daemon  3.751MB
Step 1/5 : FROM circleci/postgres
 ---> 75c1bebe6967
Step 2/5 : ENV POSTGRES_HOST_AUTH_METHOD trust
 ---> Using cache
 ---> 9e9ac65f7acb
Step 3/5 : ENV POSTGRES_USER postgres
 ---> Using cache
 ---> 290a35194e99
Step 4/5 : ENV POSTGRES_DB postgres
 ---> Using cache
 ---> abe5df98a77f
**Step 5/5 : COPY db/migrations/*up.sql /docker-entrypoint-initdb.d/**
COPY failed: no source files were specified
make: *** [Makefile:43: g-docker] Error 1

Geapefurit:
The error just appears only in linux.

jarrel-b:
Thanks for reporting! This isn't needed to run the project so removing it from the README.

Issue #608, opened 2022-10-27
"Go get" fails due to willf/bitset 


Geapefurit:
When I run "go get -u -d ./...",the following error occus.

go: github.com/willf/bitset@v1.3.3: parsing go.mod:
        module declares its path as: github.com/bits-and-blooms/bitset
                but was required as: github.com/willf/bitset

I have noticed the same problem with other project,link: https://github.com/blugelabs/bluge/issues/64

jarrel-b:
Good catch! As a workaround you can run `go get -u=patch -d ./...` so that only the patched version is installed

PR #611, opened 2022-10-27
Remove circle image and update README

jarrel-b:
Addresses https://github.com/gallery-so/go-gallery/issues/609 and https://github.com/gallery-so/go-gallery/issues/608

* Removes the old circleci postgres image since we create a new instance per test suite
* Removes some inaccurate README stuff

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,14 +212,6 @@ Skip longer running tests with the `-short` flag:
 go test -short
 ```
 
-### Integration Testing
-
-Run integration tests against ethereum mainnet:
-
-```bash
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -11,7 +11,7 @@
 ```bash
 $ git clone git@github.com:gallery-so/go-gallery.git
 $ cd go-gallery
-$ go get -u -d ./...
+$ go get -u=patch -d ./...
```
jarrel-b, line 14: This only install patched versions instead of minor + patched versions

radazen (approved):
LGTM!

PR #612, opened 2022-10-27
Add NOT NULL constraint to tokens.owned_by_wallets

radazen:
`tokens.owned_by_wallets` should never be null, and bad things happen when it _is_ null (like tokens not showing up in collections). This migration adds a `NOT NULL` constraint to the column.

With a constraint in place, we'll have an error to track down whenever we attempt to null one of these columns, instead of allowing it to happen silently (and then noticing later that some tokens are broken).


radazen:
I applied this migration on dev already, and it seems to be working fine. We do have to hard delete ~50 soft-deleted tokens with NULL `owned_by_wallets` columns before adding the constraint, but that's a small price to pay.

jarrel-b (approved):
🚢

PR #613, opened 2022-10-28
Fix duplicate entries in community owners

radazen:
## What's new?
- Rearranged the community owners pagination query to run `distinct` _before_ paginating. Previously, we'd get a page of results and then run `distinct` on it, ensuring that users were unique within a page of results but not within the result set as a whole. By rearranging the query, we're now specifying that we want a unique set of users, and _then_ we page through that unique set.

- Added an index on the tokens table to speed up these queries. I tried adding an index on `users(universal, created_at, id)` too, but it didn't get used, so I didn't add it to this PR.


db/gen/coredb/batch.go
```diff
@@ -1365,6 +1368,12 @@ type GetOwnersByContractIdBatchPaginateParams struct {
 	PagingForward      bool
 }
 
+// Note: sqlc has trouble recognizing that the output of the "select distinct" subquery below will
```
radazen, line 1371: So this is actually pretty annoying. We had discussed leaving comments in sqlc queries before, and I said "yeah, but it's annoying because the comments get included in the query itself!"

I was wrong about that -- the comments get included as documentation on the generated function. This is annoying, since the query function we'll call from our code now has my query comment as its documentation. Not sure what the best approach is here, but the comment is helpful to have in the query.sql file.

jarrel-b (approved):

PR #614, opened 2022-10-28
Group collection feed events

jarrel-b:
#### New feed event type
Collection events (adding tokens, adding a description, creating a collection) may now get grouped into an update collection feed event to avoid things like this:

image

```
type CollectionUpdatedFeedEventData implements FeedEventData @goEmbedHelper {
    eventTime: Time
    owner: GalleryUser @goField(forceResolver: true)
    action: Action
    collection: Collection @goField(forceResolver: true)
    newCollectorsNote: String
    newTokens: [CollectionToken] @goField(forceResolver: true)
    isNewCollection: Boolean
}
```

#### Some feed refactoring
* The feed can now handle groupings of arbitrary actions and merges events based on some simple rules
* Renames a few feed functions to better distinguish between an operation that works on a raw `Event` vs a `FeedEvent`
* Renames methods in `EventRepository` and `FeedEventRepository` to be a bit more intuitive

### Follow ups
* [ ] Frontend to integrate new event type


db/queries/core/query.sql
```diff
@@ -340,26 +346,31 @@ SELECT * FROM feed_events WHERE id = $1 AND deleted = false;
...
-
--- name: GetLastFeedEvent :one
-SELECT * FROM feed_events
-    WHERE owner_id = $1 AND action = $2 AND event_time < $3 AND deleted = false
-    ORDER BY event_time DESC
-    LIMIT 1;
+-- name: GetLastFeedEventForUser :one
+select * from feed_events where deleted = false
+    and owner_id = $1
+    and action = any(@actions::varchar[])
```
jarrel-b, line 352: Is there a way to get sqlc to generate the correct types? Ended up just creating a helper function to cast from the persist type to the postgres type

radazen: Something that frequently works (but not always, depending on how the query is set up):
- Create a new type for []Action: `type ActionList []Action`
- Set up a wildcard entry in the sqlc.yaml config for `*.actions` -> `ActionList`

Wildcard entries seem to apply to named parameters where possible, and the type alias trick lets you map parameters to slices where sqlc otherwise wouldn't allow it.

jarrel-b: Worked like a charm, thanks @radazen !

feed/merge.go
```diff
@@ -0,0 +1,88 @@
...
+			action = persist.ActionCollectionUpdated
+		}
+
+		// Not every event has tokens attached to it, so we check that the event is relevant first.
+		collectionTokenIDs := c.event.Data.CollectionTokenIDs
+		if other.Action == persist.ActionCollectionCreated || other.Action == persist.ActionTokensAddedToCollection {
+			collectionTokenIDs = other.Data.CollectionTokenIDs
+		}
+
+		// Not every event has a collector's note attached to it, so we check that the event is relevant first.
```
Robinnnnn, line 66: these comments are so nice 😍

Robinnnnn (approved):
this is incredible, jarrel!! 🔥🔥🔥 

do you mind updating the excalidraw to reflect the latest architecture? wanna make sure we have clear documentation of all the moving pieces!

radazen (approved):
Awesome. Looking forward to seeing this on the feed!

Also, the updated method names gave me a better sense of what was happening, so thanks for those changes too!

PR #619, opened 2022-11-01
tls insecure verify for certain projects [GAL-579]

benny-conn:
Changes:

- **Changed** the default http client in the RPC package to use the tls config param `InsecureSkipVerify` so that it can connect to http links with unvalidated certificates. This is for links like https://vibes.art/vibes/jpg/8095.jpg that are valid when opened in the browser but invalid when using a default `http.Get` request. The problem here, is this insecure parameter is supposedly just for testing, so I am inclined not even to merge this PR because I feel like they should just get valid certificates and it is not our fault that it is not loading. But this is the solution to the bug of these NFTs not appearing. 
 
I tried finding ways around this bug other ways as well such as detecting the cert error and trying to just use the raw URL in those cases but that results in an unknown media type since we need to get request the URL to even predict what type of media it will be.

### DIFF
--- a/indexer/nfts.go
+++ b/indexer/nfts.go
@@ -829,14 +829,8 @@ func getUpdateForToken(pCtx context.Context, uniqueHandlers uniqueMetadatas, tok
 		newMetadata = md
 	}
 
-	name, ok := util.GetValueFromMap(newMetadata, "name", util.DefaultSearchDepth).(string)
-	if !ok {
-		name = ""
-	}
-	description, ok := util.GetValueFromMap(newMetadata, "description", util.DefaultSearchDepth).(string)
-	if !ok {
-		description = ""
-	}
+	name, description := media.FindNameAndDescription(pCtx, newMetadata)
+
 	image, animation := media.KeywordsForChain(chain, imageKeywords, animationKeywords)
 
 	newMedia, err := media.MakePreviewsForMetadata(pCtx, newMetadata, persist.Address(contractAddress.String()), tokenID, newURI, chain, ipfsClient, arweaveClient, storageClient, tokenBucket, image, animation)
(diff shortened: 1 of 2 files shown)


indexer/nfts.go
```diff
@@ -829,14 +829,8 @@ func getUpdateForToken(pCtx context.Context, uniqueHandlers uniqueMetadatas, tok
...
-	name, ok := util.GetValueFromMap(newMetadata, "name", util.DefaultSearchDepth).(string)
-	if !ok {
-		name = ""
-	}
-	description, ok := util.GetValueFromMap(newMetadata, "description", util.DefaultSearchDepth).(string)
-	if !ok {
-		description = ""
-	}
+	name, description := media.FindNameAndDescription(pCtx, newMetadata)
+
```
benny-conn, line 833: This change is not related to this PR, just saw this and wanted to get it fixed since that `media.FindNameAndDescription(pCtx, newMetadata)` does exactly what the lines deleted does

radazen:
Hmm. Yeah, strongly agree with not merging as-is. If we're going to go this route, we should find a way to enable this sort of behavior on a case-by-case basis, and manually allow it for projects that are using self-signed certificates. Otherwise we're sacrificing security best practices everywhere to the benefit of a small number of projects. Let's discuss during GM tomorrow?

radazen:
I tried to reproduce the issue, but this works on my machine without any errors:

`response, err := http.Get("https://vibes.art/vibes/jpg/8095.jpg")`

Is there a different URL I should try?

Robinnnnn:
- don't expect self-signed / insecure certificates to fly
- figure out why there are discrepancies among different machines. loads on a web browser but not in the go runtime

PR #621, opened 2022-11-01
add address universal delete

benny-conn:
Changes:

- **Changed** authenticate step to remove universal users so that in any place where a wallet might authenticate with a nonce, we prepare ahead of time by removing the universal user

### DIFF
--- a/service/auth/auth.go
+++ b/service/auth/auth.go
@@ -194,17 +194,9 @@ func (e NonceAuthenticator) GetDescription() string {
 
 func (e NonceAuthenticator) Authenticate(pCtx context.Context) (*AuthResult, error) {
 	asChainAddress := e.ChainPubKey.ToChainAddress()
-	nonce, userID, universal, _ := GetUserWithNonce(pCtx, asChainAddress, e.UserRepo, e.NonceRepo, e.WalletRepo)
+	nonce, userID, _ := GetUserWithNonce(pCtx, asChainAddress, e.UserRepo, e.NonceRepo, e.WalletRepo)
 	if nonce == "" {
 		return nil, ErrNonceNotFound{ChainAddress: asChainAddress}
-	}
-
-	if universal {
-		err := e.UserRepo.Delete(pCtx, userID)
-		if err != nil {
-			return nil, err
-		}
-		userID = ""
 	}
 
 	if e.WalletType != persist.WalletTypeEOA {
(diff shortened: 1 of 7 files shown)


service/user/user.go
```diff
@@ -187,55 +200,6 @@ func AddWalletToUser(pCtx context.Context, pUserID persist.DBID, pChainAddress p
 	return nil
 }
 
-// AddAddressToUserToken adds a single address to a user in the DB because a signature needs to be provided and validated per address
-func AddAddressToUserToken(pCtx context.Context, pUserID persist.DBID, pChainAddress persist.ChainAddress, pWalletType persist.WalletType, addressAuth auth.Authenticator,
-	userRepo persist.UserRepository, tokenRepo persist.TokenGalleryRepository, contractRepo persist.ContractRepository, ethClient *ethclient.Client, ipfsClient *shell.Shell, arweaveClient *goar.Client, stg *storage.Client) error {
```
benny-conn, line 192: This code is not used anywhere, got rid of it

radazen:
Oops, also meant to add: I _think_ this has the same net effect as removing the user in the authenticator, but is a little more explicit about what circumstances we delete under (turning the address into a real user or adding it as a wallet).

service/auth/auth.go
```diff
@@ -194,11 +194,19 @@ func (e NonceAuthenticator) GetDescription() string {
...
 		return nil, ErrNonceNotFound{ChainAddress: asChainAddress}
 	}
 
+	if universal {
+		err := e.UserRepo.Delete(pCtx, userID)
+		if err != nil {
+			return nil, err
+		}
+		userID = ""
+	}
```
Robinnnnn, line 208: what if this step succeeds, but a later step fails? i.e. the universal user gets deleted _and_ the authentication step fails? probably not a huge deal to delete a universal user, but something to keep in mind

benny-conn:
Moved the delete step to the create wallet tx like @radazen suggested, still had to include some login in the authenticator because otherwise we will not get to the DB step because there are checks in both the create user and add wallet functions for whether a user exists after authentication. I just made it so the authenticator only returns a user ID for non-universal users given that it should not be possible to authenticate as a universal user anyway.

Robinnnnn (approved):

service/auth/auth.go
```diff
@@ -325,7 +325,7 @@ func GetUserWithNonce(pCtx context.Context, pChainAddress persist.ChainAddress,
 	if err != nil {
 		return nonceValue, userID, err
 	}
-	if user.ID != "" {
+	if user.ID != "" && !user.Universal.Bool() {
```
radazen, line 328: It feels a little deceptive not to return a user ID at all if it's a universal user. What if we return a `*persist.User` instead of a userID, and then it can be nil if there isn't a user, and otherwise we return the actual user that we looked up? That way, callers can check fields of the user to determine how they want to proceed (in this case, by checking `user.Universal`).

benny-conn: Sounds good to me!

Robinnnnn (approved):

PR #622, opened 2022-11-01
name and description nets

benny-conn:
Changes:

- **Added** another check at the indexer level for tokens with valid media and metadata but without a name or description so that those fields can get updated even when the media is all good
- **Changed** some naming as well as started using a consistent shorthand function for getting metadata fields from the metadata `media.FindNameAndDescription`


Robinnnnn (approved):
amazingggg thank you ser!

PR #623, opened 2022-11-01
Benny/tez pagination fix

benny-conn:
Changes:

- **Added** extra check to make sure max limit is set before using it in tezos pagination


jarrel-b (approved):
🚢

PR #627, opened 2022-11-02
Fix owned_by_wallets null constraint error

jarrel-b:
**Changes**
Tokens returned by the OpenSea provider would occasionally return tokens with no owner. The fix is to use the address returned by the OpenSea API versus using the input `address` as the owner.

Tested by refreshing a few communities and verifying that the `owned_by_wallets` field is not null and also owned by the expected wallet.

### DIFF
--- a/service/multichain/opensea/opensea.go
+++ b/service/multichain/opensea/opensea.go
@@ -162,7 +162,7 @@ func (p *Provider) GetTokensByWalletAddress(ctx context.Context, address persist
 		FetchAssets(ctx, assetsChan, persist.EthereumAddress(address.String()), "", "", "", 0, nil)
 	}()
 
-	return assetsToTokens(ctx, address, assetsChan, p.ethClient)
+	return assetsToTokens(ctx, assetsChan, p.ethClient)
 }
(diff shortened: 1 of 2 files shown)


service/multichain/opensea/opensea.go
```diff
@@ -191,8 +190,7 @@ func (p *Provider) GetTokensByTokenIdentifiers(ctx context.Context, ti multichai
 		defer close(assetsChan)
 		FetchAssets(ctx, assetsChan, "", persist.EthereumAddress(ti.ContractAddress), TokenID(ti.TokenID.Base10String()), "", 0, nil)
 	}()
-	// TODO: Fill in this address or change something else
-	tokens, contracts, err := assetsToTokens(ctx, "", assetsChan, p.ethClient)
+	tokens, contracts, err := assetsToTokens(ctx, assetsChan, p.ethClient)
```
radazen, line 193: Seems like we've arrived at the "something else," huh? 😅

radazen (approved):
Looks good to me! Thanks for tracking this down!

PR #630, opened 2022-11-02
remove tokens before update for contract

benny-conn:
Changes:

- **Added** some new `token_gallery` repository functions for bulk upserting tokens and deleting old tokens for a contract
- **Changed** multichain owners refresh to use the new repository functions to ensure that any token from a given contract that was not included in the update (in other words, a token that is no longer owned by the user it should be owned by given that the tokens in the update are all of the tokens for that contract) gets deleted

Considerations:

- Currently this will delete tokens that are owned by universal users and non-universal users, meaning if a user has a token from a contract, puts it in their collection, then sells the token, the next time that community gets refreshed the token will disappear from their collection just the same as if they synced their tokens. This is expected behavior, but to be on the safer side, we could just remove universal user tokens because we can guarantee that it would not affect a user's gallery.
- I experimented with a few different solutions here but the primary issue with solutions involving single transactions was that the upsert function is recursive. My solution was to ensure that the final run of the upsert (after all recursive upserts are done) would delete all tokens from that contract before any upserts were done from that contract. It might look something like this:

Upserting 1000 tokens:

1. Mark down start time as `now`
2. Upsert 500 tokens after we have marked down `now`, their `last_updated` will be greater than or equal to `now`
3. Upsert the final 500 tokens, their `last_updated` will also be greater than or equal to `now`
4. Delete tokens from contract before `now`

This still leaves open the possibility of something happening the short window between all of the upserts and `now` but until we start using the unnest method of upserting we won't be able to fit this all into one transaction to use something more guarunteed like `now()` at the database level


graphql/dataloader/dataloaders.go
```diff
@@ -197,7 +196,7 @@ func NewLoaders(ctx context.Context, q *db.Queries, disableCaching bool) *Loader
 
 	loaders.WalletByChainAddress = NewWalletLoaderByChainAddress(defaults, loadWalletByChainAddress(q), WalletLoaderByChainAddressCacheSubscriptions{
 		AutoCacheWithKey: func(wallet db.Wallet) persist.ChainAddress {
-			return persist.NewChainAddress(wallet.Address, persist.Chain(wallet.Chain.Int32))
+			return persist.NewChainAddress(wallet.Address, wallet.Chain)
```
benny-conn, line 199: All changes involving the chain was because I added a wildcard sqlc variable that will turn chains into our chain type, just as we do with pretty much every other type we have.

radazen (commented):
Looking good, Benny! I left some comments for particular lines, but I think we can also improve the logical flow of things overall. Some thoughts:

**Transactions:** I think we can actually get away without making anything transactional (shoutout to @jarrel-b). Consider the case where we have 10000 broken into chunks of 1000. If those updates all set their time to whatever `now()` is when they run, and the final deletion step only removes tokens upserted before the first chunk's update time, I think we'll avoid concurrency issues entirely. This update must finish upserting all tokens before it deletes anything, and even if we allowed a second sync to run concurrently on this same contract, it seems to me like the requirements of (a) finishing all upserts before deleting, and (b) only deleting things before the first upsert's start time) will keep the results correct.

Also notable: I think we don't need to worry about rolling back if we fail part way through an update. If we upsert a few chunks and then error out and never make it to the deletion step, at worst we've just added more valid token owners without removing invalid ones, and that's probably fine for our purposes.

**Timestamps:** To make this work without transactions, I think we want to use the Postgres DB's `now()` function instead of using the go runtime's `time.Now()` to make sure that even concurrent callers have a consistent view of the current time. We could have the upsert query return `now()`, which AFAIK should be the same timestamp given to all upserts in that query, since `now()` is actually the time the transaction started (and every Postgres query is its own transaction if it isn't included in one manually!).

**New column:** I think we also need a new `last_synced` column on the tokens table to make this work. The `last_updated` column gets updated for all sorts of reasons, and a token could avoid deletion if some unrelated update happened to bump its `last_updated` time prior to the deletion step.

Thinking out loud: I feel like this entire approach (progressive upserts, final delete) should eventually apply to our regular "sync my wallets" token upserts, too. The goal is the same: upsert whatever the provider says they have, then delete anything that wasn't inserted. Assuming we do use this mechanism for syncing, can we use a single `last_synced` column for both contract-based syncs _and_ user-based syncs? Or should it be `last_contract_sync` and `last_wallet_sync`?

Robinnnnn:
@radazen mind taking another crack at this whenever you're next on? thanks!

PR #631, opened 2022-11-02
Remove isNewCollection from schema

jarrel-b:
Realized that we can simplify the grouping logic a bit. If a create event exists, then we can just classify the grouped event as a CollectionCreatedEvent instead of a CollectionUpdatedEvent.

In other words, the CollectionUpdatedEvent is only shown when you add tokens AND a new note to an existing collection.


Robinnnnn (approved):
🔥

PR #633, opened 2022-11-03
GAL-517 Add Verification Email template

kaitoo1:
This PR adds the template for the Verification Email.

It uses Handlebars, a templating language supported by Sendgrid.
[docs](https://www.google.com/search?client=safari&rls=en&q=sendgrid+handlebars&ie=UTF-8&oe=UTF-8&safari_group=7)

There are currently 3 variables in the template which the backend can populate:
- username
- verificationUrl
- unsubscribeUrl

I haven't tested email client compatibility yet, will do so using a tool like Litmus before we launch. The goal of this PR is so that we have an email template to test with that's close to the final thing.

What the html looks like rendered in a browser:
Screen Shot 2022-11-03 at 15 37 55

### DIFF
--- a/emails/templates/verificationEmail.html
+++ b/emails/templates/verificationEmail.html
@@ -1,0 +1,103 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
+  <html
+    dir="ltr"
+    xmlns="http://www.w3.org/1999/xhtml"
+    xmlns:v="urn:schemas-microsoft-com:vml"
+    xmlns:o="urn:schemas-microsoft-com:office:office">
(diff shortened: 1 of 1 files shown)


kaitoo1:
@benny-conn I made a folder called `templates` to hold all of our email template files. feel free to rename or move it where it makes sense to you!

Robinnnnn (approved):
LFG

benny-conn (commented):
This looks fantastic! Currently the way I set up the backend to use templates, however, is to have the templates loaded from Sendgrid directly. Would it be alright if I upload this HTML to Sendgrid and we can store all of the templates there instead of in this repo? That also might make it easier if a non-engineer were to edit it like Jess, Mike, or Kyt.

kaitoo1:
@benny-conn yeah absolutely! do you know if there's any type of versioning on Sendgrid?

PR #634, opened 2022-11-03
Don't include hidden or deleted collections for displayed contracts

radazen:
Quick fix to filter badges from hidden and deleted collections out of the "displayed contracts" query. I also changed the `JOIN` to an `INNER JOIN` since we don't need the row if we can't find the collection, and made everything lowercase cause that's a cool new thing we're doing. 😎


Robinnnnn (approved):

radazen:
Forgot I had this open! I'll modify it to include hidden collections as previously discussed.

PR #635, opened 2022-11-03
new certs

benny-conn:
Changes:

- **Added** extra root certificates to the`_deploy` directory
- **Added** dockerfile changes to copy them and use them in the required services
- **Changed** `rpc` `http.Client` init to read from the `_deploy` directory if there to load in those roots for use locally.


service/rpc/rpc.go
```diff
@@ -192,9 +196,39 @@ func newClientForIPFS(projectID, projectSecret string, continueOnly bool) *http.
...
 // newHTTPClientForRPC returns an http.Client configured with default settings intended for RPC calls.
 func newHTTPClientForRPC(continueTrace bool, spanOptions ...sentry.SpanOption) *http.Client {
+	// get x509 cert pool
+	pool, err := x509.SystemCertPool()
+	if err != nil {
+		panic(err)
+	}
+
+	// walk every file in the tls directory and add them to the cert pool
+	filepath.WalkDir("_deploy/root-certs", func(path string, d fs.DirEntry, err error) error {
```
benny-conn, line 206: will fail silently if this directory does not exist. for production deployments, the roots should have been added to the system cert pool through the dockerfile already so this code will do nothing

Robinnnnn (approved):
🔥🔥🔥

radazen:
Awesome! A few questions:

- Have you confirmed that https links aside from ones that use the Sectigo cert still work?

- Could we document where the cert came from, why we're using it, etc? Right now it's just `sectigo.crt` in the root-certs directory, and it'd be awesome to know which Sectigo cert it is, where I'd go to download it, etc. Future-us will thank current-us for adding that. Maybe we add a section to the README, or add a readme file for certs elsewhere?

PR #647, opened 2022-11-07
[GAL-538] Speed Up sync

jarrel-b:
Here's my pass at optimizing token syncs. Benchmarking with the user `poppel` who has about 2100 tezos tokens, a fresh sync took  ~8min compared to about ~15min, so about a 40-50% improvement!

**Changes**
* If a token has both an image URL and a video URL, we now process both URLs in parallel vs handling them serially
* Previously, we would use the `cat` IPFS call, and if that failed we try the `HTTP` call. Now we make both calls in parallel and return whichever responds first.
* When predicting the media type, we first try to `HEAD` the resource vs `GET` to prevent downloading the token's content twice
* Checking if an object exists and purging it from imgix is now done asynchronously
* Previously we would write an object, then update it's metadata afterward. Now the storage writer is configured to include the object metadata to save an extra request
* Tokens with the `Video` media type now have their video and thumbnail processed in parallel

**Further optimizations / testing**
* GCP has an experimental gRPC client which might have better upload speed
* The golang client doesn't support batch uploads yet, but this is something we could play around with
* We validate if media exists in cloud storage for a token before replying with the storage URL. I think this check isn't needed in the context of a sync because we know that the media was just written by the sync


service/media/media.go
```diff
@@ -519,114 +521,52 @@ func cacheRawAnimationMedia(ctx context.Context, reader io.Reader, bucket, fileN
...
-		return err
-	}
+		logger.For(ctx).Infof("thumbnailing %s", videoURL)
+		if err := thumbnailVideoToWriter(videoURL, sw); err != nil {
+			errCh <- fmt.Errorf("could not write to bucket %s for %s: %s", bucket, fileName, err)
+		}
 
-	logger.For(ctx).Infof("storage copy took %s", time.Since(timeBeforeCopy))
+		if err := sw.Close(); err != nil {
+			errCh <- err
```
benny-conn, line 546: Same here?

service/rpc/rpc.go
```diff
@@ -469,7 +470,7 @@ func GetDataFromURI(ctx context.Context, turi persist.TokenURI, ipfsClient *shel
 }
 
 // GetDataFromURIAsReader calls URI and returns the data as an unread reader with the headers pre-read
-func GetDataFromURIAsReader(ctx context.Context, turi persist.TokenURI, ipfsClient *shell.Shell, arweaveClient *goar.Client) (util.FileHeaderReader, error) {
+func GetDataFromURIAsReader(ctx context.Context, turi persist.TokenURI, ipfsClient *shell.Shell, arweaveClient *goar.Client) (*util.FileHeaderReader, error) {
```
benny-conn, line 473: This is huge! This could have led to some weird bugs later if it wasn't a pointer, great catch!

service/media/media.go
```diff
@@ -519,114 +521,52 @@ func cacheRawAnimationMedia(ctx context.Context, reader io.Reader, bucket, fileN
...
-		return fmt.Errorf("could not write to bucket %s for %s: %s", bucket, fileName, err)
-	}
+		sw := newObjectWriter(ctx, client, bucket, fileName, "image/jpeg")
 
-	if err := sw.Close(); err != nil {
-		return err
-	}
+		logger.For(ctx).Infof("thumbnailing %s", videoURL)
+		if err := thumbnailVideoToWriter(videoURL, sw); err != nil {
+			errCh <- fmt.Errorf("could not write to bucket %s for %s: %s", bucket, fileName, err)
```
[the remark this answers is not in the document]
jarrel-b: Good catch yes we should

service/media/media.go
jarrel-b: yup, adding!

PR #648, opened 2022-11-07
Use GCP-compatible fields for logging severity and timestamps

radazen:
As it turns out, Google Cloud Logging can extract severity and timestamps from our log entries if we add JSON fields with the required names. This PR adds those two fields to our logrus output when logging on GCP, and also removes some duplication where we had been copy/pasting the same logger initialization code in every service.

@jarrel-b One thing I'm wondering: what's the `configureRootContext` / `NewContextWithLogger` stuff used for? From what I can tell, the indexer initializes a JSONFormatter-style logger, but then it ends up using the `logrus.New()` logger created in `configureRootContext` everywhere, which means it misses out on the benefits of JSON logging on GCP (namely: being able to filter logs or add summary headers using fields in the JSON payload). Similarly, I expect it wouldn't use the GCPFormatter added in this PR, and thus wouldn't benefit from the added severity and timestamp info. The same is probably true of the other services that use `configureRootContext`, though I didn't look closely at those.


jarrel-b:
The `configureRootContext` / `NewContextWithLogger` stuff is mainly to add the Sentry logging hook to the Indexer (when it logs stuff it gets reported to Sentry as an error or a breadcrumb). The other services that use `configureRootContext` should probably just use the GCP logger.

I can figure out the Indexer logging in a bit so its standard with all our other stuff

radazen:
> The `configureRootContext` / `NewContextWithLogger` stuff is mainly to add the Sentry logging hook to the Indexer (when it logs stuff it gets reported to Sentry as an error or a breadcrumb). The other services that use `configureRootContext` should probably just use the GCP logger.

Makes sense!

> I can figure out the Indexer logging in a bit so its standard with all our other stuff

That'd be awesome. Thanks!

jarrel-b (approved):
🚢

Robinnnnn (approved):
😍

PR #650, opened 2022-11-08
update email, notification settings, and resend verification

benny-conn:
Another bite sized chunk of email progress

Changes:

- **Added** mutations for updating email notification settings and resending verification emails
- **Added** update email

### DIFF
--- a/db/queries/core/query.sql
+++ b/db/queries/core/query.sql
@@ -551,5 +551,8 @@ UPDATE users SET email_verified = $2 WHERE id = $1;
 -- name: UpdateUserEmail :exec
 UPDATE users SET email = $2, email_verified = 0 WHERE id = $1;
 
+-- name: UpdateUserEmailUnsubscriptions :exec
+UPDATE users SET email_unsubscriptions = $2 WHERE id = $1;
+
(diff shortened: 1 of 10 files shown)


kaitoo1:
just checking - for resending verification emails, is it correct that only a signed in user can call this mutation for just themselves?

Robinnnnn (approved):

jarrel-b (approved):
🚢 🚢 🚢

db/queries/core/query.sql
```diff
@@ -551,5 +551,8 @@ UPDATE users SET email_verified = $2 WHERE id = $1;
 -- name: UpdateUserEmail :exec
 UPDATE users SET email = $2, email_verified = 0 WHERE id = $1;
 
+-- name: UpdateUserEmailUnsubscriptions :exec
+UPDATE users SET email_unsubscriptions = $2 WHERE id = $1;
```
jarrel-b, line 555: Would we want `deleted` here?

benny-conn: It could be, but also if we update a deleted user that doesn't sound like an issue and it sounds like a sort of impossible situation to get in. I think a lot of our other updates disregard deleted.

PR #653, opened 2022-11-08
Benny/emails initial deploy ready

benny-conn:
Changes:

- **Added** notification email templating 
- **Added** verification status checks to sending automated emails, requiring emails to be sent only to admin emails in dev and local and verified emails in prod
- **Added** a few more banned usernames

For @kaitoo1 and anyone else testing:

To shoot out notifications to admin emails test you need to send a post request the `/send/notifications` endpoint on the emails service that I will deploy to dev once this is merged. You will need to manually update your `email_verified` status to `3` in the DB to get the admin email status.

### Post merge

- [ ] Email services will be auto-deployed by actions when this PR is merged. We have to add the `EMAILS_HOST` to the env of the dev backend and prod backend
- [ ] Open up dev email service to public traffic so we can call it directly


emails/send.go
```diff
@@ -96,17 +114,67 @@ func sendNotificationEmails(queries *coredb.Queries, s *sendgrid.Client) gin.Han
...
+
+			asJSON, err := json.Marshal(data)
+			if err != nil {
+				return err
 			}
+			asMap := make(map[string]interface{})
+			err = json.Unmarshal(asJSON, &asMap)
+			if err != nil {
+				return err
+			}
```
benny-conn, line 164: this is kind of silly, if anyone has a better idea for this let me know haha

emails/send.go
```diff
@@ -119,18 +187,130 @@ func sendNotificationEmails(queries *coredb.Queries, s *sendgrid.Client) gin.Han
...
+				Action: "viewed your gallery",
+			}, nil
+		}
+
+		return notificationEmailDynamicTemplateData{}, fmt.Errorf("no viewer ids")
+	default:
+		return notificationEmailDynamicTemplateData{}, fmt.Errorf("unknown action %s", n.Action)
+	}
+}
+
```
benny-conn, line 296: main part needing review, probably could do some optimization here this is just a first pass at this logic and everything worked when tested, if anything starkly wrong is here or there is an obvious optimization, let me know!

emails/send.go
```diff
@@ -100,9 +118,59 @@ func sendNotificationEmails(queries *coredb.Queries, s *sendgrid.Client) gin.Han
...
+			if err != nil {
+				return err
 			}
+			asMap := make(map[string]interface{})
+			err = json.Unmarshal(asJSON, &asMap)
+			if err != nil {
+				return err
+			}
+
+			logger.For(c).Debugf("sending notifications email to %s with data %+v", u.Email, asMap)
```
Robinnnnn, line 170: we won't send anything if the user doesn't have any notifications right?

emails/send.go
```diff
@@ -124,18 +192,130 @@ func sendNotificationEmails(queries *coredb.Queries, s *sendgrid.Client) gin.Han
 	}
 }
 
+func notifToTemplateData(ctx context.Context, queries *coredb.Queries, n coredb.Notification) (notificationEmailDynamicTemplateData, error) {
```
Robinnnnn, line 195: would love to see tests for these various cases!

emails/send.go
```diff
@@ -124,18 +192,130 @@ func sendNotificationEmails(queries *coredb.Queries, s *sendgrid.Client) gin.Han
...
+
 func runForUsersWithNotificationsOnForEmailType(ctx context.Context, emailType persist.EmailType, queries *coredb.Queries, fn func(u coredb.User) error) error {
 	errGroup := new(errgroup.Group)
 	var lastID persist.DBID
 	var lastCreatedAt time.Time
 	var endTime time.Time = time.Now().Add(24 * time.Hour)
+	requiredStatus := persist.EmailVerificationStatusVerified
+	if isDevEnv() {
+		requiredStatus = persist.EmailVerificationStatusAdmin
+	}
```
Robinnnnn, line 310: is this just an override we can set to pretend that an email address is verified?

Robinnnnn (approved):

emails/send.go
benny-conn: this is an override we can set to have emails get sent to those addresses with this status on the dev and local environments. so we could set for example our emails to admin and not worry about sending thousands of users test emails

emails/send.go
benny-conn: yep!

emails/send.go
```diff
@@ -89,8 +93,22 @@ func sendVerificationEmail(dataloaders *dataloader.Loaders, queries *coredb.Quer
...
+	PreviewText    string       `json:"previewText"`
+}
+type notificationsEmailDynamicTemplateData struct {
+	Notifications []notificationEmailDynamicTemplateData `json:"notifications"`
+	Username      string                                 `json:"username"`
+}
+
 func sendNotificationEmails(queries *coredb.Queries, s *sendgrid.Client) gin.HandlerFunc {
 
+	searchLimit := int32(10)
```
jarrel-b, line 110: Is the searchLimit greater than the resultLimit because we expect some of these notifications to not be usable?

emails/send.go
```diff
@@ -100,9 +118,63 @@ func sendNotificationEmails(queries *coredb.Queries, s *sendgrid.Client) gin.Han
 			m.SetFrom(from)
 			p := mail.NewPersonalization()
 			m.SetTemplateID(viper.GetString("SENDGRID_NOTIFICATIONS_TEMPLATE_ID"))
-			p.DynamicTemplateData = map[string]interface{}{
-				"notifications": []string{"notification 1", "notification 2"},
+			notifs, err := queries.GetRecentUnseenNotifications(c, coredb.GetRecentUnseenNotificationsParams{
```
jarrel-b, line 121: Is this email sent on a schedule or is fired after some criteria are met? If it's on a schedule, wondering if we need something to ensure that notifications don't get re-used? Is the idea that a user will visit the site as a result of the email then "see" those notifications or does sending this email marks those notifications as "seen"?

jarrel-b (approved):

emails/send.go
kaitoo1: > the idea that a user will visit the site as a result of the email then "see" those notifications


yup this is correct! it won't mark those notifications as seen, so those notifications should still show up in the app.

kaitoo1: though we should make sure we limit the query to notifications that happened after the last time the email was sent

so that if I'm not an active user, I don't get an email with notifications from weeks ago

kaitoo1: ^ @benny-conn 
(if this limitation isn't included in the PR, it can be a follow up too - doesn't block this PR)

db/queries/core/query.sql
```diff
@@ -455,6 +455,17 @@ SELECT * FROM notifications WHERE owner_id = $1 AND deleted = false
              CASE WHEN NOT @paging_forward::bool THEN (created_at, id) END DESC
     LIMIT $2;
 
+-- name: GetUserUnseenNotifications :many
+SELECT * FROM notifications WHERE owner_id = $1 AND deleted = false AND seen = false
+    AND (created_at, id) < (@cur_before_time, @cur_before_id)
+    AND (created_at, id) > (@cur_after_time, @cur_after_id)
+    ORDER BY CASE WHEN @paging_forward::bool THEN (created_at, id) END ASC,
```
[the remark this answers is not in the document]
benny-conn: I believe so but I will check!

emails/send.go
```diff
@@ -124,18 +192,130 @@ func sendNotificationEmails(queries *coredb.Queries, s *sendgrid.Client) gin.Han
...
+	case persist.ActionAdmiredFeedEvent:
+		feedEvent, err := queries.GetFeedEventByID(ctx, n.FeedEventID)
+		if err != nil {
+			return notificationEmailDynamicTemplateData{}, fmt.Errorf("failed to get feed event for admire %s: %w", n.FeedEventID, err)
+		}
+		collection, _ := queries.GetCollectionById(ctx, feedEvent.Data.CollectionID)
+		data := notificationEmailDynamicTemplateData{}
+		if collection.ID != "" {
+			data.CollectionID = collection.ID
+			data.CollectionName = collection.Name.String
```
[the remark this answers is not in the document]
benny-conn: Good point!

emails/send.go
benny-conn: Tbh this was mostly for testing because I was running into errors, that was the intention behind it though. I'd say we can make them both the same and expect the notifications to be valid but for now this is fine.

emails/send.go
benny-conn: Will definitely follow up, sounds like it might require a non-trivial amount of work to mark notifications as "email seen"

PR #654, opened 2022-11-09
unsubscriptions [GAL-588]

benny-conn:
Changes:

- **Added** a mutation for unsubscribing from an email type through an auth token singed by the email service so that users can click the unsub link in their emails
- **Added** re-subscription functionality in the email service, no mutation for this just yet
- **Changed** the way settings are updated by moving the update to the email service just like where email verification statuses are set, once again though this functionality along with setting verification status could be set on the backend but thought at the very least to keep the update functionality on the same service for now.

Considerations:

- Sendgrid allows you to include `{{{unsubscribe}}}` in a template and have their built-in unsubscribe page show up that will keep track of unsubscriptions entirely on their end so that as long as we send the group ID of each email type along with the email (which I set up in this PR), the emails will get sent or not sent based on if the user is not subsubscribed. This makes me wonder if we should even be storing `email_unsubscriptions` at all on our end. Here are the pros and cons I see:

**Pros** of storing `email_unsubscriptions`:

- Can greatly reduce the work that the email service needs to perform when sending emails, especially bulk emails. If we know internally without having to call a Sendgrid API what a user is unsubscribed to we can exclude them from the bulk sends and all of the work the email service does for each user to send the email (like prepare all of their notifications for an notifications email).
- Easy access to the information for displaying to a user. We could call the Sendgrid API to tell users if they are subscribed or not in their settings but having this information stored makes it super easy.

**Cons** of storing `email_unsubscriptions`:

- Unsubscription logic is our responsibility. If there is a bug somewhere that sends an email to an unsubscribed user that user would be pretty upset and I feel like there might even be legal implications haha but that is just a feeling. Sendgrid is trustworthy so offloading this to them could save us brain trauma.
- Keeping track of email types and unsubscriptions can get messy when we are adding new types of emails and we have to make sure that they are supported everywhere (backend, graphQL schema, email service). If we offload this to Sendgrid we just have to make sure that when the email is sent the correct email group is passed along and we can get rid of that type handling altogether.
- Right now, the unsubscribe link would use a JWT to allow an un-authed user to unsubscribe but a JWT might be weird for this because if it lasts too long and gets exposed someone could use it to unsubscribe the user if they were to resubscribe. And we would want these JWTs to be long lasting so a user can unsubscribe even a week later. This isn't likely to happen but it's possible. To prevent this we would have to store unsubscribe nonces in the database and track their use. A little more DB logic we can avoid if we offload to Sendgrid.

Right now, we double up unsubscriptions by storing them on our end and also updating Sendgrid unsubscribe groups. We could also just store everything related to unsubs just on our end and not keep any consistency with Sendgrid. The idea behind this consistency would be just to have an added layer of protection against ourselves to ensure we don't spam users with emails they unsubbed to. If you vote for us to just offload all of the logic to Sendgrid, disregard this point entirely.

Personally, I think it sounds best to let Sendgrid this handle this whole unsubscribe system and we just have mutations for adding, removing, and retrieving emails from the Sendgrid unsubscribe lists.


jarrel-b:
I'm a fan of having sendgrid handle the unsubscription stuff so that it becomes the single source of truth for subscriptions, and so we don't have to worry about the complexity of managing consistency between our db and theirs. Especially if it makes our application code simpler and easier to reason about too!

kaitoo1:
agreed with Jarrel! +100 to having a single source of truth . 
looking into how to integrate their preference data into our custom ui to manage user preferences

graphql/resolver/schema.resolvers.helpers.go
```diff
@@ -979,6 +979,20 @@ func updateUserEmailNotificationSettings(ctx context.Context, input model.Update
 
 }
 
+func unsubscribeFromEmailType(ctx context.Context, input model.UnsubscribeFromEmailTypeInput) (*model.UnsubscribeFromEmailTypePayload, error) {
+
+	// TODO this token needs to be added to the template data for the email in a future when the templates are all in place
+
+	if err := emails.UnsubscribeFromEmailTypes(ctx, "", input.Token, []model.EmailUnsubscriptionType{input.Type}); err != nil {
```
jarrel-b, line 986: feel free to ignore: maybe we could create a helper `UnscubscribeFromEmailTypesByJWT` so its clear that userID is an optional param

jarrel-b (approved):

benny-conn:
Alrighty, so we discussed the pros and cons of storing `email_subscriptions` and it looks like the conclusion is it is worth it to keep consistency because:

1. Sendgrid can rate limit us :(
2. Having the knowledge of who is subscribed to emails that we are sending directly from the backend feels right
3. Easier to show user everything and allow them to change it.

That means this PR can definitely be merged. Right now we are using JWTs though for the unsubscribe token that will be used to verify that they clicked a link we sent them, which is not secure enough for this case so I will create a follow up PR for turning this JWT into a token that we store in the DB and keep track of if it is used or not. But for now there is no harm in merging this to unblock everyone else if it is good to go.

graphql/resolver/schema.resolvers.helpers.go
benny-conn: Love this idea!

PR #656, opened 2022-11-09
Fix actor_id foreign key error

jarrel-b:
**Changes**

* Drops the non-nullable constrain from the `actor_id` column of `events` since its possible to have events not initiated by a gallery user (like viewing a gallery when unauthenticated)

* Fixed a bug that popped up after fixing that issue where the wrong notification event was being queried, resulting in no rows returned

* Changed the `caption` go backed typed from our `persist.NullString` to the builtin `database/sql.NullString` (sidenote: I was having issues having our `persist.NullString` type insert `NULL` values into the db, but didn't dig in too deeply)

### DIFF
--- a/sqlc.yaml
+++ b/sqlc.yaml
@@ -145,8 +145,6 @@ sql:
             go_type: "github.com/mikeydub/go-gallery/service/persist.NullString"
           - column: "*.media_url"
             go_type: "github.com/mikeydub/go-gallery/service/persist.NullString"
-          - column: "*.caption"
-            go_type: "github.com/mikeydub/go-gallery/service/persist.NullString"
           - column: "*.chain"
             go_type: "github.com/mikeydub/go-gallery/service/persist.Chain"
           - column: "*.email_type"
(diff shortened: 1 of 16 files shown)


Robinnnnn (approved):

sqlc.yaml
```diff
@@ -100,6 +100,8 @@ sql:
             go_type: "github.com/mikeydub/go-gallery/service/persist.NullInt"
           - column: "events.external_id"
             go_type: "github.com/mikeydub/go-gallery/service/persist.NullString"
+          - column: "events.actor_id"
+            go_type: "database/sql.NullString"
```
benny-conn, line 104: Just to help me wrap my head around this, could I get a quick TLDR why we wouldn't still map this to a `persist.DBID` which has scanning support for null values?

benny-conn (approved):
Looks great! One question just to help me understand.

sqlc.yaml
jarrel-b: Ah interesting! I didn't realize that it supported nulls. I think there may be an issue in that case bc the `ActorID` should be the zero-valued `persist.DBID` but it was still being inserted as an empty string. I'll try to dig in a bit more since it would be great just to use the original type

PR #663, opened 2022-11-10
fan out subscriptions and cleanup

benny-conn:
Changes:

- **Changed** notification subscription logic to create multiple subscriptions and delete them on application termination

The reason why this PR is necessary is because of the discovery that GCP Pub/Sub only sends messages once to a given subscription, meaning if we want to fan out messages to every server, each server needs to have a unique subscription to the topic being pushed to.


service/notifications/notifications.go
```diff
@@ -223,9 +225,21 @@ func (h viewedNotificationHandler) Handle(ctx context.Context, notif db.Notifica
...
-	sub := n.pubSub.Subscription(viper.GetString("PUBSUB_SUB_NEW_NOTIFICATIONS"))
+	sub, err := n.pubSub.CreateSubscription(context.Background(), fmt.Sprintf("new-notifications-%s", persist.GenerateID()), pubsub.SubscriptionConfig{
+		Topic: n.pubSub.Topic(viper.GetString("PUBSUB_TOPIC_NEW_NOTIFICATIONS")),
+	})
+	if err != nil {
+		logger.For(nil).Errorf("error creating updated notifications subscription: %s", err)
+		panic(err)
+	}
+
+	n.pubsubSubscriptions = append(n.pubsubSubscriptions, sub)
```
benny-conn, line 236: this is the bulk of this PR and why we need to add cleanups

jarrel-b:
What's the consequence of not deleting a subscription and leaving it around? It looks like GCP automatically expires subscriptions based on the subscription configuration, and it'd be nice to avoid the complexity with the cleanup. Wondering if it makes sense to let GCP manage the cleanup for us?

benny-conn:
> What's the consequence of not deleting a subscription and leaving it around? It looks like GCP automatically expires subscriptions based on the subscription configuration, and it'd be nice to avoid the complexity with the cleanup. Wondering if it makes sense to let GCP manage the cleanup for us?

You are right, they are deleted after 31 days. There is a max of 10k subs to a topic and I don't think the server will stop or start more than 10k times in a month so very good point!

PR #668, opened 2022-11-11
unsub adjustments/bug fixes

benny-conn:
Changes:

- **Added** a new emails endpoint just for updating unsubs by user ID
- **Changed** unseen count to use DB (bug fix)
- **Changed** window for grouping views vs other grouped events

### DIFF
--- a/emails/subscriptions.go
+++ b/emails/subscriptions.go
@@ -10,6 +10,7 @@ import (
 	"github.com/mikeydub/go-gallery/db/gen/coredb"
 	"github.com/mikeydub/go-gallery/graphql/model"
 	"github.com/mikeydub/go-gallery/service/auth"
+	"github.com/mikeydub/go-gallery/service/logger"
 	"github.com/mikeydub/go-gallery/service/persist"
 	"github.com/mikeydub/go-gallery/util"
 	"github.com/sendgrid/sendgrid-go"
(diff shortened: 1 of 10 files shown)


kaitoo1:
thank you bennyyyyy

emails/subscriptions.go
```diff
@@ -10,46 +10,111 @@ import (
...
-	EmailTypes []model.EmailUnsubscriptionType `json:"email_types" binding:"required"`
+var emailTypes = []model.EmailUnsubscriptionType{model.EmailUnsubscriptionTypeAll, model.EmailUnsubscriptionTypeNotifications}
+
+type UpdateSubscriptionsTypeInput struct {
+	UserID persist.DBID                 `json:"user_id,required"`
+	Unsubs persist.EmailUnsubscriptions `json:"unsubscriptions" binding:"required"`
+}
+
+type UnsubInput struct {
+	JWT    string                          `json:"jwt,omiterequiredmpty"`
```
jarrel-b, line 29: Is this struct tag right: `omiterequiredmpty`?

emails/subscriptions.go
```diff
@@ -10,46 +10,111 @@ import (
...
+			ID:                   input.UserID,
+			EmailUnsubscriptions: input.Unsubs,
+		})
+		if err != nil {
+			util.ErrResponse(c, http.StatusInternalServerError, err)
+			return
+		}
+
+		if err := errGroup.Wait(); err != nil {
+			logger.For(c).Errorf("error unsubscribing user %s from email types %+v: %s", input.UserID, input.Unsubs, err)
```
jarrel-b, line 97: Should we return an error code here or its ok to return a 200?

emails/subscriptions.go
```diff
@@ -10,46 +10,111 @@ import (
...
+				})
+			default:
+				util.ErrResponse(c, http.StatusBadRequest, fmt.Errorf("unsupported email type: %s", emailType))
 				return
 			}
+
+		}
+
+		logger.For(c).Infof("unsubscribing user %s from email types: %+v", input.UserID, input.Unsubs)
+		err = queries.UpdateUserEmailUnsubscriptions(c, coredb.UpdateUserEmailUnsubscriptionsParams{
```
jarrel-b, line 87: Is it necessary that the calls to sendgrid are successful before updating our subscriptions, or this can be updated regardless of sendgrid's status?

jarrel-b (approved):

emails/subscriptions.go
benny-conn: This can be updated regardless 👍

emails/subscriptions.go
benny-conn: It is ok, if sendgrid does not update we don't need to fail because sendgrid will fail when a user is already on a list for example. It seems like we just want to **try** to keep in sync with sendgrid. If we do want to return an error here, we just have to make sure it does not bubble to the frontend because it is still not something the frontend should handle.

PR #672, opened 2022-11-14
Use forwarded client IP for externalID

jarrel-b:
Before we were using `gc.RemoteIP()` as the value in the hash to generate the `externalD`, but this IP is of the final proxy's IP address. This PR updates the generation to use the forwarded IP from the `X-Forwarded-For` header which is luckily already being delivered to the backend.

Ref: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

### DIFF
--- a/publicapi/gallery.go
+++ b/publicapi/gallery.go
@@ -202,6 +202,6 @@ func IsPrivate(ip net.IP) bool {
 	}
 	// Following RFC 4193, Section 8. IANA Considerations which says:
 	//   The IANA has assigned the FC00::/7 prefix to "Unique Local Unicast".
-	IPv6len := 16
-	return len(ip) == IPv6len && ip[0]&0xfe == 0xfc
+	IPv6Len := 16
+	return len(ip) == IPv6Len && ip[0]&0xfe == 0xfc
 }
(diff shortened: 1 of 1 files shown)


publicapi/gallery.go
```diff
@@ -152,10 +154,54 @@ func (api GalleryAPI) ViewGallery(ctx context.Context, galleryID persist.DBID) (
...
+		proxied := strings.Split(header, ",")
+		forwarded = append(forwarded, proxied...)
+	}
+
+	// Find the first valid, non-private IP that is
+	// closest to the client from left to right.
+	for _, address := range forwarded {
+		if ip := net.ParseIP(address); ip != nil && !IsPrivate(ip) {
+			hash := sha256.New()
+			hash.Write([]byte(viper.GetString("BACKEND_SECRET") + ip.String()))
```
Robinnnnn, line 174: does this need to get added to the deployment file?

Robinnnnn (approved):
thanks jarrel!!

publicapi/gallery.go
jarrel-b: Nope, this was added in the last PR related to this

PR #675, opened 2022-11-15
move deleting and thumbnailing [GAL-664]

benny-conn:
Changes:

- **Changed** deleting of medias to happen after the medias are cached or not cached so that we don't accidentally delete something that the other caching process needs
- **Changed** thumb-nailing to be synchronous because otherwise it might try to thumbnail a video that does not exist on GCP yet

### DIFF
--- a/service/media/media.go
+++ b/service/media/media.go
@@ -147,6 +147,20 @@ func MakePreviewsForMetadata(pCtx context.Context, metadata persist.TokenMetadat
 		mediaType = vidResult.mediaType
 	}
 
+	if !imgResult.cached && imgResult.mediaType.IsImageLike() {
+		logger.For(pCtx).WithFields(logrus.Fields{"tokenURI": truncateString(turi.String(), 100), "imgURL": truncateString(imgURL, 100), "vURL": truncateString(vURL, 100), "name": name}).Debug("MakePreviewsForMetadata imgResult not cached, deleting cached version if any")
+		go deleteMedia(context.Background(), tokenBucket, fmt.Sprintf("image-%s", name), storageClient)
+	}
+	if !vidResult.cached && vidResult.mediaType.IsAnimationLike() {
+		logger.For(pCtx).WithFields(logrus.Fields{"tokenURI": truncateString(turi.String(), 100), "imgURL": truncateString(imgURL, 100), "vURL": truncateString(vURL, 100), "name": name}).Debug("MakePreviewsForMetadata vidResult not cached, deleting cached version if any")
+		go deleteMedia(context.Background(), tokenBucket, fmt.Sprintf("video-%s", name), storageClient)
+	}
+
+	if !imgResult.cached && !vidResult.cached {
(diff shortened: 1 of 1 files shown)


service/media/media.go
```diff
@@ -147,6 +147,23 @@ func MakePreviewsForMetadata(pCtx context.Context, metadata persist.TokenMetadat
...
+		logger.For(pCtx).WithFields(logrus.Fields{"tokenURI": truncateString(turi.String(), 100), "imgURL": truncateString(imgURL, 100), "vURL": truncateString(vURL, 100), "name": name}).Debug("MakePreviewsForMetadata imgResult not cached, deleting cached version if any")
+		go deleteMedia(context.Background(), tokenBucket, fmt.Sprintf("image-%s", name), storageClient)
+	}
+	// if nothing was cached in the video step and the video step did process a video type, delete the now stale cached video
+	if !vidResult.cached && vidResult.mediaType.IsAnimationLike() {
+		logger.For(pCtx).WithFields(logrus.Fields{"tokenURI": truncateString(turi.String(), 100), "imgURL": truncateString(imgURL, 100), "vURL": truncateString(vURL, 100), "name": name}).Debug("MakePreviewsForMetadata vidResult not cached, deleting cached version if any")
+		go deleteMedia(context.Background(), tokenBucket, fmt.Sprintf("video-%s", name), storageClient)
+	}
+
+	// if something was cached but neither media type is animation type, we can assume that there was nothing thumbnailed any thumbnail is stale
```
jarrel-b, line 161: nit: should this say something along the lines of "we can assume that there was nothing thumbnailed therefore any thumbnail is stale"

service/media/media.go
```diff
@@ -614,70 +623,63 @@ outer:
...
-				return mediaType, nil
+				return mediaType, true, nil
 			}
 			logger.For(pCtx).Infof("DECENTRALIZED STORAGE: caching %f mb of raw media with type %s for %s at %s-%s", float64(contentLength)/1024/1024, mediaType, mediaURL, ipfsPrefix, name)
 			if mediaType == persist.MediaTypeAnimation {
 				timeBeforeCache := time.Now()
 				err = cacheRawAnimationMedia(pCtx, reader, bucket, fmt.Sprintf("%s-%s", ipfsPrefix, name), storageClient)
 				if err != nil {
-					return mediaType, err
+					return mediaType, true, err
```
jarrel-b, line 669: Would this be `false`?

service/media/media.go
```diff
@@ -614,70 +623,63 @@ outer:
...
 				}
 				logger.For(pCtx).Infof("cached animation for %s in %s", name, time.Since(timeBeforeCache))
-				return mediaType, nil
+				return mediaType, true, nil
 			}
 			timeBeforeCache := time.Now()
 			err = cacheRawMedia(pCtx, reader, bucket, fmt.Sprintf("%s-%s", ipfsPrefix, name), contentType, storageClient)
 			if err != nil {
-				return mediaType, err
+				return mediaType, true, err
```
jarrel-b, line 677: Would this be `false`?

jarrel-b (approved):
Thanks for crushing, seems like a hairy bug

PR #676, opened 2022-11-16
polish items [GAL-663 GAL-688 GAL-681]

benny-conn:
Changes:

- **Added** email preverify mutation
- **Changed** notification update to set unseen to true
- **Changed** email JWTs to include the email and use it to verify the correct email is being used
- **Changed** `go-redis` version to 9 so that we can use `redislock`
- **Changed** notification dispatcher to first obtain a distributed lock for the given action-owner combination so that we don't accidentally update a notification twice at the same time, causing information to fall through the cracks.
- **Changed** colorglyphs handler to use correct sorting ([this](https://opensea.io/assets/ethereum/0x60f3680350f65beb2752788cb48abfce84a4759e/73) now equals [this](https://storage.googleapis.com/prod-token-content/svg-0x60f3680350f65beb2752788cb48abfce84a4759e-49))
- **Changed** sqlc generated code to use sqlc 1.16 (which required some naming changes in the query file that doesn't actually affect any queries concretely)
- **Added** a `persist.Email` type that for now just ensures emails are inserted in lowercase to the DB but can potentially used later for validation.

Considerations:

- ~~As far as data falling through the cracks for notifications, going to continue looking into things. It does not sound like kaito's admire fell through the cracks because of a concurrent update race condition. Going to look around and see if there is anything else that could allow this to happen as well as play around with it locally.~~ Found the bug causing this, two parameters of the same type were switched causing a DB search to fail because one of the params had an ID and the other did not.


graphql/resolver/schema.resolvers.go
```diff
@@ -935,6 +937,32 @@ func (r *mutationResolver) UpdateNotificationSettings(ctx context.Context, setti
 	return resolveViewerNotificationSettings(ctx)
 }
 
+func (r *mutationResolver) PreverifyEmail(ctx context.Context, email string) (model.PreverifyEmailPayloadOrError, error) {
+	// todo we could have the frontend send the source? right now I don't see any other sources of verification other than signing up
+	result, err := emails.PreverifyEmail(ctx, email, "signup")
```
jarrel-b, line 942: How does sendgrid use `source`, is this just for our own internal tracking to determine where an email came from?

service/notifications/notifications.go
```diff
@@ -111,6 +119,12 @@ func (d *notificationDispatcher) AddHandler(action persist.Action, handler notif
 
 func (d *notificationDispatcher) Dispatch(ctx context.Context, notif db.Notification) error {
 	if handler, ok := d.handlers[notif.Action]; ok {
+		l, err := d.lock.Obtain(ctx, lockKey{ownerID: notif.OwnerID, action: notif.Action}.String(), maxLockTimeout, &redislock.Options{RetryStrategy: redislock.LinearBackoff(1 * time.Second)})
```
jarrel-b, line 122: Why do we lock notifications again? Is this to prevent notifications from a user that is spamming someone? Would this prevent these notifications since the lock is released once the notification is handled, rather than based on rate? Or is `maxLockTimeout` setting the rate limit?

jarrel-b (approved):

service/notifications/notifications.go
jarrel-b: edit: whoops just read your PR description, makes sense!

graphql/resolver/schema.resolvers.go
benny-conn: Yep just for internal tracking it seems, this could be any string

service/notifications/notifications.go
benny-conn: This just ensures that no single notification is updated concurrently. If that was the case, one of the updates might include information that overrides the other. This lock will block for a `maxLockTimeout` of 2 minutes so that if the other notification update is taking forever, this one will just move forward and update anyway (I will make a change to disregard the error because in the case we cannot get the lock we still want to run the handler)

db/queries/core/query.sql
```diff
@@ -13,16 +13,16 @@ SELECT * FROM users WHERE id = ANY(@user_ids) AND deleted = false
     LIMIT $1;
 
 -- name: GetUserByUsername :one
-SELECT * FROM users WHERE username_idempotent = lower(sqlc.arg(username)) AND deleted = false;
+SELECT * FROM users WHERE username_idempotent = lower(sqlc.arg('username')) AND deleted = false;
```
benny-conn, line 16: all of the `sqlc.arg` changes are because sqlc 1.16 complained about batched queries not having parameters and although they do with things like `$1` or `@param` they want `sql.arg` apparently

PR #678, opened 2022-11-17
PII Handling

radazen:
Initial pass at PII Handling. There are still some things left to do.

What this does:
- Drops the users.email column
- Adds a `pii_users` table to contain pii
- Adds a `dev_metadata_users` table for metadata about pii that can be used for development
- Adds a `users_with_pii` view that joins `users` + `pii_users` tables to create a pseudo-table with all user data
- Uses the term `pii` everywhere to make it obnoxiously clear that we should be careful with email addresses

What this doesn't do:
- It doesn't apply appropriate permissions to restrict access to the new `pii_users` table or `users_with_pii` view. I still need to figure this part out.
- It doesn't remove PII from logging statements, but that's an easy fast follow for tomorrow.
- It doesn't add better handling for the uniqueness constraint error we throw when an email address is already taken

Other notes:
- This is my first experiment with Postgres views + sqlc. Results are generally good, but sqlc creates a `UsersWithPii` type based on the table name, instead of a `UserWithPii` type like we'd want.
- I'm not sold on the names I picked for tables and view. The difference between `pii_users` and `users_with_pii` isn't obvious, imho.
- The view is kind of frozen in time; even though it contains `users.*`, that resolves to "the columns that existed when the view was created." As we add new columns, they won't automatically appear in the view. We'll have to drop/recreate the view to make that happen.
- The view has been a very convenient way to create a consistent sqlc type for a queries that need to return joined data


publicapi/user.go
```diff
@@ -61,6 +61,22 @@ func (api UserAPI) GetUserById(ctx context.Context, userID persist.DBID) (*db.Us
 	return &user, nil
 }
 
+func (api UserAPI) GetUserWithPIIById(ctx context.Context, userID persist.DBID) (*db.UsersWithPii, error) {
+	// Validate
```
radazen, line 65: Necessary follow-up: I need to restrict this to the authed user only. Rule of thumb for the public API is "assume any of these methods can be accessed by the public," and if that were true, this function would give out PII to anyone.

util/helpers.go
```diff
@@ -389,3 +390,7 @@ func TruncateWithEllipsis(s string, length int) string {
 	}
 	return s[:length] + "..."
 }
+
+func IsNullOrEmpty(s sql.NullString) bool {
```
radazen, line 394: I can take this out. I had been using `sql.NullString` for emails, but updated when merging with main.

db/queries/core/query.sql
```diff
@@ -572,7 +579,19 @@ select u.* from users u, user_roles ur where u.deleted = false and ur.deleted =
...
+    returning @user_id
+),
+
+upsert_metadata as (
+    insert into dev_metadata_users (user_id, has_email_address) values (@user_id, (@email_address is not null))
+        on conflict (user_id) do update set has_email_address = excluded.has_email_address
+    returning @user_id
+)
+
+update users set email_verified = 0 from upsert_pii, upsert_metadata where users.id = @user_id;
```
jarrel-b, line 594: Just for my understanding, does `upsert_pii` and `upsert_metadata` need to be included in the update statement in order for them to be "executed"?

emails/send.go
```diff
@@ -49,33 +49,34 @@ func sendVerificationEmail(dataloaders *dataloader.Loaders, queries *coredb.Quer
...
 
-		user, err := dataloaders.UserByUserID.Load(input.UserID)
+		userWithPII, err := queries.GetUserWithPIIByID(c, input.UserID)
 		if err != nil {
 			util.ErrResponse(c, http.StatusBadRequest, err)
 			return
 		}
 
-		if user.Email == "" {
+		if userWithPII.PiiEmailAddress.String() == "" {
```
jarrel-b, line 58: Love how explicit this is 👌

jarrel-b (approved):
Looks great!

db/queries/core/query.sql
radazen: Whoa! Good call. I thought they needed to be referenced in order to run, but as it turns out, they don't!

PR #679, opened 2022-11-17
[GAL-480] Add Objkt Provider

jarrel-b:
**Changes**
This PR adds the `ObjktProvider` to Tezos and also some refactoring along the way:

I broke up the multichain `ChainProvider` interface into several, smaller interfaces. This should make it easier to add new providers and also makes it easier to test specific functionality of our providers. 

I also added a generic `FallbackProvider` which calls a "backup" provider for token responses that contain no metadata, which happens semi-frequently for the tzkt API. Tezos is configured with tzkt as the main provider, and objkt as the backup. I went this route because objkt has pretty strict usage requirements and was a little worried that we'd hit their usage limits if we were to call it for each tezos sync as we do for OpenSea. Note this means that this provider only replaces tokens, it doesn't add tokens missing from the first provider as what happens with eth (but we haven't had an issue with missing tokens on tezos yet, so this should be ok)

This could be a little premature given our current scale, but luckily we can easily configure the objkt provider to be another "primary" tezos provider if we want to since it also implements the new `tokenRefresherFetcher` interface.


benny-conn (approved):
Love the separation of the interfaces into smaller interfaces!!

PR #684, opened 2022-11-17
[GAL-699] Bundle same feed events

jarrel-b:


event/event.go
```diff
@@ -286,7 +286,7 @@ func (h notificationHandler) findOwnerForNotificationFromEvent(event db.Event) (
 		}
 		return feedEvent.OwnerID, nil
 	case persist.ResourceTypeUser:
-		return event.UserID, nil
+		return event.SubjectID, nil
```
jarrel-b, line 289: For a user created event, the `SubjectID` is the same as the created user ID. For follow events, the `SubjectID` the user that was followed.

benny-conn (approved):

PR #690, opened 2022-12-06
Benny/redeem merch

benny-conn:
Changes:

- **Added** mutation for redeeming merch and getting the discount codes, with signature verification as well as on chain redeeming

TODO:

- script for adding in the token IDs with discount codes to the DB


publicapi/merch.go
```diff
@@ -0,0 +1,215 @@
...
+
+	tx, err := mer.Redeem(auth, asBigs)
+	if err != nil {
+		return nil, err
+	}
+
+	logger.For(ctx).Infof("redeemed merch items with tx: %s", tx.Hash())
+
+	return result, nil
+}
```
benny-conn, line 215: This is the bulk of what needs review

cmd/merch_prepop/main.go
```diff
@@ -0,0 +1,80 @@
...
+		elapsed := time.Since(start)
+		fmt.Printf("Took %s", elapsed)
+	}()
+
+	pg := postgres.NewPgxClient()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer cancel()
+
+	discountCodeFile, err := os.Open("./discount_codes.json")
```
Robinnnnn, line 36: let's make sure to gitignore this path!

Robinnnnn, line 36: let's make sure to gitignore this path!

publicapi/merch.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+	}
+
+	// verify signature
+
+	// user should have signed the tokenIDs in place of the usual nonce
+	valid, err := api.multichainProvider.VerifySignature(ctx, sig, fmt.Sprintf("%v", tokenIDs), address, walletType)
+	if err != nil {
+		return nil, err
+	}
```
Robinnnnn, line 110: does this mean the message to sign will just display the token IDs?

maybe the statement to sign should be "Verify that you own the following merch token IDs: 1, 40, 69"

Robinnnnn (commented):
could we add some docs as to how the flow works? i feel like there's just enough logic here such that it warrants docs, maybe stately?

publicapi/merch.go
benny-conn: The statement to sign will be exactly the same as the one we use with signin, except instead of a nonce we use the token IDs as an array

benny-conn:
Added an excalidraw!

publicapi/merch.go
benny-conn: In other words it will look like: "Gallery uses this...: [14, 24, etc.]". Doing it like this will make it so we don't have to create another verify signature function specifically for this type of raw signature verify. Our current signature verify will always use that prepend.

Robinnnnn: amazing, thank you!

Robinnnnn (approved):
woooo!! LGTM!

publicapi/merch.go
```diff
@@ -0,0 +1,216 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to get token URI for token %v: %w", tokenID, err)
+		}
+
+		objectType, ok := uriToMerchType[uri]
+		if ok {
+
+			// first check if the token ID has already been redeemed, then just return the code
+			discountCode, err := api.queries.GetMerchDiscountCodeByTokenID(ctx, tokenID)
+			if err != nil {
```
jarrel-b, line 144: I think we'll want to catch for a specific error (I think no rows returned) vs another error before we try to redeem the merch, since we'll probably want to exit early if its an unexpected error

publicapi/merch.go
```diff
@@ -0,0 +1,216 @@
...
+					TokenHex:   tokenID,
+					ObjectType: int32(objectType),
+				})
+				if err != nil {
+					return nil, fmt.Errorf("failed to redeem token %v: %w", tokenID, err)
+				}
+			} else {
+				// override what the blockchain says about redeemed status, if the DB says it is redeemed then there is probably
+				// a transaction in progress for redeeming it, so don't kick off another redeem transaction for this token
+				redeemed[tokenID] = true
```
jarrel-b, line 157: I wonder if it makes any sense to flip the order of this instead:

1) check that the user owns the tokens
2) submit a transaction to redeem the tokens
3) update the state in the db

That way at 3) we don't need to assume the there is a pending txn and that it will succeed since at that point we know for sure all the tokens are redeemed. Also, if the txn fails at 2) tokens wouldn't be marked as redeemed when they actually aren't on chain.

publicapi/merch.go
```diff
@@ -0,0 +1,216 @@
...
+			}
+
+			if discountCode.Valid {
+				result = append(result, &model.DiscountCode{Code: discountCode.String, TokenID: tokenID.String()})
```
jarrel-b, line 161: Would we want to update `result` via index in case `discountCode` isn't `Valid` and the codes don't align positionally to the `tokenIDs` input, or it doesn't matter?

publicapi/merch.go
jarrel-b: I was also wondering between steps 1) and 2) (both in the proposed way or the current way). I think this is a really specific edge case but if we see that a user owns the token, but they have a pending transfer that completes before 2) completes (maybe because gas is really high atm), so they're able to redeem on-chain an nft they don't own and the nft is worth less to the new owner because its now redeemed.

Is it possible to combine 1) and 2) in a single transaction? I guess that would require a helper contract that checks the signer is the current owner of the token to do so? Probably not worth the effort atm?

db/migrations/core/000043_merch.up.sql
```diff
@@ -0,0 +1,14 @@
+create table if not exists merch
```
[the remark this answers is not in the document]
benny-conn: since we could add other merch items to our current contract it will probably be just this one contract

graphql/schema/schema.graphql
```diff
@@ -1384,6 +1384,26 @@ union UnsubscribeFromEmailTypePayloadOrError =
...
+input RedeemMerchInput {
+    tokenIds: [String!]!
+    address: ChainAddressInput!
+    walletType: WalletType!
+    signature: String!
+}
+
+type DiscountCode {
+    code: String!
+    tokenId: String!
```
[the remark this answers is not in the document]
benny-conn: Definitely tokenId should be optional, and maybe in the future if this will expand to other contracts we can add those other fields but for now the merch will only be one contract.

publicapi/merch.go
```diff
@@ -0,0 +1,216 @@
...
+	auth, err := bind.NewKeyedTransactorWithChainID(key, chainID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create authorized transactor: %w", err)
+	}
+	auth.Context = ctx
+
+	asBigs := make([]*big.Int, 0, len(tokenIDs))
+	for _, tokenID := range tokenIDs {
+		if redeemed[tokenID] {
+			// if the token is redeemed, don't try to redeem it again
```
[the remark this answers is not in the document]
benny-conn: We could redeem the same tokens over and over again even if they are redeemed, so it is totally reasonable to redeem all of them every time. This just allows the user to spam transactions and cost us a lot of gas. To prevent this, I have it checking each token is redeemed on chain ahead of time so that it won't keep redeeming the same tokens.

publicapi/merch.go
benny-conn: It doesn't matter in this case because we couple the codes with the token IDs in the returned result

jarrel-b: whoops missed that makes sense!

PR #691, opened 2022-12-06
Fix race condition FallbackProvider resolveTokens

jarrel-b:
This PR fixes the race condition in `resolveTokens` method of the `FallbackProvider` because of parallel writes into `usableTokens`. Its actually possible to concurrently write to a pre-allocated slice by index, avoid the need for channels.


Robinnnnn (approved):
🔥🔥🔥🔥

PR #696, opened 2022-12-07
artifact is better for hic et nunc [GAL-743]

benny-conn:
Changes:

- **Changed** hic et nunc keywords to prioritize artifact which is normally the only URI that exists for a token but is sometimes overridden by a lower quality display URI


jarrel-b (approved):
Nice!

PR #697, opened 2022-12-07
Fix indexer race conditions

jarrel-b:
This PR fixes race conditions that popped up when running `go test -race`. All these issues were because of updating indexer state, so the changes are pretty minor. Also updated the tests to include the `-race` flag so that they'll also check for race conditions.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -43,11 +43,11 @@ jobs:
       - name: Test
         run: |
           mkdir -p /tmp/test-reports
-          gotestsum --junitfile /tmp/test-reports/integration-tests.xml -- -run TestIntegrationTest ./server/... -args -chain ethereum -chainID 1
+          gotestsum --junitfile /tmp/test-reports/integration-tests.xml -- -race -run TestIntegrationTest ./server/... -args -chain ethereum -chainID 1
       - name: "Run indexer tests"
         run: |
           mkdir -p /tmp/test-reports
-          gotestsum --junitfile /tmp/test-reports/indexer-integration-tests.xml -- ./indexer/...
+          gotestsum --junitfile /tmp/test-reports/indexer-integration-tests.xml -- -race ./indexer/...
   unit:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


indexer/indexer.go
```diff
@@ -300,17 +313,13 @@ func (i *indexer) startPipeline(ctx context.Context, start persist.BlockNumber,
...
-		i.lastSyncedBlock = start.Uint64()
-	}
 	logger.For(ctx).Warnf("Finished processing %d blocks from block %d in %s", blocksPerLogsCall, start.Uint64(), time.Since(startTime))
 }
 
 func (i *indexer) startNewBlocksPipeline(ctx context.Context, topics [][]common.Hash) {
 	span, ctx := tracing.StartSpan(ctx, "indexer.pipeline", "startNewBlocksPipeline", sentry.TransactionName("indexer-main:startNewBlocksPipeline"))
 	defer tracing.FinishSpan(span)
 
-	i.isListening = true
```
jarrel-b, line 313: Pulled out to a few levels above in the stack so its easier to see when it gets set

indexer/indexer.go
```diff
@@ -285,7 +299,6 @@ func (i *indexer) startPipeline(ctx context.Context, start persist.BlockNumber,
 	defer tracing.FinishSpan(span)
 
 	startTime := time.Now()
-	i.isListening = false
```
jarrel-b, line 288: Pulled out to a few levels above in the stack since this only needs to get set once

.github/workflows/tests.yml
```diff
@@ -43,11 +43,11 @@ jobs:
       - name: Test
         run: |
           mkdir -p /tmp/test-reports
-          gotestsum --junitfile /tmp/test-reports/integration-tests.xml -- -run TestIntegrationTest ./server/... -args -chain ethereum -chainID 1
+          gotestsum --junitfile /tmp/test-reports/integration-tests.xml -- -race -run TestIntegrationTest ./server/... -args -chain ethereum -chainID 1
       - name: "Run indexer tests"
         run: |
           mkdir -p /tmp/test-reports
-          gotestsum --junitfile /tmp/test-reports/indexer-integration-tests.xml -- ./indexer/...
+          gotestsum --junitfile /tmp/test-reports/indexer-integration-tests.xml -- -race ./indexer/...
```
Robinnnnn, line 50: that's so freaking cool! TIL

Robinnnnn (approved):
this is gorgeous

PR #702, opened 2022-12-08
Fix Indexer Sentry alerting and add CLI

jarrel-b:
### Fix missing Sentry alerts
I updated the `_deploy` file to use `wss://` vs `https://`  since the Sentry hook currently works only for RPC calls over websockets atm. I also removed the `dev/app-dev-indexer-server.yaml` file and the `prod/app-prod-indexer-server.yaml` file since the indexer-api now runs on Cloud Run. 

I also removed the `indexer.rpcResponseTime` alert from Sentry. Looking closer in geth where this line is logged: https://github.com/ethereum/go-ethereum/blob/master/rpc/handler.go#L365, the `duration` is actually the duration of the geth call, and not the actual JSONRPC response time from Alchemy. I fixed the span duration to `100ms`to account for this. I think at some point we'll want to fork geth if we want to collect more useful metrics, but since we're mostly interested in call count this should be ok for now.

### Added a CLI
I forgot what options are available to the Indexer when I started debugging, so I added a CLI which uses cobra: https://github.com/spf13/cobra

Running `go run cmd/indexer/main.go --help` now brings up a useful message:
```
go run cmd/indexer/main.go --help
An NFT indexer lovingly built by your friends at Gallery.
                Source code is available at https://github.com/gallery-so/go-gallery.

Usage:
  indexer [flags]
  indexer [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  help        Help about any command
  server      Run the indexer server

Flags:
  -r, --enable-rpc        always enable RPC calls
  -f, --from-block uint   first block to process
  -h, --help              help for indexer
  -p, --port uint         port to serve on (default 4000)
  -q, --quiet             hide debug logs
  -t, --to-block uint     last block to process

Use "indexer [command] --help" for more information about a command.
```

I also updated some of the logic around input validation that should be a bit more intuitive. Namely, `--from-block` and `--to-block` are required if either is set. And both are required if `--enable-rpc` is not set.


indexer/cmd/root.go
```diff
@@ -0,0 +1,94 @@
...
+	enableRPC      bool
+	quietLogs      bool
+	serviceKeyFile string
+)
+
+var rootCmd = &cobra.Command{
+	Use:   "indexer",
+	Short: "Retrieve all NFTs that have ever been minted",
+	Long: `An NFT indexer lovingly built by your friends at Gallery.
+                Source code is available at https://github.com/gallery-so/go-gallery.`,
```
Robinnnnn, line 28: did this auto-format the second line to have a huge space in front? 😅

Robinnnnn (approved):

benny-conn (approved):
so cool

indexer/cmd/root.go
jarrel-b: I added this haha, I didn't like how long it looked in the terminal

PR #703, opened 2022-12-08
Automatic Persisted Queries

tbezman:
### References

- [GQLGen Documentation](https://gqlgen.com/reference/apq/)
- [Apollo Spec Docs](https://www.apollographql.com/docs/apollo-server/performance/apq/)

### Benefits
- Smaller network payloads
- Smaller frontend bundle size image

### `uploadPersistedQueries` Mutation
This mutation exists so the frontend can upload all of its queries at deploy time. This gives us the following benefits.
- Never experience an APQ waterfall
- Never have to lazily load the query text json file (right now the frontend only has the query hashes in the bundle)

### `FrontendBuildAuthDirectiveHandler`
This is the authorization guard to ensure only our Vercel build pipeline can call that mutation. We don't want arbitrary users polluting our redis cache.


jarrel-b (approved):
awesome stuff!

PR #704, opened 2022-12-08
get merch tokens

benny-conn:
Changes:

- **Added** query for getting merch tokens and redeemed status
- **Added** multichain function for getting tokens of contract owned by a given wallet which required some changes, most notably to the `(*multichain.Provider).upsertTokens` function. In the dedupe step, there is now a final optional step that will include every one of a user's current tokens so that we can upsert small chunks of tokens without messing up a user's full list of tokens. This will basically make the upsert append/update only, no tokens will get removed. This optional step should only occur when we are searching for and updating a sub-section of a user's tokens, where as with a full sync, we do not want to include a user's current tokens in the upsert because some tokens may have been transferred out. In this case, we want the providers to be the source of truth.

### DIFF
--- a/publicapi/merch.go
+++ b/publicapi/merch.go
@@ -3,6 +3,7 @@ package publicapi
 import (
 	"context"
 	"crypto/x509"
+	"encoding/json"
 	"fmt"
 	"math/big"
 	"strings"
@@ -40,6 +41,15 @@ var uriToMerchType = map[string]int{
 	"ipfs://QmSPdA9Gg8xAdVxWvUyGkdFKQ8YMVYnGjYcr3cGMcBH1ae": merchTypeCard,
 }
 
+type merchAttribute struct {
+	TraitType string `json:"trait_type"`
+	Value     string `json:"value"`
+}
+
+type merchMetadata struct {
+	Attributes []merchAttribute `json:"attributes"`
+}
+
 type MerchAPI struct {
 	repos              *postgres.Repositories
 	queries            *db.Queries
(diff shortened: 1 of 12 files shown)


publicapi/merch.go
```diff
@@ -51,6 +61,81 @@ type MerchAPI struct {
...
+					}
+					break
+				}
+			}
+		}
+
+		merchTokens[i] = t
+	}
+
+	return merchTokens, nil
```
Robinnnnn, line 136: we can safely assume that if the address doesn't hold any tokens, the endpoint would just return an empty array instead of a failure right?

service/multichain/eth/eth.go
```diff
@@ -115,6 +115,36 @@ func (d *Provider) GetTokensByContractAddress(ctx context.Context, contractAddre
 	return tokensToChainAgnostic(tokens.NFTs), contract, nil
 }
 
+// GetTokensByContractAddressAndOwner retrieves tokens for a contract address owned by owner on the Ethereum Blockchain
+func (d *Provider) GetTokensByContractAddressAndOwner(ctx context.Context, owner, contractAddress persist.Address, limit, offset int) ([]multichain.ChainAgnosticToken, multichain.ChainAgnosticContract, error) {
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("%s/nfts/get?contract_address=%s&address=%s&limit=%d&offset=%d", d.indexerBaseURL, contractAddress, owner, limit, offset), nil)
```
Robinnnnn, line 121: any chance the indexer would be behind after the user makes the purchase?

this is probably an OK edge case, as long as we let the user know there may be a small delay before they can redeem post-purchase

service/multichain/multichain.go
```diff
@@ -401,6 +397,51 @@ func (p *Provider) GetCommunityOwners(ctx context.Context, communityIdentifiers
...
+
+	addressToContracts, err := p.upsertContracts(ctx, contracts)
+	if err != nil {
+		return nil, err
+	}
+
+	user, err := p.Repos.UserRepository.GetByChainAddress(ctx, wallet)
+	if err != nil {
+		return nil, err
+	}
```
Robinnnnn, line 440: this is an unlikely scenario, but realized the user may not exist! since the wallet connector on the merch page is stateless.

i think we can reasonably assume that everyone who buys merch will have an account, but worth flagging

service/multichain/tezos/objkt.go
```diff
@@ -337,6 +337,86 @@ func (p *TezosObjktProvider) GetTokensByContractAddress(ctx context.Context, con
...
+				OwnerAddress:    holder.Holder_Address,
+				BlockNumber:     persist.BlockNumber(token.Level),
+			}
+			returnTokens = append(returnTokens, agnosticToken)
+		}
+	}
+
+	return returnTokens, agnosticContract, nil
+}
+
```
Robinnnnn, line 419: wow, thanks for handling the tezos providers!!

Robinnnnn (approved):

publicapi/merch.go
benny-conn: yep!

service/multichain/eth/eth.go
benny-conn: the indexer tends to be pretty quickly up to date but yes this is a possibility of course.

service/multichain/multichain.go
benny-conn: is there some way you want to handle this case or should we expect them to have an account?

jarrel-b (approved):
lgtm! just had a few comments / suggestions

service/multichain/multichain.go
```diff
@@ -855,18 +896,22 @@ func (p *Provider) upsertTokens(ctx context.Context, allTokens []chainTokens, ad
 	return newTokens, nil
 }
 
-func (p *Provider) prepareTokensOfContractForUser(ctx context.Context, allTokens []chainTokens, addressesToContracts map[string]persist.DBID, allUsersTokens []persist.TokenGallery, user persist.User, timeStamp time.Time) ([]persist.TokenGallery, error) {
+func (p *Provider) prepareTokensOfContractForUser(ctx context.Context, allTokens []chainTokens, addressesToContracts map[string]persist.DBID, user persist.User, timeStamp time.Time, includeCurrentTokens bool) ([]persist.TokenGallery, error) {
```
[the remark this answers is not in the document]
benny-conn: great idea, will experiment with this after this PR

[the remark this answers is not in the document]
benny-conn: nvm, just changed it to this!

service/multichain/multichain.go
Robinnnnn: let's assume they have an account for now!

PR #705, opened 2022-12-08
allowlist script

benny-conn:
Changes:

- **Added** script for turning a list of user IDs into a list of the first ethereum address they each have


Robinnnnn (approved):
2 strong

PR #711, opened 2022-12-09
Retool mutations

tbezman:
Two new mutations with `@retoolAuth`

- `syncTokensForUsername`: Same inputs as `syncTokens` except we take a `username` because we're not logged in
- `banUserFromFeed` to insert a user into the blocked table.

I left some comments in the PR w/ some questions / uncertainty.

### DIFF
--- a/db/queries/core/query.sql
+++ b/db/queries/core/query.sql
@@ -400,6 +400,9 @@ select * from feed_events where deleted = false
 -- name: IsFeedUserActionBlocked :one
 SELECT EXISTS(SELECT 1 FROM feed_blocklist WHERE user_id = $1 AND action = $2 AND deleted = false);
 
+-- name: BlockUserFromFeed :exec
+INSERT INTO feed_blocklist (id, user_id, action) VALUES ($1, $2, $3);
+
 -- name: GetAdmireByAdmireID :one
 SELECT * FROM admires WHERE id = $1 AND deleted = false;
 
--- a/go.mod
+++ b/go.mod
@@ -60,7 +60,7 @@ require (
 	github.com/shurcooL/graphql v0.0.0-20200928012149-18c5c3165e3a
 	github.com/sirupsen/logrus v1.8.1
 	github.com/spf13/afero v1.8.2 // indirect
-	github.com/spf13/cobra v1.6.1 // indirect
+	github.com/spf13/cobra v1.6.1
 	github.com/spf13/viper v1.10.1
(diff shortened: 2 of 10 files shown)


go.mod
```diff
@@ -60,7 +60,7 @@ require (
 	github.com/shurcooL/graphql v0.0.0-20200928012149-18c5c3165e3a
 	github.com/sirupsen/logrus v1.8.1
 	github.com/spf13/afero v1.8.2 // indirect
-	github.com/spf13/cobra v1.6.1 // indirect
+	github.com/spf13/cobra v1.6.1
```
tbezman, line 63: this happened automatically, thoughts?

db/queries/core/query.sql
```diff
@@ -400,6 +400,9 @@ select * from feed_events where deleted = false
 -- name: IsFeedUserActionBlocked :one
 SELECT EXISTS(SELECT 1 FROM feed_blocklist WHERE user_id = $1 AND action = $2 AND deleted = false);
 
+-- name: BlockUserFromFeed :exec
+INSERT INTO feed_blocklist (id, user_id, action) VALUES ($1, $2, $3);
```
tbezman, line 404: it looks like I can't have an `ON CONFLICT` here since `user_id` is not unique on this table. not sure if this is a concern

go.mod
jarrel-b: 👍 the indexer uses cobra for its cli now

db/queries/core/query.sql
jarrel-b: Yup! The idea was that we ban specific (user, action)'s

graphql/schema/schema.graphql
```diff
@@ -1503,6 +1515,8 @@ type Mutation {
     # Retool Specific Mutations
     addRolesToUser(username: String!, roles: [Role]): AddRolesToUserPayloadOrError @retoolAuth
     revokeRolesFromUser(username: String!, roles: [Role]): RevokeRolesFromUserPayloadOrError @retoolAuth
+    syncTokensForUsername(username: String!, chains: [Chain!]!): SyncTokensForUsernamePayloadOrError @retoolAuth
+    banUserFromFeed(username: String!, action: String!): BanUserFromFeedPayloadOrError @retoolAuth
```
jarrel-b, line 1519: Think `action` could be `Action!` instead

jarrel-b (approved):

PR #721, opened 2022-12-13
correct string type

benny-conn:
Changes:

- **Changed** ownership check to use the correct string method


graphql/resolver/schema.resolvers.go
```diff
@@ -1003,7 +1003,10 @@ func (r *mutationResolver) RedeemMerch(ctx context.Context, input model.RedeemMe
 	for i, id := range input.TokenIds {
 		tokenIDList[i] = persist.TokenID(id)
 	}
-	codes, err := publicapi.For(ctx).Merch.RedeemMerchItems(ctx, tokenIDList, persist.NewChainPubKey(persist.PubKey(input.Address.String()), input.Address.Chain()), input.Signature, input.WalletType)
+	if input.Address == nil {
+		return nil, fmt.Errorf("address is required")
```
benny-conn, line 1007: temporary, just put this quickly so jaks could use it, idk why the address is a pointer when it has a `!` in the schema but will look into it more

PR #725, opened 2022-12-13
Add multichain chain validation

jarrel-b:
This PR adds a validation step that's ran when the multichain Provider is instantiated. This should help ensure that the features we expect are supported for each chain and prevent issues were providers are silently skipped over because they don't implement an interface.


Robinnnnn (approved):
thanks for doing this man 🙏

server/server.go
```diff
@@ -183,7 +183,7 @@ func setDefaults() {
 		logger.For(nil).Info("running in non-local environment, skipping environment configuration")
 	} else {
 		fi := "local"
-		if len(os.Args) > 0 {
+		if len(os.Args) > 1 {
```
jarrel-b, line 186: Think this should be `1`, since `0` is the name of the executable

PR #726, opened 2022-12-14
initial mutations and schema adjustments (db and graphql)

benny-conn:
Changes:

- **Added** the mutations, inputs/types, and adjustments from [this doc](https://www.notion.so/gallery-so/Multi-Gallery-Eng-Spec-2fc908501a9943028cc06372a279796f)
- **Added** new gallery info and position columns to db as well as a `featuredGallery` user column and a unique index on `(position,owner_user_id)`
- **Added** queries and updated queries in preparation for multi-gallery. For example, including the new gallery `position` column in ordering and inserts.
- **Added** resolvers for new fields like `featuredGallery` on users and `name`, `description`, `hidden`, `position`, `tokenPreviews` on galleries
- **Added** new gallery fields to user on-boarding process

TODO:

- [x] constrain deletion to not delete featured gallery or final gallery (or re-assign the featured gallery if we want to go that direction product-wise)
- [x] `gallery_id` in collections table to ensure a collection from another gallery does not get included in another gallery's update
- [x] mutation for updating everything involved with a gallery at once (gallery info, collection order/collection info/new collections/deleted collections, token order)


radazen (commented):
Looking good, @benny-conn! I left a bunch of comments/suggestions/discussion points, but I think everything is going in the right direction! Also happy to discuss any of my comments on a call if you'd like!

publicapi/gallery.go
```diff
@@ -103,6 +172,83 @@ func (api GalleryAPI) UpdateGalleryCollections(ctx context.Context, galleryID pe
...
+		return err
+	}
+
+	userID, err := getAuthenticatedUser(ctx)
+	if err != nil {
+		return err
+	}
+
+	for _, position := range positions {
+		err = api.queries.UpdateGalleryPosition(ctx, db.UpdateGalleryPositionParams{
```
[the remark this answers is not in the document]
radazen: I think this should work as a query (tested with different values on a different table, but the gist should be correct!):

```
with updates as (
    select unnest(@gallery_ids::text[]) as id, unnest(@position::text[]) as position
)
update galleries g set position = updates.position from updates where g.id = updates.id;
```

graphql/schema/schema.graphql
```diff
@@ -459,7 +465,9 @@ type NotificationsConnection @goEmbedHelper {
     pageInfo: PageInfo
 }
 
-type Viewer {
+type Viewer @goGqlId(fields: ["userId"]) {
+  id: ID!
+  userId: DBID!
```
tbezman, line 470: not sure if this needs to be here. generally we keep relational ids out of the schema

graphql/schema/schema.graphql
```diff
@@ -1150,6 +1158,14 @@ type LoginPayload {
 type LogoutPayload {
     viewer: Viewer
 }
+input CreateUserInput {
+	username: String!
+	bio: String
+	email: String
+	galleryName: String
+	galleryDescription: String
+    galleryPosition: String!
```
tbezman, line 1167: feels weird, can we just assume a default here? keeps the schema extra pure. 

I also think we should just use the go lib to assume a default value. We could assume `m` but it looks like the library assumes `a0` as a default. https://github.com/rocicorp/fracdex

maybe we should just import this lib now to get it out of the way. i don't want it to be a source of friction in the future.

graphql/schema/schema.graphql
```diff
@@ -1504,6 +1594,13 @@ type Mutation {
 
     viewGallery(galleryId: DBID!): ViewGalleryPayloadOrError
 
+    createGallery(input: CreateGalleryInput!): CreateGalleryPayloadOrError @authRequired
+	updateGalleryHidden(input: UpdateGalleryHiddenInput!): UpdateGalleryHiddenPayloadOrError @authRequired
+	deleteGallery(galleryId: DBID!): DeleteGalleryPayloadOrError @authRequired
+    updateGalleryOrder(input: UpdateGalleryOrderInput!): UpdateGalleryOrderPayloadOrError @authRequired
+	updateGalleryInfo(input: UpdateGalleryInfoInput!): UpdateGalleryInfoPayloadOrError @authRequired
+	updateFeaturedGallery(galleryId: DBID!): UpdateFeaturedGalleryPayloadOrError @authRequired
+
```
tbezman, line 1603: seems like we're consolidating a few of these into a single mutation. Here's what we jammed on for reference.

```graphql
input UpdateCollectionInput {
    dbid: DBID
    name: String!
    collectorsNote: String!
    tokens: [DBID!]!
    layout: CollectionLayoutInput!
    tokenSettings: [CollectionTokenSettingsInput!]!
}

input CreateCollectionInputV2 {
    name: String!
    collectorsNote: String!
    tokens: [DBID!]!
    layout: CollectionLayoutInput!
    tokenSettings: [CollectionTokenSettingsInput!]!
}


input UpdateGalleryInput {
    galleryId: DBID!

    name: String
    description: String

    caption: String

    hiddenCollections: [DBID!]
    deletedCollections: [DBID!]

    collections: [UpdateCollectionInput]
    createdCollections: [CreateCollectionInputV2]
}
```

db/migrations/core/000044_multi_gallery.up.sql
```diff
@@ -0,0 +1,7 @@
+alter table galleries add column name varchar;
+alter table galleries add column description varchar;
+alter table galleries add column hidden boolean NOT NULL default false;
+alter table galleries add column position varchar NOT NULL default 'm';
+alter table users add column featured_gallery varchar;
+
+create unique index if not exists position_idx on galleries (position, owner_user_id) where deleted = false;
```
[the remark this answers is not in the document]
benny-conn: Wow! Thank you for this amazing explanation!

publicapi/gallery.go
```diff
@@ -31,6 +36,231 @@ type GalleryAPI struct {
...
+		err = q.UpdateCollectionTokens(ctx, db.UpdateCollectionTokensParams{
+			ID:   collection.Dbid,
+			Nfts: collection.Tokens,
+		})
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
benny-conn, line 238: This is what needs reviewing. 

I tried getting `updateCollectionsInfoAndTokens` to perform one update with unnests but we can't unnest an array of arrays of tokens so I had to perform a separate update right there at the end. I forgot if we found a solution for the multi-dimensional array unnest @radazen ?

service/persist/postgres/postgres.go
```diff
@@ -232,3 +227,27 @@ type Repositories struct {
...
+		EarlyAccessRepository: NewEarlyAccessRepository(pq, queries),
+		WalletRepository:      NewWalletRepository(pq, queries),
+		AdmireRepository:      NewAdmireRepository(queries),
+		CommentRepository:     NewCommentRepository(pq, queries),
+	}
+}
+
+func (r *Repositories) BeginTx(ctx context.Context) (pgx.Tx, error) {
+	return r.pool.BeginTx(ctx, pgx.TxOptions{})
+}
```
benny-conn, line 253: Not sure if this was the best place to create a tx for use with the `coredb.Queries` type. Was going to pass in the pool to the public API but I thought it would be weird to pass in such a low level type when the public API also had access to the higher level repos and queries that used it. Let me know if there is a better place for creating a tx.

publicapi/gallery.go
radazen: I put together a proof-of-concept query that unnests a pair of extra arrays (`start_indices[]` and `end_indices[]`) to slice a single multi-dimensional array into appropriate sub-arrays...but it's a little messy, and I haven't tested the approach well enough to recommend it for real use yet. For our purposes here, I think a loop is fine for now, and we can revisit if the slicing thing works well*.

\* - I also think there's a better chance of having it work well now that we have generics available to us! To really make it work, I needed a utility method that would append elements to the parameter struct and also append the appropriate indices into the start/end arrays, but without generics, I couldn't write a reusable version of that.

radazen: But while we're talking about it, the technique would be something like:

```
insert into tokens (id, owned_by_wallets)

select (
  id,
  owned_by_wallets[owned_by_wallets_start_index:owned_by_wallets_end_index]
)

from (
  select (
    unnest(@id::varchar(255)[]) as id,
    unnest(@owned_by_wallets_start_index::int[]) as owned_by_wallets_start_index,
    unnest(@owned_by_wallets_end_index::int[]) as owned_by_wallets_end_index,
    @owned_by_wallets::varchar(255)[] as owned_by_wallets,
  )
)
```

where `owned_by_wallets` is a one-dimensional array with all `owned_by_wallets` values and the index arrays specify which subset of `owned_by_wallets` belongs to each unnested row.

service/persist/postgres/postgres.go
radazen: Yup, looks A-OK to me! At some point I think this will live in the `storage` layer that'll act as a wrapper around raw `Queries`, but for the time being, this is a good place!

publicapi/gallery.go
```diff
@@ -31,6 +36,231 @@ type GalleryAPI struct {
...
+	curGal, err := api.loaders.GalleryByGalleryID.Load(update.GalleryID)
+	if err != nil {
+		return db.Gallery{}, err
+	}
+
+	tx, err := api.repos.BeginTx(ctx)
+	if err != nil {
+		return db.Gallery{}, err
+	}
+
```
radazen, line 97: I think you want to `defer tx.Rollback(ctx)` here to make sure the transaction gets rolled back if the commit doesn't succeed.

publicapi/gallery.go
benny-conn: Ahhh I see, that is super clever!!

publicapi/gallery.go
```diff
@@ -31,6 +36,231 @@ type GalleryAPI struct {
...
+		"created_collections": {update.CreatedCollections, "created_collections"},
+	}); err != nil {
+		return db.Gallery{}, err
+	}
+
+	curGal, err := api.loaders.GalleryByGalleryID.Load(update.GalleryID)
+	if err != nil {
+		return db.Gallery{}, err
+	}
+
```
radazen, line 92: Once we have `curGal`, we probably want to check `getAuthenticatedUser(ctx)` and verify that the currently authed user owns the gallery they're about to update

db/queries/core/query.sql
```diff
@@ -633,5 +633,41 @@ select discount_code from merch where token_id = @token_hex and redeemed = true
...
+with updates as (
+    select unnest(@gallery_ids::text[]) as id, unnest(@positions::text[]) as position
+)
+update galleries g set position = updates.position, last_updated = now() from updates where g.id = updates.id and deleted = false;
+
+-- name: UpdateGalleryInfo :exec
+update galleries set name = @name, description = @description, last_updated = now() where id = @id and deleted = false;
+
+-- name: UpdateGallery :exec
+update galleries set name = @name, description = @description, collections = @collections, last_updated = now() where galleries.id = @id and galleries.deleted = false and (select count(*) from collections c where c.id = any(@collections) and c.gallery_id = @gallery_id and c.deleted = false) = array_length(@collections, 1);
```
radazen, line 649: Very slick

PR #728, opened 2022-12-14
Fix syncTokensForUsername for Retool and deprecate old admin behavior

tbezman:
Since we're moving to Retool for admin-esque things. I've separated out the logic for syncing tokens based on the use case.

1. Syncing tokens as an admin (from Retool). Takes a username and is assumed to be authenticated by the time you're in that context.
2. Syncing tokens as a user. This mutation should not take a `userId`. Totally made sense before when we didn't have Retool.

This is technically a breaking change (removing a parameter from a mutation) but we're not actually using it on the frontend so I'm not super concerned.

### DIFF
--- a/publicapi/token.go
+++ b/publicapi/token.go
@@ -289,16 +289,27 @@ func isAdminUser(userID persist.DBID) bool {
 	}
 }
 
-func (api TokenAPI) SyncTokens(ctx context.Context, chains []persist.Chain, asUserID *persist.DBID) error {
-	// No validation to do
-	authedUserID, err := getAuthenticatedUser(ctx)
-	if err != nil {
-		return err
+func (api TokenAPI) SyncTokensAdmin(ctx context.Context, chains []persist.Chain, userID persist.DBID) error {
+	if err := api.throttler.Lock(ctx, userID.String()); err != nil {
+		return ErrTokenRefreshFailed{Message: err.Error()}
 	}
 
-	userID := authedUserID
-	if asUserID != nil && isAdminUser(authedUserID) {
-		userID = *asUserID
+	defer api.throttler.Unlock(ctx, userID.String())
+
+	if err := api.multichainProvider.SyncTokens(ctx, userID, chains); err != nil {
+		// Wrap all OpenSea sync failures in a generic type that can be returned to the frontend as an expected error type
+		return ErrTokenRefreshFailed{Message: err.Error()}
+	}
+
+	return nil
+}
+
+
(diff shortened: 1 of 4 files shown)


publicapi/token.go
```diff
@@ -289,16 +289,27 @@ func isAdminUser(userID persist.DBID) bool {
...
 	}
 
-	userID := authedUserID
-	if asUserID != nil && isAdminUser(authedUserID) {
-		userID = *asUserID
+	return nil
+}
+
+
+func (api TokenAPI) SyncTokens(ctx context.Context, chains []persist.Chain) error {
```
tbezman, line 308: assumes authenticated user

publicapi/token.go
```diff
@@ -289,16 +289,27 @@ func isAdminUser(userID persist.DBID) bool {
 	}
 }
 
-func (api TokenAPI) SyncTokens(ctx context.Context, chains []persist.Chain, asUserID *persist.DBID) error {
-	// No validation to do
-	authedUserID, err := getAuthenticatedUser(ctx)
-	if err != nil {
-		return err
+func (api TokenAPI) SyncTokensAdmin(ctx context.Context, chains []persist.Chain, userID persist.DBID) error {
```
tbezman, line 292: assumes you're in an admin authenticated context and continues to sync the user.

radazen (approved):
Looks good to me! One comment below, though it's more a comment on the pattern for admin methods in general and not about this particular PR.

--

Do you know if we have a mechanism on the backend to quickly verify that a user has the appropriate role to perform an admin action? I think a good rule of thumb for the public API layer's responsibilities is "could we replace GraphQL with something else and keep using this same public API without modification?" For that to work, the public API has to be responsible for handling authorization.

The schema directive is still super helpful -- it's a great way to self-document admin methods and to return a consistent error from the GraphQL layer if someone doesn't have the necessary role -- but imho it shouldn't be the only thing stopping a non-admin from calling an admin-only method in the public API. (auth works similarly; schema directive errors out if you're not authenticated, but public API methods that require auth will also throw errors if the user isn't authenticated)

FWIW, I don't think this is a blocker -- no one can call that public API method without using GraphQL atm -- but I think it's something we should start doing with our admin methods overall.

tbezman:
@radazen makes sense. 

I think I'd need to back up here and factor out some of the Retool auth middleware so it looks something like this?

1. Gin middleware to check if we're authorizing with Retool
2. `@retoolAuth` handler checks the gin context to see if we're authed with Retool
3. helper `canCommitAdminAction` checks the gin context to see if we are either
  - Authed with retool
  - Authed as an admin user

PR #730, opened 2022-12-14
trim space and un escape svg

benny-conn:
Changes:

- **Added** a space trim to the final media URLs because the browser freaks out when there is trailing space on an image url
- **Added** an unescape to SVGs because often times svgs will be query escaped in the token URI (have to test that this does not affect the already unescaped svgs before merging)


Robinnnnn (approved):
LGTM!! Does the metadata processing package come with unit tests?

PR #731, opened 2022-12-15
Upgrade services to go1 .19

benny-conn:
Changes:

- **Added** docker files separately for prod and dev, dev using `go1.20`


benny-conn:
1.19 instead (stable version)

Robinnnnn (approved):

PR #732, opened 2022-12-15
[GAL-558] Better handling for Sentry traces

radazen:
## The Problem
Sentry limits us to 1000 spans in our traces*. We create a new span for every GraphQL resolver, which means that we can go over the 1000-span limit pretty easily. Viewing a large gallery can generate thousands -- or even tens of thousands -- of spans. When this happens, any spans past the first 1000 are dropped, leaving our traces incomplete (and often lacking usable data).

\* - No longer technically true. They've removed the hard limit, but we're still limited by the amount of data we send in a single transaction, which is what the span limit was there to enforce. TL;DR: we still want to reduce the number of spans we send, and 1000 is still a sane limit.

## How This PR Addresses the Problem
If we can only keep 1000 spans, we should try to keep the 1000 _most important_ spans. Many of the spans generated by our GraphQL resolvers are trivial dataloader cache hits that take fractions of a millisecond. Those seem like great spans to get rid of, right?

The approach taken in this PR is:

- Let Sentry record up to 100k spans (chosen somewhat arbitrarily)
- Make sure every parent span fully encapsulates its child spans (i.e. a parent should never have a shorter span duration than its children)
  - This ensures that we'll never leave an orphaned child span. Parent durations are always >= their children, so they'll never get filtered _before_ their children
- Keep all top-level spans (spans without parents) and all spans longer than `x milliseconds`
- `while # spans > 1000...`
  - double our `x milliseconds` value, do another filtering pass
- If we couldn't reduce our total span count below 1000 in `n` filtering passes, truncate our span array to 1000

Eventually, we end up with the <= 1000 spans with the greatest impact. For every span we filter out, we "roll it up" by including metadata about it in its nearest non-filtered ancestor span. For example, a `tokens` resolver that filtered out 401 child spans had this metadata attached:

image

^ These are 401 trivial cache hits that took ~2.3ms total across many noisy spans.

## Performance
This filtering only happens on transactions that will actually be sent to Sentry, which is currently (I think) 5% of queries. In my testing, filtering 9500 spans down to ~300 took 2.5ms. Even with more spans, I think the impact of filtering is worth the time taken. Filtering stats are recorded in the transaction's context info, like this:

image

## Other stuff
This PR also uses an updated version of our `dataloaden` fork to improve dataloader tracing. `db.get` calls are now grouped under a new `dataloader.fetch` span type, so we can see which db calls are part of the same dataloader batch:

image

## One final question
Should we always filter out insignificantly short spans? Right now I'm _not filtering any spans_ if a trace has fewer than 1000 spans. So a view of my not-very-big Gallery, which generates 135 total spans, would keep all of its spans -- even the 0.05ms cache hits. In theory, this sounds useful, since it's nice to have the additional data if we have the budget for it. In practice, all of those tiny spans just feel noisy. I feel like I've been able to see the big picture more easily in spans that have had filtering applied to them.


Robinnnnn (commented):
this is outrageously cool ezra, thank you 👏 👏 👏 

i haven't looked at the PR in depth but +1 on filtering out tiny spans

service/sentry/sentry.go
```diff
@@ -39,23 +39,6 @@ var logToSentryLevel = map[logrus.Level]sentry.Level{
 }
 var sentryTrailLimit = 8
 
-type authContext struct {
```
radazen, line 42: I had to get rid of these as part of upgrading from sentry 0.13.0 to 0.15.0. `Context` has to be a `map[string]interface{}` now, so I defined them inline as maps.

service/sentry/sentry.go
```diff
@@ -339,6 +327,287 @@ func TransactionNameSafe(name string) sentry.SpanOption {
...
+
+// Sentry uses milliseconds for its trace fields, and it keeps things consistent if we do it too
+func durationToMsFloat(duration time.Duration) float64 {
+	return float64(duration.Microseconds()) / 1000.0
+}
+
+// SpanFilterEventProcessor applies a progressive filter to spans, removing the shortest spans until the total span count
+// is less than the specified maxSpans value. Initially, spans shorter than minSpanDuration will be dropped, but each filter
+// pass (up to maxFilterPasses) will double the minSpanDuration until enough spans have been filtered out.
+func SpanFilterEventProcessor(ctx context.Context, maxSpans int, minSpanDuration time.Duration, maxFilterPasses int) sentry.EventProcessor {
```
radazen, line 342: I have a strong preference for _avoiding_ functions that are this long, but it's also nice being able to define intermediate types (`rollup`, `spanData`) scoped to a function so they aren't just hanging around as real types for the rest of the package to use. Still, this function is probably too long, and I probably should have made the it shorter and defined the types outside of it so sub-functions could use them, too.

service/sentry/sentry.go
```diff
@@ -339,6 +327,287 @@ func TransactionNameSafe(name string) sentry.SpanOption {
...
+			}
+		}
+
+		var filteredSpans []*spanData
+
+		reportedFilterPasses := 0
+		reportedMinSpanDuration := time.Duration(0)
+
+		// Keep filtering and doubling the minSpanDuration until we've filtered out enough spans
+		for filterPass, minDurationForPass := 1, minSpanDuration; filterPass <= maxFilterPasses; filterPass++ {
```
jarrel-b, line 422: neat syntax! whats the need for `minSpanDuration` in the loop? was `minDurationForPass` originally intended to be a multiple of `minSpanDuration`?

service/sentry/sentry.go
```diff
@@ -339,6 +327,287 @@ func TransactionNameSafe(name string) sentry.SpanOption {
...
+			reportedMinSpanDuration = minDurationForPass
+
+			spans = allowedSpans
+			minDurationForPass *= 2
+		}
+
+		var filterPath []*spanData
+
+		// Roll filtered spans up to their nearest unfiltered ancestor
+		for _, span := range filteredSpans {
```
jarrel-b, line 450: This chunk of code is super neat 👍

service/sentry/sentry.go
```diff
@@ -339,6 +327,287 @@ func TransactionNameSafe(name string) sentry.SpanOption {
...
+				relativeEnd := intervalEnd.Sub(rollupTo.RawSpan.StartTime)
+
+				statsForName := fmt.Sprintf("%s (%d) | %.3fms avg | [%.3fms - %.3fms] range", name, len(r.RawSpans),
+					durationToMsFloat(avgDuration), durationToMsFloat(relativeStart), durationToMsFloat(relativeEnd))
+
+				childStats := make(map[string]interface{})
+				stats[statsForName] = childStats
+
+				if r.Children != nil {
+					getRollupStats(rollupTo, r.Children, childStats)
```
jarrel-b, line 546: Would the `stats` arg need to be a pointer since values are copied, or do maps work differently?

jarrel-b (approved):
Took me awhile to grok but as always the comments were super helpful, looks awesome!

jarrel-b:
> this is outrageously cool ezra, thank you 👏 👏 👏
> 
> i haven't looked at the PR in depth but +1 on filtering out tiny spans

Also +1 on tiny spans since they don't seem very actionable

service/sentry/sentry.go
radazen: Good call! `minSpanDuration` had previously been used for some reporting at the end, but that's no longer the case. `minDurationForPass` is unnecessary; I can just use `minSpanDuration` directly and keep doubling its value instead of using a second variable.

service/sentry/sentry.go
radazen: I think slices are the weird one here. AFAIK maps and slices are both value types that contain a pointer to an underlying data structure, so copies of their values will continue to point to the same underlying data structure...but with slices, appending creates a _new_ value that may not point to the same underlying array anymore, and definitely won't have the same length parameter. That doesn't seem to be true of a map; altering its contents updates the existing underlying data structure, and all copies of the `map` value type continue to point to the same underlying structure.

At least, that's my understanding of how it works!

PR #734, opened 2022-12-16
dev notif email options, more debug messages, notif week check

benny-conn:
Changes:

- **Added** optional inputs to the dev send notifications handler that will allow you to send a notification email on behalf of a user to a specified email
- **Changed** view notification handler to check for duplicate users in the week like it describes in its description comment, instead of by the day like it is currently doing

To send an email to a specific email address for the notifs of a given user ID, send a post request with the given struct tags as JSON to the `/send/notifications` endpoint:

```
type sendNotificationEmailHttpInput struct {
	UserID         persist.DBID  `json:"user_id"`
	ToEmail        persist.Email `json:"to_email"`
	SendRealEmails bool          `json:"send_real_emails"`
}
```

If you do not include `UserID` or `ToEmail` it will send to all emails. The new field `SendRealEmails` can be used either way to specify if we actually want to send the email or just log what would have been sent.

This endpoint also requires you to pass in the header `Authorization: ${ADMIN_PASS}` where ADMIN_PASS is in the env file


benny-conn:
Merge after updating to `go1.19` because this uses `atomic.Uint64` which does not exist yet in 1.16

Robinnnnn (approved):

PR #735, opened 2022-12-19
only notifications in last week for email

benny-conn:
Changes:

- **Changed** recent notification email to only use notifications created after one week before the email being sent


Robinnnnn (approved):
🔥

PR #743, opened 2022-12-23
Add tests

jarrel-b:
This PR adds the initial set of tests for the backend mostly around user creation. The files of interest are:
* `graphql/graphql_test.go` - contains the actual test themselves
* `graphql/testdata/operations.gql` - contains queries used in testing
* `graphql/fixtures_test.go` - contains re-usable test fixtures


docker/docker.go
```diff
@@ -11,277 +11,195 @@ import (
 	"time"
```
jarrel-b, line 11: No need to review this file! `docker` is just used for testing and I wanted to re-write it this because I hardly knew go at the time.

graphql/graphql_test.go
```diff
@@ -0,0 +1,592 @@
...
+	run      func(t *testing.T)
+	fixtures []fixture
+}
+
+func TestGraphQL(t *testing.T) {
+	tests := []testCase{
+		{
+			title:    "test user API",
+			run:      testGraphQL_User,
+			fixtures: []fixture{useDefaultEnv, usePostgres, useRedis},
```
jarrel-b, line 46: The idea here is that child tests under `testGraphQL_User` will run using the same environment, and also child tests themselves may optionally also use their own fixtures.

graphql/graphql_test.go
```diff
@@ -0,0 +1,592 @@
...
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.title, withSetup(test.run, test.fixtures...))
+	}
+}
+
+func testGraphQL_User(t *testing.T) {
+	tests := []testCase{
+		{title: "should create a user", run: testCreateUser(newNonceFixture{})},
```
jarrel-b, line 56: Tests may also specify their own fixtures that get called when the test is executed, this fixture for example returns a  test wallet and nonce.

graphql/graphql_test.go
```diff
@@ -0,0 +1,592 @@
...
+		)
+
+		require.Empty(t, response.CreateUser.Message)
+		assert.Equal(t, username, *response.CreateUser.Viewer.User.Username)
+	}
+}
+
+func testUserByUsername(userF newUserFixture) func(t *testing.T) {
+	return func(t *testing.T) {
+		userF.setup(t)
```
jarrel-b, line 101: I formatted the tests in the `GivenWhenThen` structure:
```go
func() test {
  Given - initial test setup
  // line break
  When - the piece of code under tests
  // line break
  Then - all the assertions and checks
}
```
So it should be super clear what is the piece of code that is being tested

radazen (approved):
This is great, Jarrel! I love how clean everything feels. The refactoring and clean-up of related parts of the codebase are also 🔥. I left a few comments, pretty much all in the spirit of "can we reduce friction/boilerplate even more?", but I think the PR itself is solid as-is!

graphql/graphql_test.go
```diff
@@ -0,0 +1,722 @@
...
+	}
+}
+
+func testAddWallet(userF newUserFixture) func(t *testing.T) {
+	return func(t *testing.T) {
+		userF.setup(t)
+		c := defaultClient()
+		walletToAdd := newWallet(t)
+		nonce := newNonce(t, c, walletToAdd)
+		var response = struct {
```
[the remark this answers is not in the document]
jarrel-b: @tbezman found a neat golang library that autogenerates both the input and response types which looks promising so far, so we might not need this at all!

graphql/fixtures_test.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/mikeydub/go-gallery/service/persist/postgres"
+	"github.com/mikeydub/go-gallery/service/task"
+	"github.com/stretchr/testify/require"
+)
+
+// fixture runs setup and teardown for a test
+type fixture func(t *testing.T)
+
+// withFixtures sets up each fixture before running the test
+func withFixtures(test func(t *testing.T), fixtures ...fixture) func(t *testing.T) {
```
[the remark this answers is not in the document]
jarrel-b: I like it!

graphql/graphql_test.go
```diff
@@ -0,0 +1,722 @@
...
+			}
+		}{}
+
+		post(t, c, ops.Op("addUserWalletMutation"), &response,
+			withJWT(newJWT(t, userF.id)),
+			client.Var("chainAddress", map[string]string{
+				"address": walletToAdd.address,
+				"chain":   "Ethereum",
+			}),
+			client.Var(
```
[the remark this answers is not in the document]
jarrel-b: Hopefully this goes away as well

graphql/graphql_test.go
jarrel-b: Yea good point! The idea was that all the fields are meant to be output only fields that are set when `setup` is called. I liked the idea of seeing which fixtures a test uses by looking at the function signature, but after writing more tests it became cumbersome to update the function args each time I needed to add a fixture to the test, so creating them inside the test themselves is much simpler.

PR #748, opened 2023-01-03
Change subscription expiration from 31 days to 3 days

jarrel-b:
Background:
Whenever a backend instance starts, it creates a new subscription to the notification pubsub topics so that it can receive notifications published by other instances. However, there is a quota limit of 10,000 subscriptions per project: https://cloud.google.com/pubsub/quotas. I noticed that the dev backend was failing to start because we had already exceeded the quota for that project. 

This PR reduces the expiration of a subscription from 31 days to 3 days. This means that if a subscriber hasn't consumed from that subscription in 3 days, that subscription will be deleted. In the future, I think we can look into re-using subscriptions or cleaning up the subscription when the server spins down.


tbezman (approved):
makes sense to me!

radazen (approved):
🕵️🐛

PR #750, opened 2023-01-06
Add pubsub emulator and cleanup backend setup

jarrel-b:
This PR removes the nginx container from the `docker-compose.yml` file. We previously needed nginx to proxy AppEngine tasks from the CloudTasks emulator to services running locally, but since we don't use AppEngine we no longer need it. The PR also removes the `cloud-task-emulator` submodule from the repo. Instead, the emulator is started when `docker compose` is run. 

The PR also adds a PubSub emulator to the `docker-compose.yml` which is used when the backend is run locally. I also updated `_encrypted_local` to configure the local env vars to use the emulator. I also updated the subscription behavior to create the topic if it doesn't already exist. This is more a convenience for local development, but this behavior looks to already be the default behavior in some other clients.

The PR also adds some new functionality to the tests:
* `useNotificationTopics` fixture which starts a running PubSub emulator and notification topics
* `serverFixture` which starts a live test server for end to end tests (will be useful for testing notifications)


README.md
```diff
@@ -124,64 +124,6 @@ This is available for live environments:
 ```bash
 $ curl api.gallery.so/alive
 ```
-
-## Testing services via Cloud Tasks
```
radazen, line 128: I'm a little sad to see this wonderful diagram go

radazen (approved):
👍 👍 👍

Robinnnnn (approved):
Such a legend

PR #753, opened 2023-01-09
Use CTE and skip "is null" check to speed up gallery_id migration

radazen:
On my local database, reworking this SQL migration statement takes it from ~40 seconds to ~300ms, which is fast enough to run without scheduling downtime! The second `gallery_id` migration statement takes about 3s still, but I think that's fine for our purposes.


benny-conn (approved):
Wow!

PR #754, opened 2023-01-10
Retool add wallet mutation

jarrel-b:
This PR adds a private authenticator which rotates the nonce for the user and then adds the wallet (but does not verify anything e.g. a signature), but otherwise goes through the same validation a user typically would.  The mutation is decorated with the `retool` directive so that it can only be called by retool, and the method itself performs does the same authorization check so that it can't be called inadvertently. 

It also fixes a graphQL serialization bug I noticed during testing where it couldn't marshal the `GnosisSafe` enum type correctly.


publicapi/admin.go
```diff
@@ -74,3 +76,49 @@ func (api *AdminAPI) RemoveRolesFromUser(ctx context.Context, username string, r
 
 	return &user, err
 }
+
+type authenticator struct {
```
radazen, line 80: Given that we're in the `publicapi` package, this `authenticator` type will be accessible to the entire public API, right? I'm concerned that it might be too easy to accidentally use/misuse it in non-admin contexts if it's available to the the rest of the API.

It's a bit of an annoying problem, because it'd be cool to move admin stuff into an `adminapi` package to avoid misuse of private functionality, but then admin functions wouldn't have access to private `publicapi` helpers, which are also useful. Long-term, maybe the right path is to extract common helpers from `publicapi` into another package that both an `adminapi` and `publicapi` can use...but for now, maybe we just define our private authenticator type _inside_ of `AddWalletTouserUnchecked` so it's not available for anything else to use?

publicapi/admin.go
```diff
@@ -74,3 +76,49 @@ func (api *AdminAPI) RemoveRolesFromUser(ctx context.Context, username string, r
...
+	authMethod func(context.Context) (*auth.AuthResult, error)
+}
+
+func (a authenticator) GetDescription() string { return "" }
+func (a authenticator) Authenticate(ctx context.Context) (*auth.AuthResult, error) {
+	return a.authMethod(ctx)
+}
+
+func (api *AdminAPI) AddWalletToUserUnchecked(ctx context.Context, username string, chainAddress persist.ChainAddress, walletType persist.WalletType) error {
+	if err := auth.RetoolAuthorized(ctx); err != nil {
```
radazen, line 90: Idea (that I wish I had thought of when adding the `@authRequired` directive): in the style of other golang `MustBlahBlah` functions that panic if their criteria isn't satisfied, what about something like `RequireRetoolAuthorization(ctx)` that panics if the caller isn't authed?

Pro: less room for error! The current flow requires each admin method to implement the error checking correctly, which makes it more likely that a typo could open up a security hole. This same flaw exists in @authRequired methods today!

Con: a panic isn't the ideal way to handle an expected error

But actually, I don't think the con matters, because the GraphQL handler takes care of graceful error handling. We should expect that callers will be rejected by the GraphQL retool directive if they aren't authorized. If they _still_ make it here somehow, I think a panic is totally reasonable.

radazen (approved):
Looks good to me! Left a couple of comments/requests.

publicapi/admin.go
jarrel-b: I like it!

publicapi/admin.go
jarrel-b: Ohh yes, I actually originally implemented it where the `authenticator` was scoped only in `AddWalletTouserUnchecked`. Problem was that I don't think there's a way in go to declare methods within a function unfortunately. It doesn't look like there any `publicapi` helpers that the admin api uses atm, so I think I'll move it over to its own `adminapi`. You think at that this point it makes sense to have another route just for admin functionality, i.e. `/glry/admin/query`?

radazen: > Problem was that I don't think there's a way in go to declare methods within a function unfortunately.

Doh! Yeah, makes sense that you can't define receiver methods inside another function. Bummer.

> You think at that this point it makes sense to have another route just for admin functionality, i.e. something like/glry/admin/query?

I'd _really_ like to separate admin stuff out into a separate endpoint, but I don't think there's an easy way to do it with gqlgen. I've tried to reason through it like this:
- Split our schema into multiple schemas (public and admin)
- Serve only the public stuff at one endpoint, and serve both at an `admin` endpoint
- gqlgen serves everything from every schema its config file points to, so in addition to splitting up the schemas, we'd need multiple gqlgen config files (one that pulls in public + admin schemas, one that only pulls in public)
- It doesn't seem like gqlgen is set up to handle multiple config files, and in the best case, we'd still probably have 2 config files to maintain (with both containing the same mappings from generated types to our own internal types)

But it'd be awesome if we could make it work somehow, because serving admin stuff at a separate route would also let us lock that entire route down with GCP auth (maybe Retool has a GCP service key it uses or something), and we'd still be able to use the retool auth stuff on top of that for extra security. And it'd be great not to have admin functions in our public schema!

jarrel-b:
@radazen mind taking another pass? I moved the `publicapi/admin.go` into a package named `adminapi`. Realized that the `adminapi` also uses the validation helpers in `publicapi` so I moved those helpers into to a new package called `inputcheck` (open to other names 😄  )

radazen (approved):
👍 👍 👍 

Discussion thoughts:

- At the point where we're moving validation methods to a separate package, would it make sense to move them into our `validate` package? I'm on the fence; it feels like it should go there, but `validate` has a lot going on already and one of the nice things about having a private `validateFields` method in `publicapi` was that it was obvious what method you should use to validate stuff.

- I was initially wondering if `AdminAPI` should still be part of the `PublicAPI` struct, or if it should be its own thing now since it's in its own package (with its own `adminapi.For(ctx)` etc). But after thinking about it for a while, imho it does still belong in the `PublicAPI` struct, because it's part of our public API surface and accessible to anyone who hits our GraphQL query endpoint. TL;DR: looks great, I think it's correct the way it is, just thinking out loud!

jarrel-b:
> 👍 👍 👍
> 
> Discussion thoughts:
> 
> * At the point where we're moving validation methods to a separate package, would it make sense to move them into our `validate` package? I'm on the fence; it feels like it should go there, but `validate` has a lot going on already and one of the nice things about having a private `validateFields` method in `publicapi` was that it was obvious what method you should use to validate API input parameters.
> * I was initially wondering if `AdminAPI` should still be part of the `PublicAPI` struct, or if it should be its own thing now since it's in its own package (with its own `adminapi.For(ctx)` etc). But after thinking about it for a while, imho it does still belong in the `PublicAPI` struct, because it's part of our public API surface and accessible to anyone who hits our GraphQL query endpoint. TL;DR: looks great, I think it's correct the way it is, just thinking out loud!

had pretty much the same line of thinking: originally I just wanted to exports the validation methods in `publicapi`, but that would've led to an import cycle. Although `AdminAPI` having its own struct would've resolve it, I felt that it didn't quite merit its own api yet. I was a bit back and forth on moving it to `validate` for the reason you mentioned.  But now that I think about it if something should really belong in the package, it probably should just go in there even if theres a lot in the namespace already. And I also don't know what else we would use `inputcheck` for besides a bridge to fix the imports since everything in `validate` are all technically input checks lol

PR #755, opened 2023-01-11
primary wallet set and get

benny-conn:
Changes:

- **Added** a new column called `primary_wallet_id` to the users table
- **Added** a query field on the `GalleryUser` type for this wallet
- **Added** a mutation for setting this field


db/migrations/core/000045_primary_wallet.up.sql
```diff
@@ -0,0 +1,6 @@
+alter table users add column primary_wallet_id varchar(255) references wallets(id);
+
+-- every user has at least 1 wallet right?
```
benny-conn, line 3: I can check this tomorrow morning, not in a place where I can connect to dev currently and check

Robinnnnn (approved):
thank you ser!!!

db/migrations/core/000045_primary_wallet.up.sql
radazen: Looks like there are 42 users with empty `wallets` arrays, but all of those users are deleted. There are also 7 users with `wallets` arrays that reference wallet IDs that don't actually exist, but all of those users are deleted too. If it makes things easier, imho we could just hard delete all 49 of those users so they don't violate any foreign key constraints or screw up any migrations.

db/queries/core/query.sql
```diff
@@ -601,6 +601,9 @@ update users set email_verified = 0 where users.id = @user_id;
 -- name: UpdateUserEmailUnsubscriptions :exec
 UPDATE users SET email_unsubscriptions = $2 WHERE id = $1;
 
+-- name: UpdateUserPrimaryWallet :exec
+update users set primary_wallet_id = @wallet_id where users.id = @user_id and (select exists(select 1 from wallets where wallets.id = @wallet_id and wallets.owner_user_id = @user_id and wallets.deleted = false));
```
radazen, line 605: I think using `from` syntax here would be faster and more idiomatic:

```
update users set primary_wallet_id = @wallet_id from wallets
    where users.id = @user_id and wallets.id = @wallet_id
    and wallets.owner_user_id = @user_id and wallets.deleted = false;
```

I tried running both the `from` and `select` versions to test that assumption, but as it turns out, we don't have a `wallets.owner_user_id` column! So I think we just have to check whether the wallet is part of the user's wallets array:

```
update users set primary_wallet_id = @wallet_id from wallets
    where users.id = @user_id and wallets.id = @wallet_id
    and wallets.id = any(users.wallets) and wallets.deleted = false;
```


With that last version, I did test `from` vs sub-`select`, and the `from` version is about 3x faster according to `EXPLAIN ANALYZE`. Of course, both of them execute in milliseconds, so the difference doesn't matter much in practice.

radazen (approved):
Looks great! I think the SQL statement for updating a user's primary wallet will need to be tweaked before merging, but everything is 👍 👍  otherwise!

radazen (commented):
Doh, thought of another question immediately after submitting that last review: what should we do if the user tries to remove a wallet that happens to be their primary wallet? Disallow it and make them pick a new primary wallet first, or let them do it and pick one of their remaining wallets as primary?

PR #758, opened 2023-01-11
Refactor external ID

jarrel-b:
This PR removes some copy and pasted code to check if an IP is private, in favor of the `ip.IsPrivate` method that's available on the `net/http` package. It also simplifies the `getExternalID` function to use the gin context's `ClientIP` method that I just found out about which will also try to find the forwarded client IP.


benny-conn (approved):
Beautiful, I didn't even know an IP could be private

PR #761, opened 2023-01-13
Add trending users

jarrel-b:
This PR adds a new query, `trendingUsers`, which returns the top 10 trending users based on gallery views. It uses redis as a side cache to store the trending users: if there is a cache miss, the app will fetch the results from the database, update redis, and then return the response.

**Considerations**
* If there's a cache miss like an hour from when the key was inserted and a lot of users land on the feed at once, we could be making a lot of redundant calls to the database simultaneously since every client will try to populate the cache. I think this is OK given our current traffic and that the queries themselves aren't that expensive vs the added complexity of a distributed lock.

**TODOs**
* [x] Handle the case where a trending user is deleted, or if they change their name


graphql/graphql_test.go
```diff
@@ -242,18 +244,56 @@ func testViewsAreRolledUp(t *testing.T) {
...
 	clientB := authedServerClient(t, serverF.server.URL, viewerB.id)
-	resp, err = viewGalleryMutation(context.Background(), clientB, userF.galleryID)
-	_ = (*resp.ViewGallery).(*viewGalleryMutationViewGalleryViewGalleryPayload)
-	require.NoError(t, err)
+	viewGallery(t, clientB, userF.galleryID)
 
 	// TODO: Actually verify that the views get rolled up
 }
 
+func testTrendingUsers(t *testing.T) {
```
Robinnnnn, line 255: 😍

publicapi/feed.go
```diff
@@ -208,6 +213,46 @@ func (api FeedAPI) PaginateGlobalFeed(ctx context.Context, before *string, after
...
+		return users, nil
+	case err != nil:
+		return nil, err
+	default:
+		err := json.Unmarshal(byt, &users)
+		if err != nil {
+			return nil, err
+		}
+		return users, nil
+	}
```
Robinnnnn, line 253: if there _is_ a cache hit, we still want to check if it's stale, and if so kick off an async update right? sorry if this is already handled.

graphql/graphql_test.go
Robinnnnn: a nice TODO might be to test the cache behavior

Robinnnnn (approved):
looks incredible, thank you jarrel!!

> I think this is OK given our current traffic and that the queries themselves aren't that expensive vs the added complexity of a distributed lock.

agreed with the traffic comment, but let's make sure we can keep an eye on it! would be great to be able to see how often the heavy queries are occurring

graphql/graphql_test.go
jarrel-b: good idea, ill add a test for this

graphql/model/models.go
```diff
@@ -119,3 +121,40 @@ type ErrInvalidIDType struct {
...
 }
+
+type Window struct {
+	time.Duration
+	Name string
+}
+
+var (
+	lastSevenDaysWindow = Window{7 * 24 * time.Hour, "LAST_7_DAYS"}
+	allTimeWindow       = Window{1<<63 - 1, "ALL_TIME"}
```
jarrel-b, line 132: This would be ~68 years, but thinking maybe a conditional statement bases the report type in the sql query would be easier to read

Robinnnnn (approved):

PR #768, opened 2023-01-20
Use sops to manage secrets

radazen:
_**Note: Encrypted files make up the majority of this PR. You can skip them while reviewing!**_

## Why sops?
As of this PR, we'll be managing secrets with [sops](https://github.com/mozilla/sops) instead of using an encrypted zip file. Why?

- sops is more polished and battle-tested than rolling our own secrets scripts
- sops understands common config file formats (yaml, json, env) and can encrypt _values_ without encrypting keys
  - changes to secrets will be more likely to merge cleanly and we'll get meaningful diffs to review
- sops can use different key management backends
  - in our implementation, we're using GCP's KMS service, but we could switch to local keys again in the future if we want to
- _sops lets us keep all secrets encrypted on-disk at all times_
  - instead of keeping secrets decrypted in a .gitignored directory, they stay encrypted on disk and get decrypted in memory when we want to use them

## How do I use it?
- To view or edit files: `sops <filename>`
  - This opens whatever `$(EDITOR)` is set to in your environment, or vim if not set
  - The file gets decrypted and opened in your editor. When you save your changes and exit, the content gets re-encrypted and written back to disk

## How does our infrastructure use it?
- sops is written in go and has an accompanying decryption library that can be used directly from go code
  - we can easily read encrypted files from disk in contexts where we had been reading unencrypted files
- sops has a command-line argument to create a temporary decrypted file, invoke another command, and then remove the temporary file
  - we can use this in contexts like deployment, where we need to pass an --env-file argument

## What's actually changed?
- `_deploy` and `_local` have been removed
- we now have a `secrets` directory that gets committed to the repository
- each subdirectory of `secrets` has its own associated GCP KMS keyring, such that access to `secrets/dev` does not imply access to `secrets/prod` (or vice versa)
- files for local development are now located under the appropriate parent directory:
  - config files for developing locally against `dev` services are in `secrets/dev/local`
  - config files for developing locally against `prod` services are in `secrets/prod/local`
  - config files for purely local development are in the confusingly named `secrets/local/local`
- everything from `_deploy/Makefile` has been moved into our root `Makefile`
  - the root Makefile invokes `sops` to include secrets where necessary
- service keys, config files, etc use new utility methods to load encrypted files from disk

## Cloud SQL Proxy
This PR also includes support for Cloud SQL Proxy!

- Cloud SQL Proxy runs locally on your machine and uses your gcloud credentials to open a secure tunnel between your machine and our SQL instances
- It does not require allowlisting IP addresses or SSL certificates
- _**Everyone should use this for when connecting directly to our SQL instances. We should not distribute certificates or allowlist IP addresses for day-to-day access.**_
  - Local config files have been updated to point to `localhost` ports, assuming Cloud SQL Proxy is in use
- To run it, just: `make start-sql-proxy`

## What's left to do?
- More documentation:
  - `brew install sops` in our root README
  - A `secrets/README` file explaining how everything works
  - Better explanation of how `sops` works in our root Makefile
  - Make sure updated Github Actions work

## Considerations:
- I think `secrets/local/local` looks dumb, but I do like keeping the pattern of `secrets/target-env/current-env`
- I'm not sure if `local` secrets is a concept we need anymore. Many services use the `service-key-dev.json` service account when `ENV == local`, which makes me wonder if it's possible to develop locally _without_ access to our dev services.
- I had to be "creative" with our `Makefile` to get it to work with sops. It's possible that `make` isn't the correct tool for the job at this point, and maybe we should switch to something else for our deployments in the future.


docker/cloud_sql_proxy/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+      - "6545:6545"
+    volumes:
+      - ~/.config/gcloud/application_default_credentials.json:/config
+    command: >
+      --address=0.0.0.0 --credentials-file=/config
+      '${SQL_CONNECTION_NAME_PROD_DB}?port=6543'
+      '${SQL_CONNECTION_NAME_PROD_DB_REPLICA}?port=6544'
+      '${SQL_CONNECTION_NAME_PROD_ETH_INDEXER}?port=6545'
+
+  cloud-sql-proxy-dev:
```
radazen, line 25: I split these into two separate services so someone who only has access to `dev` services can still use the proxy.

indexer/cmd/root.go
```diff
@@ -44,12 +43,6 @@ var rootCmd = &cobra.Command{
 	Long: `An NFT indexer lovingly built by your friends at Gallery.
                 Source code is available at https://github.com/gallery-so/go-gallery.`,
 	Args: func(cmd *cobra.Command, args []string) error {
-		switch manualEnv {
```
radazen, line 47: I looked through the code and found that we ultimately do not use these service keys in the indexer anymore, so I removed them.

Makefile
```diff
@@ -1,3 +1,313 @@
...
+$(DEPLOY)-%-feedbot                    : PORT           := 4123
+$(DEPLOY)-%-feedbot                    : TIMEOUT        := $(FEEDBOT_TIMEOUT)
+$(DEPLOY)-%-feedbot                    : CPU            := $(FEEDBOT_CPU)
+$(DEPLOY)-%-feedbot                    : MEMORY         := $(FEEDBOT_MEMORY)
+$(DEPLOY)-%-feedbot                    : CONCURRENCY    := $(FEEDBOT_CONCURRENCY)
+$(DEPLOY)-$(DEV)-feedbot               : SERVICE        := feedbot-dev
+$(DEPLOY)-$(PROD)-feedbot              : SERVICE        := feedbot
+
+# Service name mappings
+$(PROMOTE)-%-backend                   : SERVICE := default
```
jarrel-b, line 155: feel free to ignore: this should probably be `backend` now that it no longer runs on App Engine, but we actually don't use these commands in our deployments anymore

Makefile
```diff
@@ -1,3 +1,313 @@
+#####
```
jarrel-b, line 1: Very happy I don't need to `cd` into `_deploy` anymore!

Makefile
```diff
@@ -52,3 +364,46 @@ format-graphql:
...
+
+migrate-prod-coredb: start-prod-sql-proxy confirm-prod-migrate
+	migrate -path ./db/migrations/core -database "postgresql://$(POSTGRES_USER):$(POSTGRES_PASSWORD)@localhost:6543/$(POSTGRES_DB)?sslmode=disable" up
+
+
+#----------------------------------------------------------------
+# End of targets
+#----------------------------------------------------------------
+
+endif
```
jarrel-b, line 409: Could you help me understand why do the targets need to be defined inside the `else` statement again?

jarrel-b (approved):
Looks great to me! And I was able to setup everything super easily. I want to spend some time later to understand the make features that get used, but don't want to block the PR due to my unfamiliarity with it. Excited that we'll never have to deal with those encrypted files ever again 🙌

Makefile
radazen: I'll add a better write-up to the makefile, but the gist is that we need to recursively call `make` in order to read encrypted variables into the environment. The if/else/endif stuff is basically doing this:

```
if we haven't exported REQUIRED_SOPS_FILES yet...
    %: <-- the only target defined is a wildcard; all targets go through this path
        export REQUIRED_SOPS_FILES
        call make again

else if there are still SOPS files to import:
    %: <-- same as above, this catches all targets
        sops exec-env make <-- use sops to call make with secrets added to environment

else if we've finished all of the above:
    actual targets go here
```

We want a single target (e.g. `make deploy-prod-backend`) to do different things on different invocations of `make` depending on what "import sops stuff" step we're on, so the `if` statements control what targets are actually defined for each step.

Makefile
radazen: Gotcha. I'm all for deleting unnecessary stuff!

PR #769, opened 2023-01-20
Add hot feed

jarrel-b:
This PR adds a new query `trendingFeed` which returns a `FeedConnection` to trending feed events. How we find what's trending is currently very simple: we count the number of interactions each feed event has had in a rolling 72 hour window and return the top 100 IDs. This is essentially an edge rank with no biases added yet, we plan to add weights once we see emerging behavior come up from the feed.

After computing the result, we cache the trending feed event IDs in redis. To handle edge cases that can arise from paging through the ranking with an old cursor, the cursor returned by `trendingFeed` actually encodes the ranking itself. This is feasible since `trendingFeed` always returns a fixed result set. The cursor is encoded as:
```
<current position><the number of ids encoded><id1><id2>....<idN>
```
This is nice since each client will always get a consistent view of whats trending, there is no possibility of duplicates and missing events, and there's no need to query the cache once a client has a cursor.

**Post Merge TODOs**
* [ ] Run migration to add new index for feed interaction queries
* [ ] Run migration to add new index for collection owner queries

### DIFF
--- a/db/queries/core/query.sql
+++ b/db/queries/core/query.sql
@@ -697,10 +697,12 @@ update collections set nfts = @nfts, last_updated = now() where id = @id and del
 insert into collections (id, version, name, collectors_note, owner_user_id, gallery_id, layout, nfts, hidden, token_settings, created_at, last_updated) values (@id, 0, @name, @collectors_note, @owner_user_id, @gallery_id, @layout, @nfts, @hidden, @token_settings, now(), now()) returning id;
 
 -- name: GetTrendingUserIDs :many
-with rollup as (select e.gallery_id, count(*) view_count from events e where action = 'ViewedGallery' and e.created_At >= @window_end group by e.gallery_id)
+with rollup as (
+	select e.gallery_id, count(*) view_count from events e where action = 'ViewedGallery' and e.created_At >= @window_end group by e.gallery_id
+)
 select p.id from (
 	select u.id, row_number() over(order by sum(view_count) desc, max(u.created_at) desc) as position
-	from galleries g, users u, rollup r
+	from rollup r, galleries g, users u
 	where r.gallery_id = g.id and g.owner_user_id = u.id and u.deleted = false and g.deleted = false
 	group by u.id
 ) p
--- a/db/queries/core/query.sql
+++ b/db/queries/core/query.sql
@@ -365,6 +365,14 @@ SELECT * FROM feed_events WHERE owner_id = sqlc.arg('owner_id') AND deleted = fa
             CASE WHEN NOT sqlc.arg('paging_forward')::bool THEN (event_time, id) END DESC
     LIMIT sqlc.arg('limit');
 
+-- name: PaginateTrendingFeed :many
+select f.* from feed_events f join unnest(@feed_event_ids::text[]) with ordinality t(id, pos) using(id) where f.deleted = false
+  and t.pos < @cur_before_pos::int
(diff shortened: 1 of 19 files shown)


db/queries/core/query.sql
```diff
@@ -699,3 +707,10 @@ select p.id from (
 	group by u.id
 ) p
 where position <= @size::int;
+
+-- name: GetTrendingUsersByIDs :many
+select users.* from users join unnest(@user_ids::varchar[]) with ordinality t(id, pos) using (id) where deleted = false order by t.pos asc;
```
radazen, line 712: I really like this syntax! A while back, I added `GetContractsByIDs` as a faster alternative to pgx batching (since we end up looking up thousands of contracts by their IDs, seemed easier to put them in a single query than have thousands of batched queries executing in sequence, even if batching does speed them up!). But I did it like this:

`select * from contracts where id = any(@contract_ids) and deleted = false;`

And then I used a map to manually order results on the backend, ensuring that the output contract results were in the same order as the input contract IDs.

Your syntax achieves that same result much more elegantly, without having to sort anything on the backend, and the performance appears to be the same. I think this is definitely the pattern we should adopt for looking up many rows by their IDs in the future!

graphql/fixtures_test.go
```diff
@@ -193,18 +196,61 @@ type userWithTokensFixture struct {
...
+	tokenIDs := syncTokens(t, ctx, c, user.id)
 	return userWithTokensFixture{user, tokenIDs}
 }
 
+type userWithFeedEventsFixture struct {
+	userWithTokensFixture
+	feedEventIDs []persist.DBID
+}
+
+// newUserWithTokensFixture generates a new user with feed events pre-generated
```
radazen, line 218: `newUserWithTokensFixture` -> `newUserWithFeedEventsFixture`?

graphql/resolver/schema.resolvers.go
```diff
@@ -1339,6 +1338,23 @@ func (r *queryResolver) GlobalFeed(ctx context.Context, before *string, after *s
 	}, nil
 }
 
+func (r *queryResolver) TrendingFeed(ctx context.Context, before *string, after *string, first *int, last *int) (*model.FeedConnection, error) {
```
radazen, line 1341: It's awesome that existing feed stuff (`eventsToFeedEdges`, `model.FeedConnection`) can be reused for the trending feed this easily.

publicapi/feed.go
```diff
@@ -213,44 +213,124 @@ func (api FeedAPI) PaginateGlobalFeed(ctx context.Context, before *string, after
...
+func (api FeedAPI) TrendingUsers(ctx context.Context, report model.Window) ([]db.User, error) {
+	calcFunc := func(ctx context.Context) ([]persist.DBID, error) {
+		return api.queries.GetTrendingUserIDs(ctx, db.GetTrendingUserIDsParams{
+			WindowEnd: time.Now().Add(-time.Duration(report.Duration)),
+			Size:      20,
+		})
+	}
+
+	ttl := time.Hour
+	if report.Duration > 7*24*time.Hour {
```
radazen, line 310: A comment here could be helpful! If I'm reading it correctly: we want a cache TTL of 1 hour if we're calculating trending users over the past week or less, but a cache TTL of 1 day if we're calculating trending users over a longer period of time?

jarrel-b: Good call, and thats correct!

PR #774, opened 2023-01-24
first pass at boolean bucket

benny-conn:
Changes:

- **Added** `user_experiences` to the graph QL schema and DB schema and added resolvers and mutation for updating and reading them for a `viewer`

Considerations:

- Mentioned this in slack but something feels weird about returning experiences as an array to the graphQL schema because then the frontend has to sift through the array to get the type it wants. Should it just be a giant type with the correct field names so the frontend can just resolve as needed? For example:
```
type UserExperiences {
  multiGalleryBanner: Boolean
  tutorialPassed: Boolean
  ...
}
```


jarrel-b (approved):
looks good to me!

PR #775, opened 2023-01-24
Add sandbox env

jarrel-b:
This PR brings back up [api.sandbox.gallery.so](https://api.sandbox.gallery.so) for the sandbox.

**Changes**
* Added a `sandbox.yml` workflow which currently only deploys the backend to the sandbox env. This workflow can be adhoc and a branch can be provided to deploy that specific branch.
* Moved the CI build dependencies (sentry, sops, gcloud) to a separate action `build-deps` which gets called in our workflows
* Added `ETH_PRIVATE_KEY` to our secret vars to replace the secrets managed by Secrets Manager
* Updates `backend-sandbox-env.yaml` which values from `backend-env.yaml`

**Post Merge Tasks** 
*  [ ] When deploying to `dev` and `prod` a new revision on Cloud Run will first need to be manually added thats removes the Secrets Manager connection. Cloud Run will complain that the env var is being set in two places otherwise. This revision will fail to startup because it's missing the env var, but Cloud Run won't move any traffic to it anyways.
* [ ] Delete the secret from Secrets Manager so it isn't in two places

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ migrate-prod-coredb  : REQUIRED_SOPS_SECRETS := $(SOPS_PROD_SECRETS)
 
 # Environment-specific settings
 $(DEPLOY)-$(DEV)-%                : ENV                    := $(DEV)
-$(DEPLOY)-$(SANDBOX)-%            : ENV                    := $(DEV)
+$(DEPLOY)-$(SANDBOX)-%            : ENV                    := $(SANDBOX)
 $(DEPLOY)-$(PROD)-%               : ENV                    := $(PROD)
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -48,7 +48,7 @@ migrate-prod-coredb  : REQUIRED_SOPS_SECRETS := $(SOPS_PROD_SECRETS)
 
 # Environment-specific settings
 $(DEPLOY)-$(DEV)-%                : ENV                    := $(DEV)
-$(DEPLOY)-$(SANDBOX)-%            : ENV                    := $(DEV)
+$(DEPLOY)-$(SANDBOX)-%            : ENV                    := $(SANDBOX)
```
jarrel-b, line 51: Changed this to `$(SANDBOX)` so its flagged on Sentry as the `sandbox` env

Robinnnnn (approved):
🤌

PR #779, opened 2023-01-26
Rotate auth JWTs

radazen:
It's time to rotate the JWTs we use for auth tokens! Since invalidating them immediately would log users out, this PR uses a transitional approach that checks for a JWT_SECRET_ROTATING_OUT variable, too.

How it works:
- new tokens are issued with JWT_SECRET
- existing tokens are validated with JWT_SECRET
  - if validation fails, existing tokens are validated with JWT_SECRET_ROTATING_OUT

We can remove all of the rotation stuff next week. Our JWT TTL on prod was set to 3 days (again, somehow!), so it won't take long for all the old tokens to expire.

Follow-ups: the rotation code I added for email JWTs is a little messy, but I'd like to see if we can clean everything up with generics and consolidate our JWT functions (as per the comment at the top of the emails JWT file!)

### DIFF
--- a/emails/jwt.go
+++ b/emails/jwt.go
@@ -2,6 +2,7 @@ package emails
 
 import (
 	"errors"
+	"strings"
 	"time"
 
 	"github.com/dgrijalva/jwt-go"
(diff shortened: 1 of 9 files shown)


jarrel-b (approved):
🚢 🚢 🚢

PR #782, opened 2023-01-27
Use correct agg function in trending query

jarrel-b:
The old query was double counting legacy views because the granularity of the legacy views table is lower than the rollup CTE. In this case, we should use `min` or `max` or `avg` instead of `sum`.


Robinnnnn (approved):

radazen (approved):
Well that was fast!

PR #784, opened 2023-01-31
initial pass for twitter auth

benny-conn:
Changes:

- **Added** a new `external_socials` jsonb column to the users pii table
- **Added** a new `social_account_auth` table for storing access tokens and the like, currently supports the OAuth pattern can be extended to support different types of tokens in the future
- **Added** resolvers for connecting social accounts and resolving them
- **Added** a `socialauth` package for various social media authenticator implementations
- **Added** a `twitter` package for making calls to the twitter API (including authenticating with the API)
- **Added** a `SocialsAPI` to the `PublicAPI` for interacting with the various socials, for now just for creating the twitter authenticator


Robinnnnn (commented):
thanks for crushing this benny! do you think we could include some basic tests?

service/twitter/twitter.go
```diff
@@ -0,0 +1,153 @@
...
+		err = util.GetErrFromResp(accessResp)
+		return AccessTokenResponse{}, fmt.Errorf("failed to get access token, returned status: %s", err)
+	}
+
+	var accessToken AccessTokenResponse
+	if err := json.NewDecoder(accessResp.Body).Decode(&accessToken); err != nil {
+		return AccessTokenResponse{}, err
+	}
+
+	err = a.queries.UpsertSocialMediaOAuth(ctx, coredb.UpsertSocialMediaOAuthParams{
```
radazen, line 109: Is this the only place we use the `queries` object in here? It feels a little deceptive for a function called `get...` to write to the database. If this is the only database interaction in the Twitter service, maybe we can move it to the calling method (something in the public API?) and write the token info there?

db/migrations/core/000059_social_accounts.up.sql
```diff
@@ -0,0 +1,26 @@
...
+    last_updated timestamp with time zone default CURRENT_TIMESTAMP not null,
+    user_id varchar(255) not null references users(id),
+    provider varchar not null,
+    access_token varchar,
+    refresh_token varchar
+);
+
+grant select on pii.social_account_auth to access_ro_pii;
+grant all on pii.social_account_auth to access_rw_pii;
+alter table pii.social_account_auth owner to access_rw_pii;
```
radazen, line 24: We should be able to omit the grant/alter statements here; everything gets handled automatically as long as we `set role to access_rw_pii;` (and we'd get a permission denied error if we didn't!)

radazen (commented):
🔥 🔥 🔥

radazen (approved):
🙌🙌🙌

Awesome! Looking great, Benny. Thanks for following up on comments so quickly, too!

PR #786, opened 2023-02-01
Rotate magic link keys

radazen:
This PR rotates the keys for our magic links, and prod and dev/local now have separate keys.


jarrel-b (approved):
👍

PR #787, opened 2023-02-01
Postgres Permissions

radazen:
## What's new?
- New non-login postgres roles:
  - access_ro
  - access_rw
  - access_ro_pii
  - access_rw_pii

- New login postgres roles:
  - gallery_backend (has access_rw, access_rw_pii)
  - gallery_migrator (has access_rw, access_rw_pii)
  - gallery_team_ro (has access_ro)
  - gallery_team_rw (has access_rw)
  - gallery_team_ro_pii (has access_ro, access_ro_pii)
  - gallery_team_rw_pii (has access_rw, access_rw_pii)

## Why?
We want to be proactive about protecting our users' data! Using a single `postgres` superuser account for everything isn't sustainable as we continue to grow.

## Implementation Details
Up to this point, PII has been stored in `pii_`-prefixed tables in the `public` scheme, and we've managed access control on a per-table basis. This PR moves those tables to a new schema, `pii`, which further separates them from non-pii data and allows us to implement access permissions on a schema-wide basis.

The `public` schema is now owned by `access_rw`. Both `access_rw` and `access_rw_pii` may write data in the public schema, but only `access_rw` may create new objects (tables, views, etc) there. This ensures that `access_rw` will be the _owner_ of everything in the public schema, and will have permissions to alter/drop/manage anything in that schema.

The `pii` schema is owned by `access_rw_pii`, and only `access_rw_pii` may write data or create objects in the `pii` schema. As above, this ensures that `access_rw_pii` will be the owner of everything in the `pii` schema, and can manage all objects therein.

Migrations will be run with the `gallery_migrator` account, which defaults to `access_rw` but can elevate to `access_rw_pii` when it needs to work with objects in the `pii` schema. This is a much easier way of managing PII access than we've had in the past! A migration example:

```
-- Altering public schema, works the same as it always did
create table my_new_data (id int primary key);

-- Trying to alter data in the pii schema, will fail
create table pii.my_sensitive_data (id int primary key);

-- Elevating privileges and then altering pii schema, will succeed
set role to access_rw_pii;
create table pii.my_sensitive_data (id int primary key);

-- Can set role back to access_rw if there's more work to do in the public schema,
-- but no need to reset role if there's nothing left to migrate
```

Backend services will use the `gallery_backend` account, which defaults to `access_rw_pii` and will read/write data as usual.

The `gallery_team_*` accounts are for different levels of manual access by Gallery team members, and they mostly do what their names imply, with one caveat: `gallery_team_ro_pii` and `gallery_team_rw_pii` log in _without_ access to the `pii` schema, but can `set role to access_ro_pii` or `set role to access_rw_pii` if required. Access to sensitive data should always be intentional and not a default behavior.

## Other Stuff
- Tests have been updated to migrate up to migration 000054 using the `postgres` user, and then the `gallery_migrator` user from that point on. `54` will probably need to be changed to 55 or 56 by the time I actually merge this!

- I got rid of the initial DB schema used in our postgres Dockerfile, since we weren't keeping it up to date and a new local Postgres container should just run local migrations.

- The `public` schema is the default schema, and tables in that schema can be referenced the same as always (e.g. `table_name`, not `public.table_name`). Referencing tables in the `pii` schema requires specifying the schema name, as in: `pii.table_name`. 

- sqlc is aware that `public` is the default schema, so generated types in the public schema will have the same names as before. Since `pii` is _not_ the default schema, its generated types will be prefixed with its schema name, as in `PiiTableName`. As such, I renamed the PII tables when moving them out of the public schema:
  - `pii_for_users` -> `pii.for_users` # Generated type is `PiiForUsers`
  - `users_with_pii` -> `pii.user_view` # Generated type is `PiiUserView`


cmd/allowlist/main.go
```diff
@@ -106,7 +106,7 @@ func dedupeAddresses(addresses []persist.Address) []persist.Address {
 func setDefaults() {
 	viper.SetDefault("POSTGRES_HOST", "0.0.0.0")
 	viper.SetDefault("POSTGRES_PORT", 5432)
-	viper.SetDefault("POSTGRES_USER", "postgres")
+	viper.SetDefault("POSTGRES_USER", "gallery_backend")
```
radazen, line 109: Not sure about this one, or the ones in `merch_prepop`, `notifications_prepop`, and `update`. They all seem like command-line utilities that get run locally, at which point `gallery_team_rw` is probably the correct role to connect with.

db/migrations/core/000054_set_schema_permissions.up.sql
```diff
@@ -0,0 +1,242 @@
...
+grant access_rw to gallery_migrator;
+grant access_rw_pii to gallery_migrator;
+alter role gallery_migrator set role to access_rw;
+
+------------------------------------------------------------------------------------
+------------------------------------------------------------------------------------
+------------------------------------------------------------------------------------
+-- Anything above this section is generic and applicable to any gallery database.
+-- The remainder of the migration is specific to the existing tables in our main
+-- database.
```
radazen, line 186: I didn't apply any of this to the indexer database, and it's probably not necessary since it just has publicly available data in it. Still, if we ever do want to set up different access roles there, everything up to this point in the migration file is applicable.

service/persist/postgres/postgres.go
```diff
@@ -62,11 +73,59 @@ func getSqlConnectionString() string {
...
+
+func WithPort(port int) ConnectionOption {
+	return func(params *connectionParams) {
+		params.port = port
+	}
+}
+
 // NewClient creates a new postgres client
-func NewClient() *sql.DB {
+func NewClient(opts ...ConnectionOption) *sql.DB {
```
radazen, line 119: Needed to override _just_ the username in order to run migrations during test setup, so I implemented the options pattern here.

sqlc.yaml
```diff
@@ -128,44 +128,45 @@ sql:
             go_type: "github.com/mikeydub/go-gallery/service/persist.TokenID"
 
           # Wildcards
-          # Note: to override one of these wildcard entries, add a more specific entry (like some_table.id) above
-          - column: "*.id"
+          # Note: to override one of these wildcard entries, add a more specific entry (like some_table.id) above.
+          # Format is schema.table.column; where *.*.<column> applies to all schemas and tables.
```
radazen, line 132: Had to add an extra `*.` wildcard here to account for multiple schemas

Makefile
```diff
@@ -387,21 +387,22 @@ migrate-local-indexerdb:
...
+	migrate -path ./db/migrations/core -database "postgresql://postgres@localhost:5432/postgres?sslmode=disable" goto 54
+	migrate -path ./db/migrations/core -database "postgresql://gallery_migrator@localhost:5432/postgres?sslmode=disable" up
 
 confirm-dev-migrate:
 	@prompt=$(shell bash -c 'read -p "Are you sure you want to apply migrations to the dev DB? Type \"development\" to confirm: " prompt; echo $$prompt'); \
 	if [ "$$prompt" != "development" ]; then exit 1; fi
 
 migrate-dev-coredb: start-dev-sql-proxy confirm-dev-migrate
-	migrate -path ./db/migrations/core -database "postgresql://$(POSTGRES_USER):$(POSTGRES_PASSWORD)@localhost:6643/$(POSTGRES_DB)?sslmode=disable" up
+	@migrate -path ./db/migrations/core -database "postgresql://$(POSTGRES_MIGRATION_USER):$(POSTGRES_MIGRATION_PASSWORD)z@localhost:6643/$(POSTGRES_DB)?sslmode=disable" up
```
jarrel-b, line 398: Extra `z`?

db/migrations/core/000054_set_schema_permissions.up.sql
```diff
@@ -0,0 +1,242 @@
...
+drop role if exists gallery_team_rw;
+create user gallery_team_rw noinherit login;
+grant access_rw to gallery_team_rw;
+alter role gallery_team_rw set role to access_rw;
+
+drop role if exists gallery_team_rw_pii;
+create user gallery_team_rw_pii noinherit login;
+grant access_rw to gallery_team_rw_pii;
+grant access_rw_pii to gallery_team_rw_pii;
+-- pii roles log in without pii access, but can use "set role" to gain access when needed
```
jarrel-b, line 159: Love this pattern!

db/migrations/core/000054_set_schema_permissions.up.sql
```diff
@@ -0,0 +1,242 @@
...
+grant all on all functions in schema public to access_rw_pii;
+
+-- Grant read/write privileges on all current objects in the pii schema
+grant all on all tables in schema pii to access_rw_pii;
+grant all on all sequences in schema pii to access_rw_pii;
+grant all on all functions in schema pii to access_rw_pii;
+
+------------------------------------------------------------------------------------
+-- Create standard team login roles (note: before these can be used, a password
+-- must be set in the GCP management console)
```
jarrel-b, line 136: Will these creds be given out as needed, or stored somewhere like 1Password?

jarrel-b (approved):
looks awesome Ezra!

Makefile
radazen: Oof, good eye!

db/migrations/core/000054_set_schema_permissions.up.sql
radazen: I was figuring something like 1Password, potentially split across a couple of shared vaults to limit access.

PR #789, opened 2023-02-01
diffing improvement

benny-conn:
Changes:

- **Changed** update gallery mutation to diff better on string values


indexer/indexer.go
```diff
@@ -255,7 +256,15 @@ func newIndexer(ethClient *ethclient.Client, ipfsClient *shell.Shell, arweaveCli
 			panic(err)
 		}
 		i.lastSyncedChunk = recentDBBlock.Uint64()
-		i.lastSyncedChunk -= (i.lastSyncedChunk % blocksPerLogsCall) + (blocksPerLogsCall * defaultWorkerPoolSize)
+
```
benny-conn, line 259: This was causing an error running locally the indexer server without any seeded database. Since there was nothing in the database, the last synced chunk was `0` and when you subtract something from an unsigned integer that will lead it into the land of negative numbers, it looped back up to the max unsigned int which was greater than the current ETH block, causing an error.

publicapi/gallery.go
```diff
@@ -210,6 +223,9 @@ func (api GalleryAPI) UpdateGallery(ctx context.Context, update model.UpdateGall
 		return db.Gallery{}, err
 	}
 
+	if update.Caption != nil && *update.Caption == "" {
```
benny-conn, line 226: Superstitious code

server/server.go
```diff
@@ -183,7 +183,7 @@ func SetDefaults() {
 	viper.SetDefault("SENTRY_DSN", "")
 	viper.SetDefault("GCLOUD_FEED_QUEUE", "projects/gallery-local/locations/here/queues/feed-event")
 	viper.SetDefault("GCLOUD_WALLET_VALIDATE_QUEUE", "projects/gallery-dev-322005/locations/us-west2/queues/wallet-validate")
-	viper.SetDefault("GCLOUD_FEED_BUFFER_SECS", 5)
+	viper.SetDefault("GCLOUD_FEED_BUFFER_SECS", 20)
```
benny-conn, line 186: increased this so that I had time to make multiple updates when testing locally

db/queries/core/query.sql
```diff
@@ -283,10 +283,10 @@ ORDER BY tokens.created_at DESC, tokens.name DESC, tokens.id DESC;
 INSERT INTO events (id, actor_id, action, resource_type_id, user_id, subject_id, data, group_id, caption) VALUES ($1, $2, $3, $4, $5, $5, $6, $7, $8) RETURNING *;
 
 -- name: CreateTokenEvent :one
-INSERT INTO events (id, actor_id, action, resource_type_id, token_id, subject_id, data, group_id, caption) VALUES ($1, $2, $3, $4, $5, $5, $6, $7, $8) RETURNING *;
+INSERT INTO events (id, actor_id, action, resource_type_id, token_id, subject_id, data, group_id, caption, gallery_id, collection_id) VALUES ($1, $2, $3, $4, $5, $5, $6, $7, $8, $9, $10) RETURNING *;
 
 -- name: CreateCollectionEvent :one
-INSERT INTO events (id, actor_id, action, resource_type_id, collection_id, subject_id, data, caption, group_id) VALUES ($1, $2, $3, $4, $5, $5, $6, $7, $8) RETURNING *;
+INSERT INTO events (id, actor_id, action, resource_type_id, collection_id, subject_id, data, caption, group_id, gallery_id) VALUES ($1, $2, $3, $4, $5, $5, $6, $7, $8, $9) RETURNING *;
```
benny-conn, line 289: This was the cause of a lot of events having null galleries.

publicapi/gallery.go
```diff
@@ -266,8 +282,13 @@ func updateCollectionsInfoAndTokens(ctx context.Context, q *db.Queries, actor, g
 	}
 
 	for _, collection := range update {
+		curCol, err := q.GetCollectionById(ctx, collection.Dbid)
+		if err != nil {
```
jarrel-b, line 286: Does `update` contain both new and updated collections, or only updated collections? Just wondering if we would exit here on a new collection that hasn't been written to the db yet and skip updating the other collections

publicapi/publicapi.go
```diff
@@ -136,18 +136,26 @@ func dispatchEvent(ctx context.Context, evt db.Event, v *validator.Validate, cap
 }
 
 func dispatchEvents(ctx context.Context, evts []db.Event, v *validator.Validate, caption *string) (*db.FeedEvent, error) {
+
+	if len(evts) == 0 {
+		return nil, nil
```
jarrel-b, line 141: Not sure if its better to return a pointer to an instantiated `db.FeedEvent` vs `nil` if the caller is expecting a non-nil value if no error is returned, but think this probably ok!

jarrel-b (approved):

publicapi/gallery.go
benny-conn: It only contains updated collections, based on the way the frontend sends everything, there will never be a time that updated collections have a collection that is also created (because the ID won't be available yet). **But** I also don't see a problem with moving this step after the create collection step just in case!

PR #795, opened 2023-02-06
Benny/feed edit session

benny-conn:
Changes:

- **Added** a new `groupHandler` interface for event groups to be processed
- **Updated** the `updateGallery` input to take in an `editID` 
- **Added** new dispatchers for groups that will sidestep the feed service, grab all events with group ID (the `editID` will be the groupID for galleries) and process them immediately
- **Added** a `publishGallery` event that will shoot out the group dispatcher as well well as add an optional caption to the events
- **Removed** the wait segments for the actions involved with updating a gallery so we don't wait for any window for gallery update events (this shouldn't affect anything given that these types of events should not even reach the feed service and the function that waits for the window is not involved in the grouped handler, but updated this for code clarity)

Considerations:

- At first I thought we would stop using the feed service, but there are still events that are grouped with a delay, actually only one event type. The user followed user events. We can leave the feed service mostly as is then and it will remain extensible just in case there are more events in the future that we want to update with a delay.

TODO:

- [x] Tests for group IDs and publishing


graphql/schema/schema.graphql
```diff
@@ -702,7 +702,7 @@ type CollectorsNoteAddedToTokenFeedEventData implements FeedEventData {
   eventTime: Time
   owner: GalleryUser @goField(forceResolver: true)
   action: Action
-  token: CollectionToken @goField(forceResolver: true)
+  token: Token @goField(forceResolver: true)
```
jakzaizzat, line 705: @benny-conn Since we changed from `CollectionToken` to `Token`, what is the best way to retrieve the collection detail (name,id,etc) for feed

graphql/schema/schema.graphql
```diff
@@ -1628,12 +1629,30 @@ input UpdateGalleryInput {
   createdCollections: [CreateCollectionInGalleryInput]
 
   order: [DBID!]
+
+  # generated on the frontend
+  # will be String! in the future, not required for temp backwards compatibility
+  editID: String
```
jarrel-b, line 1635: Naming nitpick: maybe `editId` to stick with the `Id` convention?

graphql/graphql_test.go
```diff
@@ -236,7 +236,7 @@ func testCreateCollection(t *testing.T) {
 	assert.Len(t, payload.Collection.Tokens, len(userF.tokenIDs))
 }
 
-func testUpdateGalleryWithCaption(t *testing.T) {
+func testUpdateGalleryWithPublish(t *testing.T) {
```
jarrel-b, line 239: Can't comment on the line this pertains to, but there are helper functions I added `defaultLayout` and `defaultTokenSettings` that you can use to remove some of the boilerplate in building the `CreateCollectionInput` to make the tests a bit more readable

graphql/resolver/schema.resolvers.go
```diff
@@ -971,6 +975,22 @@ func (r *mutationResolver) UpdateGallery(ctx context.Context, input model.Update
 	return output, nil
 }
 
+func (r *mutationResolver) PublishGallery(ctx context.Context, input model.PublishGalleryInput) (model.PublishGalleryPayloadOrError, error) {
+	err := publicapi.For(ctx).Gallery.PublishGallery(ctx, input)
```
jarrel-b, line 979: Would it be worthwhile to check if a gallery has already been published? Would the outcome be it just generates a new feed event, and maybe updates the captions?

jarrel-b (approved):
Looks great Benny! Love how this makes the feed and captioning so much easier to reason about + tests look great too

event/event.go
```diff
@@ -157,15 +200,22 @@ func (e *eventSender) addImmediateHandler(dispatcher *eventDispatcher, action pe
...
 }
 
+func (e *eventSender) addGroupHandler(dispatcher *eventDispatcher, action persist.Action, handler groupHandler) {
+	dispatcher.addGroup(action, handler)
+	e.registry[groupKey][action] = struct{}{}
+}
+
 type eventDispatcher struct {
 	delayedHandlers   map[persist.Action]delayedHandler
 	immediateHandlers map[persist.Action]immediateHandler
```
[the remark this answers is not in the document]
benny-conn: Great idea! Def a fix it week thing.

event/event.go
```diff
@@ -237,20 +302,46 @@ type feedHandler struct {
 
 func newFeedHandler(queries *db.Queries, taskClient *cloudtasks.Client) feedHandler {
 	return feedHandler{
-		eventBuilder: feed.NewEventBuilder(queries, true),
+		eventBuilder: feed.NewEventBuilder(queries),
 		tc:           taskClient,
 	}
 }
 
+var actionsToBeHandledByFeedService = map[persist.Action]bool{
```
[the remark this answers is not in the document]
benny-conn: The purpose of this was so that we still have the delayed handlers for the events themselves. That way they all get to the database as events, this step just is responsible for sending "some" events to the feed service, now of course just one action being the follow action. In a sense, the semantics of the actions not included in this map still having "delayed" handlers makes sense. They are still being handled in a delayed manor. They just don't need to be received by the feed service.

feed/merge.go
```diff
@@ -135,14 +135,21 @@ func (c *combinedGalleryEvent) merge(eventsAsc []db.Event) *combinedGalleryEvent
 			c.caption = &event.Caption.String
 		}
 		if event.Action == persist.ActionGalleryInfoUpdated {
-			c.galleryName = event.Data.GalleryName
-			c.galleryDescription = event.Data.GalleryDescription
+			if event.Data.GalleryName != "" {
```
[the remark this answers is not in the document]
benny-conn: Great point, might need to make this a pointer or store some state of non-existence because the frontend will not include the name or desc over and over again in each save if it changes.

graphql/graphql_test.go
benny-conn: Wonderful! Did not know about these, thank you!

graphql/resolver/schema.resolvers.go
benny-conn: True. Maybe we can also store the edit_id on the feed event so we can check if a feed event with the given edit ID already exists, and in that case just update the caption.

PR #797, opened 2023-02-06
Add time factor to trending feed

jarrel-b:
This PR adds an [exponential decay](https://en.wikipedia.org/wiki/Exponential_decay) term to the trending feed calculation to penalize events that are older in age. This decay rate can be tuned using the `lambda` parameter which controls how quickly values decrease over time.

### DIFF
--- a/db/queries/core/query.sql
+++ b/db/queries/core/query.sql
@@ -787,8 +787,10 @@ update users set user_experiences = user_experiences || @experience where id = @
 select users.* from users join unnest(@user_ids::varchar[]) with ordinality t(id, pos) using (id) where deleted = false order by t.pos asc;
 
 -- name: GetTrendingFeedEventIDs :many
-select feed_event_id from events where action in ('CommentedOnFeedEvent', 'AdmiredFeedEvent') and created_at >= @window_end and feed_event_id is not null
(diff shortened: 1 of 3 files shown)


db/queries/core/query.sql
```diff
@@ -787,8 +787,10 @@ update users set user_experiences = user_experiences || @experience where id = @
 select users.* from users join unnest(@user_ids::varchar[]) with ordinality t(id, pos) using (id) where deleted = false order by t.pos asc;
 
 -- name: GetTrendingFeedEventIDs :many
-select feed_event_id from events where action in ('CommentedOnFeedEvent', 'AdmiredFeedEvent') and created_at >= @window_end and feed_event_id is not null
-group by feed_event_id order by count(*) desc, max(created_at) desc limit sqlc.arg('limit');
+select feed_events.id, date_trunc('hour', feed_events.created_at)::timestamp created_at, count(*)
```
jarrel-b, line 790: Timestamps are floored to the nearest hour since I don't think the minutes/seconds are very significant

Robinnnnn (approved):

kaitoo1 (approved):
what an impactful ~40 lines of code 🚀🚀

PR #801, opened 2023-02-08
Fix GCP storage client error

jarrel-b:
This PR fixes the unauthenticated error which was preventing media for tokens to be cached. This PR also adds additional logger fields to help make tracing individual tokens in Cloud Logs easier.

### DIFF
--- a/service/media/media.go
+++ b/service/media/media.go
@@ -20,8 +20,8 @@ import (
 
 	"github.com/mikeydub/go-gallery/service/logger"
 	"github.com/mikeydub/go-gallery/service/mediamapper"
-	"github.com/mikeydub/go-gallery/service/tracing"
 	"github.com/sirupsen/logrus"
+	"google.golang.org/api/googleapi"
 	"google.golang.org/api/option"
 
 	"cloud.google.com/go/storage"
@@ -32,7 +32,6 @@ import (
 	"github.com/mikeydub/go-gallery/util"
 	"github.com/qmuntal/gltf"
 	"github.com/spf13/viper"
-	htransport "google.golang.org/api/transport/http"
 )
 
 var errAlreadyHasMedia = errors.New("token already has preview and thumbnail URLs")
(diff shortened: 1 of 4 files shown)


service/media/media.go
```diff
@@ -78,25 +77,17 @@ var postfixesToMediaTypes = map[string]mediaWithContentType{
...
-
-	httpClient, _, err := htransport.NewClient(ctx, options...)
-	if err != nil {
-		panic(err)
+		opts = append(opts, option.WithCredentialsJSON(fi))
 	}
 
-	httpClient.Transport = tracing.NewTracingTransport(http.DefaultTransport, false)
-
-	storageClient, err := storage.NewClient(ctx, option.WithHTTPClient(httpClient), option.WithScopes(storage.ScopeFullControl))
```
jarrel-b, line 99: Could not for the life of me figure out how to successfully create the correct client via providing options to an underlying HTTP client, even if the options passed to it are the same options that `NewClient` would've passed to it anyways. Passing the same options inline to `NewClient` didn't seem to work either. This means we won't get sentry traces for now, but may take a crack again with fresh eyes. Note that providing scopes aren't needed because `NewClient` adds the proper scope already.

service/media/media.go
```diff
@@ -653,33 +652,34 @@ outer:
...
-				err = cacheRawAnimationMedia(pCtx, reader, bucket, fmt.Sprintf("%s-%s", ipfsPrefix, name), storageClient)
-				if err != nil {
-					return mediaType, false, err
-				}
-				logger.For(pCtx).Infof("cached animation for %s in %s", name, time.Since(timeBeforeCache))
-				return mediaType, true, nil
-			}
+	}
+
+	switch asURI.Type() {
```
jarrel-b, line 657: This code was pulled out of the above `switch` statement to remove some nesting

Robinnnnn (approved):

PR #809, opened 2023-02-14
Benny/twitter connections

benny-conn:
Changes:

- **Added** query for getting social connections for a given social provider (just twitter for the moment)
- **Added** twitter API functions necessary for getting following

Considerations:

- Caching twitter following for 1 week because of the rate limits and because we do not need a user's folllowing data to be up to date all the time, but this can be tweaked

TODO:

- [ ] Testing


db/queries/core/query.sql
```diff
@@ -833,3 +833,27 @@ update pii.for_users set pii_socials = @socials where user_id = @user_id;
...
 -- name: UpdateEventCaptionByGroup :exec
 update events set caption = @caption where group_id = @group_id and deleted = false;
+
+-- this query will take in enoug info to create a sort of fake table of social accounts matching them up to users in gallery with twitter connected.
+-- it will also go and search for whether the specified user follows any of the users returned
+-- name: GetSocialConnections :many
+select s.*, user_view.id as user_id, user_view.created_at as user_created_at, (f.id is not null)::bool as already_following
+from (select unnest(@social_ids::varchar[]) as social_id, unnest(@social_usernames::varchar[]) as social_username, unnest(@social_displaynames::varchar[]) as social_displayname, unnest(@social_profile_images::varchar[]) as social_profile_image) as s
+    inner join pii.user_view on user_view.pii_socials->$1::text->>'id'::varchar = s.social_id and user_view.deleted = false
+    left outer join follows f on f.follower = @user_id and f.followee = user_view.id and f.deleted = false
```
jarrel-b, line 843: I think you could omit `outer` since `left outer join` is the same as `left join`. `outer` is used so that Postgres matches the SQL standard that other databases also adhere to. Same for `inner join`and `join`. Unless they're there to be explicit then cool with me!

service/twitter/twitter.go
```diff
@@ -142,3 +206,106 @@ func (a *API) getAuthedUser(ctx context.Context, accessToken string) (TwitterIde
...
+
+	return a, nil, nil
+}
+
+func (a *API) GetFollowing(ctx context.Context) ([]TwitterIdentifiers, error) {
+	if !a.isAuthed {
+		return nil, errAPINotAuthed
+	}
+
+	redisPath := fmt.Sprintf("twitter-%s-following", a.TIDs.ID)
```
jarrel-b, line 248: A common redis convention is to delimit by `.`, you think we can use that instead of `-`? It's neat because it namespaces keys, for example if we add followers:
```
twitter.<id>.followers
```

And if we want all twitter info for a specific user we can do:
```
GET twitter.<id>.*
```

Now that I think about it, I wonder if it makes sense to do something like:
```
<id>.twitter.following
<id>.twitter.followers
<id>.farcaster.following
<id>.farcaster.followers
```
So if we want to get all social info on redis for a specific user we can just do:
```
GET <id>.*
```

jarrel-b, line 248: Is `a.TIDs.ID` a twitter ID? Wondering if we should key on the gallery userID instead to make lookup easier in the future.

graphql/resolver/schema.resolvers.helpers.go
```diff
@@ -1043,6 +1044,13 @@ func resolveDeletedNodeByID(ctx context.Context, id persist.DBID) (*model.Delete
 	}, nil
 }
 
+func resolveSocialConnectionByIdentifiers(ctx context.Context, socialId string, socialType persist.SocialProvider) (*model.SocialConnection, error) {
+	return &model.SocialConnection{
+		SocialID:   socialId,
+		SocialType: socialType,
```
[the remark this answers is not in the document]
benny-conn: There is no reliable way that I could find to get those pieces of information from just the `socialId` and `socialType` (given that this info is found on the fly from the twitter API and will be different for each social type often requiring authentication to access). I could include those fields as a part of the node ID potentially. Will chat with terence about how necessary this would be.

graphql/schema/schema.graphql
```diff
@@ -904,6 +929,15 @@ type Query {
   # Retool Specific
   usersByRole(role: Role!, before: String, after: String, first: Int, last: Int): UsersConnection
     @retoolAuth
+
+  socialConnections(
```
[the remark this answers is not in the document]
benny-conn: Left this out because I assumed we would want to be querying for possibly more than one social account type at a time and I believe that if this is in the viewer type that each new social type query would replace the data from the last one. Leaving it as a separate query allows the frontend to make as many queries as it wants and conglomerate that data somewhere else. 

It would make sense to be in the viewer type maybe if it it took in an array of social account types, that way we can guarantee that this one query will return all of the data the frontend wants all at once. I can chat with @Robinnnnn about it, but it seems we won't want to include data from multiple social providers at once and this also would really complicate pagination (do we paginate through each social account type together, do we include them side to side so when you finish with one we move into another, etc.)

publicapi/socials.go
```diff
@@ -1,25 +1,200 @@
...
+		profileImages, _ := util.Map(initialConnections, func(m model.SocialConnection) (string, error) {
+			return m.ProfileImage, nil
+		})
+		results, err := api.queries.GetSocialConnections(ctx, db.GetSocialConnectionsParams{
+			Limit:               params.Limit,
+			UserID:              userID,
+			SocialIds:           socialIDs,
+			SocialUsernames:     usernames,
+			SocialDisplaynames:  displaynames,
+			Column1:             socialProvider.String(),
```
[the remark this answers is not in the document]
benny-conn: Found a workaround to sqlc not liking named params as jsonb query inputs. Originally used `@named_param` but that did not work so I used `$1` which led to this weird name. Using `sqlc.arg('named_param')` worked though!

db/queries/core/query.sql
benny-conn: Good to know! If it doesn't affect anything, is it alright if we leave it? After learning more about what they all do, the explicitness helps my brain comprehend where joins are happening haha

service/twitter/twitter.go
benny-conn: Yes indeed it is. This twitter package is supposed to have no context on the user, however, so all we have access to is twitter identifiers like social ID and username. Everything related to the gallery user who owns the twitter account happens outside of this package.

jarrel-b: Makes sense!

PR #811, opened 2023-02-15
User suggestions based on follows

jarrel-b:
The algorithm is based on Pinterest's Pixie algorithm: https://medium.com/pinterest-engineering/an-update-on-pixie-pinterests-recommendation-system-6f273f737e1b. At a high level, the algorithm begins by selecting a set of "query nodes" from a user's following list and then runs a random walk per node, keeping track of the nodes visited during the walk. The most frequently visited nodes across all walks are the most "relevant" to the user and get returned as final suggestions. The suggestions are also shuffled so that is not a static list returned each time.

 The suggestions are available on the `Viewer` type as a new field named `suggestedUsers`, which returns a `UserConnection:
```
  suggestedUsers(before: String, after: String, first: Int, last: Int): UsersConnection
    @goField(forceResolver: true)
```

It can be paginated like a typical cursor, but if no cursor is provided, it will generate a new set of recommendations.

### Follow graph
The graph is represented as an adjacency list where the key is a user and the value is a list of users that the user follows. It is instantiated as a singleton using the `sync.Map` type to manage concurrent access and is refreshed periodically in the background every hour. I decided to keep the service in the backend since the graph itself is less than 300kb and could be moved to a separate service when we want to scale the service independently of the backend.

### Selecting query nodes
Rather than selecting all follows, the algorithm selects 30 follows at random weighted by how recent the follow is. This keeps the execution time predictable and allows it to complete in real-time, but we can tune this to be any number.

### Handling users with no followers
Each recommendation is saved to a table named `recommendation_results` which keeps track of how frequently a user gets recommended. This table is rolled up in a view that keeps track of the top 100 most frequently recommended users. We use this view for new users who have no followers, the idea being that this view is the "average" recommendation across all users.

### Post-merge TODOs
* [x] Add tests
* [ ] Run migration to add the `recommended_history` table
* [ ] Add refresh job to pg_cron to refresh `top_recommended_users` view


db/migrations/core/000061_add_recommendation_results_table.up.sql
```diff
@@ -0,0 +1,25 @@
+set role to access_rw;
```
radazen, line 1: Hmm, this is an interesting one. I was going to say that `set role to access_rw;` shouldn't be necessary, since the `gallery_migrator` role always starts out with `access_rw`. When working on this stuff, I tested that assumption by trying to apply two migrations in sequence: one that `set role to access_rw_pii` and _didn't set it back to access_rw_, and one that attempted to access the `pii` schema (which failed as expected). It seemed to me that each migration was applied by a new "login" of the migrator role.

...but I have to imagine that's not the case when we run migrations in our test suite, using the golang-migrate API directly. Hmm. So maybe it is a best practice to set the access level explicitly at the beginning of every migration, huh?

db/migrations/core/000061_add_recommendation_results_table.up.sql
```diff
@@ -0,0 +1,25 @@
...
+  id character varying(255) primary key,
+  version int default 0,
+  user_id character varying(255) references users (id),
+  recommended_user_id character varying(255) references users (id),
+  recommended_count integer,
+  created_at timestamp with time zone default current_timestamp not null,
+  last_updated timestamp with time zone default current_timestamp not null,
+  deleted boolean default false not null,
+  unique (user_id, recommended_user_id, version),
+  constraint user_id_not_equal_recommend_id_constraint check (user_id != recommended_user_id)
```
radazen, line 13: 😍

db/migrations/core/000061_add_recommendation_results_table.up.sql
```diff
@@ -0,0 +1,25 @@
...
+);
+
+drop materialized view if exists top_recommended_users;
+
+create materialized view top_recommended_users as (
```
radazen, line 18: Assuming we have a unique key here (which seems to be the case -- `recommended_user_id`), you probably want to create a unique index on the materialized view based on that column. For whatever reason, a unique index is a requirement for refreshing a materialized view with the `concurrently` keyword, which lets you keep accessing the view even while it's being refreshed.

publicapi/user.go
```diff
@@ -866,3 +867,86 @@ func (api UserAPI) UpdateUserExperience(ctx context.Context, experienceType mode
...
+		if _, userIDs, err = paginator.decodeCursor(*before); err != nil {
+			return nil, PageInfo{}, err
+		}
+	} else if after != nil {
+		if _, userIDs, err = paginator.decodeCursor(*after); err != nil {
+			return nil, PageInfo{}, err
+		}
+	} else {
+		// Otherwise make a new recommendation
+		curUserID, err := getAuthenticatedUserID(ctx)
```
radazen, line 892: Should we move the `getAuthenticatedUserID` check higher? Like, if the user got logged out for some reason and kept paginating the list, would we want to reject them or keep paging?

publicapi/user.go
```diff
@@ -866,3 +867,86 @@ func (api UserAPI) UpdateUserExperience(ctx context.Context, experienceType mode
...
+
+		results := make([]any, len(keys))
+		for i, key := range keys {
+			results[i] = key
+		}
+
+		return results, nil
+	}
+
+	paginator.CursorFunc = func(node any) (int, []persist.DBID, error) {
```
radazen, line 937: Unrelated to your PR, but I think/hope we may be able to simplify some of this pagination type conversion stuff now that we have generics available.

service/recommend/graph.go
```diff
@@ -0,0 +1,44 @@
...
+
+type graphMetadata struct {
+	Indegrees map[persist.DBID]int `json:"indegrees"`
+}
+
+func generateGraph(ctx context.Context, queries *db.Queries) (*graph, error) {
+	span, ctx := tracing.StartSpan(ctx, "recommend", "generateGraph")
+	defer tracing.FinishSpan(span)
+
+	follows, err := queries.GetFollowGraphSource(ctx)
```
radazen, line 24: This is potentially a very long list, huh? I wonder if there's a size at which this need to be paginated or something, or Postgres will happily stream as many rows as the client can handle. Probably the latter?

service/recommend/recommend.go
```diff
@@ -0,0 +1,201 @@
...
+
+	"github.com/gin-gonic/gin"
+	db "github.com/mikeydub/go-gallery/db/gen/coredb"
+	"github.com/mikeydub/go-gallery/service/logger"
+	"github.com/mikeydub/go-gallery/service/persist"
+	"github.com/mikeydub/go-gallery/util"
+)
+
+const contextKey = "recommend.instance"
+const metadataKey = "____metadata____"
```
radazen, line 17: If you're trying to ensure uniqueness of the key, this is a good approach: https://gist.github.com/ww9/4ad7b2ddfb94816a30dfdf2218e02f48

I've been trying to use it with new context keys that I create, though most of my old ones are just strings.

service/recommend/recommend.go
```diff
@@ -0,0 +1,201 @@
...
+
+func (r *Recommender) readMetadata(ctx context.Context) graphMetadata {
+	val, ok := currentGraph.Load(metadataKey)
+	if !ok {
+		panic("no metadata for graph!")
+	}
+	return val.(graphMetadata)
+}
+
+// shuffle shuffles IDs within partitions so that results are in a similar order.
```
radazen, line 179: This is very cool

radazen (approved):
This is so good. Cannot wait to see it in action. 👍 👍 👍

publicapi/user.go
jarrel-b: Good call!

service/recommend/recommend.go
jarrel-b: Nice trick! `gin` requires this to be a string, but I'll use this for maps.

db/migrations/core/000061_add_recommendation_results_table.up.sql
jarrel-b: Huh this actually worked without setting the role, I was getting a permissions error on Github the last time I ran this. Works for me!

service/recommend/graph.go
jarrel-b: Yea I'm thinking the latter. It looks like sqlc just returns the entire result set, though I'm not sure if calling `Rows.Next()` is a generator that fetches the next from the DB each time it gets called or not. Another idea was to load the graph once on startup, and only apply the diffs each time a user follows / unfollows someone. But probably won't need to do something like that for a while.

kaitoo1:
absolutely incredible stuff - worked like a charm with the frontend here: https://github.com/gallery-so/gallery/pull/1352

PR #812, opened 2023-02-15
Upgrade to latest GQLGen for Federation 2 support

tbezman:
Seems like we were falling behind a bit.

The notable changes here are
- Upgrade gqlgen to 0.17.24
- Pass in the template string directly using go embed
- Bring in the latest `modelgen` from gqlgen and apply our changes back on top


radazen (approved):
100% awesome. Thanks for doing this! Bonus points for figuring out how to update the modelgen plugin, cause I don't think I documented that particularly well. 🙈

PR #813, opened 2023-02-15
Apollo Router + Federation Support

tbezman:
The goal of this PR is to enable [Federation 2](https://www.apollographql.com/docs/federation) support in GQLGen w/ the `FeedEvent` type as a guinea pig for the [Entity Spec](https://www.apollographql.com/docs/federation/entities/)

### Callouts
- I worked w/ @jarrel-b this morning to get this thing deployed and all is good but...
  - We still don't have a nice DNS record setup yet
  - We have an empty secrets yaml file since we don't have any environment variables CC @radazen for help here.

### Entities
The entity system in Apollo land is analogous to the Node system we already have for Relay. It lets Apollo look up an arbitrary type by it's `key` field. This is particularly useful for `@defer` support.

Consider the following query
image

You can imagine, if we want to defer the `SubEventFragment`, the Apollo Router can break the query into two.
1. Get all the ids of the items in the paginated list
2. Fetch the deferred data on each of those items

Meanwhile, all of this data is getting streamed to the client as the data resolves.

### Adoption Plan
My plan is to adopt this as we're developing the mobile app to battle test it before we roll it out to web.

Here's a list of things we still need to verify
- Is deferred resolution of entities getting batched properly?
- How does all of this work with APQ?
  - It seems like Apollo Router does some handling of APQ so we might be able to just drop it from GQLGen
- CORS support on GQLGen long term will be able to disappear. Our internal graph does not have to be publicly accessible
  - Retool might be the exception to this but we'll see if that's overcome by some other event.


docker-compose.yml
```diff
@@ -64,6 +64,7 @@ services:
       "-c",
       "gcloud beta emulators pubsub start --host-port=0.0.0.0:8085 --project=gallery-local"
     ]
+
```
tbezman, line 67: don't @ me

docker/graphql-gateway/dev/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+WORKDIR /workspace
+
+RUN curl -sSL https://rover.apollo.dev/nix/v0.11.1 | sh
+RUN curl -sSL https://router.apollo.dev/download/nix/latest | sh
+
+ADD graphql/schema/schema.graphql .
+ADD docker/graphql-gateway/dev/supergraph.yaml .
+ADD docker/graphql-gateway/dev/router.yaml .
+
+ENV APOLLO_ELV2_LICENSE=accept
```
tbezman, line 12: This license just says we can't compete with Apollo basically so we're chill here

docker/graphql-gateway/dev/router.yaml
```diff
@@ -0,0 +1,10 @@
+cors:
+  allow_credentials: true
+  match_origins:
+    - "^https://gallery.*.vercel.app$"
+    - "http://localhost:3000"
```
tbezman, line 5: on dev we want to match all vercel links that start with `gallery`

docker/graphql-gateway/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+version: "3.9"
+
+services:
+  graphql-gateway-local:
+    build:
+      context: ../../
```
tbezman, line 6: we need our context to be the root of the repo so we can access the graphql schema from the server

docker/graphql-gateway/prod/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN curl -sSL https://router.apollo.dev/download/nix/latest | sh
+
+ADD graphql/schema/schema.graphql .
+ADD docker/graphql-gateway/prod/supergraph.yaml .
+ADD docker/graphql-gateway/prod/router.yaml .
+
+ENV APOLLO_ELV2_LICENSE=accept
+RUN ~/.rover/bin/rover supergraph compose --config ./supergraph.yaml > ./supergraph-schema.graphql
+
+CMD ./router --dev --config ./router.yaml --supergraph ./supergraph-schema.graphql
```
tbezman, line 15: this `--dev` flag gives us access to the apollo explorer (like graphiql)

go.mod
```diff
@@ -76,7 +76,7 @@ require (
 	github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect
 	github.com/Azure/go-autorest v14.2.0+incompatible // indirect
 	github.com/Azure/go-autorest/autorest v0.11.26 // indirect
-	github.com/Azure/go-autorest/autorest/adal v0.9.18 // indirect
+	github.com/Azure/go-autorest/autorest/adal v0.9.22 // indirect
```
tbezman, line 79: apparently the `0.9.18` version was deleted, got some weird error message.

gqlgen.yml
```diff
@@ -8,9 +8,10 @@ exec:
   package: generated
 
 # Uncomment to enable federation
-# federation:
-#   filename: graph/generated/federation.go
-#   package: generated
+federation:
+ filename: graphql/generated/federation.go
+ package: generated
+ version: 2
```
tbezman, line 13: enables federation in gqlgen

graphql/plugin/remapgen/remapgen.go
```diff
@@ -48,7 +48,9 @@ func (m *Plugin) MutateConfig(cfg *config.Config) error {
 		// Unions and interfaces can have other types assigned to them. If there are 1 or more types in our
 		// schema that can be assigned to this union or interface, create a mapping for it!
 		if (def.Kind == ast.Union || def.Kind == ast.Interface) && len(cfg.Schema.GetPossibleTypes(def)) > 0 {
-			types = append(types, typeName)
+			if typeName != "_Entity" {
+				types = append(types, typeName)
+			}
```
tbezman, line 53: This is to make sure our remapgen file does not include the `_Entity` federation type.

graphql/schema/schema.graphql
```diff
@@ -682,7 +682,7 @@ interface FeedEventData {
   action: Action
 }
 
-type FeedEvent implements Node {
+type FeedEvent implements Node @key(fields: "dbid") {
```
tbezman, line 685: Using `FeedEvent` as a guinea pig for the entity spec

secrets/dev/graphql-gateway.yml
```diff
@@ -0,0 +1,15 @@
+hello: ENC[AES256_GCM,data:9YpFCozFDSQ2nilsJMcjJtUndc1pEbgRXB1T1XrdvDIOWMPUY0HdOPbFmcNwtg==,iv:U3Nz33nnCyqhh2dFxIoTcuyaNJLVhDpOX1ObuQbJHdI=,tag:df5ylgzxPqhBbWQSVKjb8A==,type:str]
```
tbezman, line 1: this is just a non empty yaml file. If you have a suggestion I'd love to hear it.

docker/graphql-gateway/dev/router.yaml
Robinnnnn: i think you may want to drop the period after `gallery`: `^https://gallery*.vercel.app$`

Robinnnnn (approved):
really excited about this terence, you're seriously a beast for getting this up and running so quickly 🙏 

high-level questions:
1. is `@defer` support dependent on the use of an apollo federated supergraph? curious as to why and how these are linked
2. the point of a supergraph seems to be that it can stitch together multiple subgraphs. in this case, we only have a single subgraph that links to our main schema, making it a pretty simple proxy. this simple act provides support for `@defer`?
3. in the future, is the idea to create another subgraph for the mobile client?

graphql/schema/schema.graphql
tbezman: yes

tbezman:
> really excited about this terence, you're seriously a beast for getting this up and running so quickly 🙏
> 
> high-level questions:
> 
> 1. is `@defer` support dependent on the use of an apollo federated supergraph? curious as to why and how these are linked
> 2. the point of a supergraph seems to be that it can stitch together multiple subgraphs. in this case, we only have a single subgraph that links to our main schema, making it a pretty simple proxy. this simple act provides support for `@defer`?
> 3. in the future, is the idea to create another subgraph for the mobile client?

1. Yes. GQLGen does not support defer. So Apollo Router sits in front of GQLGen and splits up the queries before they arrive to GQLGen w/ some extra sauce. This works because GQLGen does support the Federation spec. See excalidraw in original description on how splitting up queries works
2. Yes, this simple act provides support for `@defer`
3. Absolutely not. Unless Ezra feels like getting all microservicy in the future, we'll have one graph for a long time

radazen (approved):
Looks A-OK to me! I'm excited to try it out.

radazen:
> 3. Absolutely not. Unless Ezra feels like getting all microservicy in the future, we'll have one graph for a long time

I definitely do not.

PR #814, opened 2023-02-15
Blurhash + Aspect Ratio Fields

tbezman:
Added a few fields to experiment on React Native. 

- `blurhash` is generated by imgix using the `fm=blurhash` query param. They send the blurhash as raw text in the response so we just read that out of the response body.
- `aspectRatio` comes from the image metadata provided by imgix. We can just grab the `PixelWidth` and `PixelHeight` properties and do some division.

I'm not sure where this will land in the future but I don't think this is the final form.


service/mediamapper/mediamapper.go
```diff
@@ -136,6 +139,68 @@ func (u *MediaMapper) GetSrcSet(sourceUrl string) string {
 	return u.urlBuilder.CreateSrcset(sourceUrl, u.srcSetParams)
 }
 
+func (u *MediaMapper) GetBlurhash(sourceUrl string) *string {
+	token := viper.GetString("IMGIX_SECRET")
+
+	urlBuilder := imgix.NewURLBuilder(assetDomain, imgix.WithToken(token), imgix.WithLibParam(false))
```
radazen, line 145: I don't think you need to grab IMGIX_SECRET or create a new `urlBuilder` here. Should be able to use the existing `u.urlBuilder` so you're not hitting viper for every blur hash URL.

service/mediamapper/mediamapper.go
```diff
@@ -136,6 +139,68 @@ func (u *MediaMapper) GetSrcSet(sourceUrl string) string {
...
+	responseString := string(responseBytes)
+
+	return &responseString
+}
+
+func (u *MediaMapper) GetAspectRatio(sourceUrl string) *float64 {
+	token := viper.GetString("IMGIX_SECRET")
+
+	urlBuilder := imgix.NewURLBuilder(assetDomain, imgix.WithToken(token), imgix.WithLibParam(false))
+	url := urlBuilder.CreateURL(sourceUrl, buildParams(getDefaultParams(), newWidthParam(smallWidth))...)
```
radazen, line 177: Same as above; can probably skip token/urlBuilder assignment and reuse `u.urlBuilder`

service/mediamapper/mediamapper.go
```diff
@@ -107,6 +102,14 @@ func setGoogleWidthParams(sourceUrl string, width int) string {
 	return sourceUrl
 }
 
+func getDefaultParams() []imgix.IxParam {
+	return []imgix.IxParam{
+		imgix.Param("auto", "format", "compress"),
+		imgix.Param("fit", "max"),
+		imgix.Param("fm", "json"),
```
radazen, line 109: I'm pretty sure we _don't_ want `imgix.Param("fm", "json")` to be a default parameter! AFAIK that would make our non-experimental URLs output JSON data about their images instead of the actual image data.

radazen (commented):
Looks good! One concern re: output format, and two trivial comments that don't matter much for an experimental thing, but otherwise LGTM!

service/mediamapper/mediamapper.go
tbezman: BIG OOF. Nice catch ezra. thanks

radazen (approved):
👌👌👌

PR #821, opened 2023-02-18
Notifications graphql

tbezman:
### Useful Links
- [Expo Push Notifications Overview](https://docs.expo.dev/push-notifications/overview/)
- [Expo Push Notifications Go SDK](https://github.com/oliveroneill/exponent-server-sdk-golang)

### What is a push notification token?

- Never expires
- Can use to send a push notification via Expo Push Notification Service
- This token is specific to the expo API, if we want to eject and use the raw services, we can do so with some extra eng lift.

### New Mutation
This new mutation accepts an arbitrary string and appends it to your list of tokens (maybe the user has multiple devices)


graphql/schema/schema.graphql
```diff
@@ -1795,14 +1795,23 @@ union MintPremiumCardToWalletPayloadOrError =
...
+
 type DisconnectSocialAccountPayload {
   viewer: Viewer
 }
 
-union DisconnectSocialAccountPayloadOrError =
-    DisconnectSocialAccountPayload
-  | ErrInvalidInput
-  | ErrNotAuthorized
+union DisconnectSocialAccountPayloadOrError = DisconnectSocialAccountPayload | ErrInvalidInput
```
radazen, line 1814: Looks like `ErrNotAuthorized` was accidentally removed here?

graphql/schema/schema.graphql
```diff
@@ -1795,14 +1795,23 @@ union MintPremiumCardToWalletPayloadOrError =
...
+  token: String!
+}
+
+type RegisterPushNotificationTokenPayload {
+  message: String
+}
+
+union RegisterPushNotificationTokenPayloadOrError =
+    RegisterPushNotificationTokenPayload
+  | ErrNotAuthorized
```
radazen, line 1808: Assuming there's some sort of validation criteria for a push notification token (max length, probably?), we'll want `ErrInvalidInput` here too.

radazen (commented):
Looks pretty straightforward! I think my biggest question is whether push notification tokens belong in the users table, or whether they should have their own `push_notifications` table. I think I lean toward putting them in their own table for a couple of reasons:

- They'll be stored in normalized form (one DB row per token) instead of denormalized form (an inline array of tokens), which AFAIK is the ideal way to do stuff with SQL unless you have a specific reason to use an array type (typically ordering or performance, I think)
- We can soft delete them when they get unregistered, which could be helpful when debugging "I'm not getting notifications issues," since we'd still have a record of the unregistered / now-deleted token

PR #823, opened 2023-02-21
Fix `go generate`

tbezman:
`genqlclient` and `gqlgen` aren't playing nice together. Since we added federation support, `gqlgen` automatically injects some schema stuff for us that `genqlclient` doesn't know about. 

I've fixed this by vendoring the federation schema and telling `genqlclient` about it explicitly.

I also fixed the issue with the missing go dep by adding `go get xxx` to `go:generate` as recommended by this thread https://github.com/99designs/gqlgen/issues/1483#issuecomment-949023482

### DIFF
--- a/gqlgen.yml
+++ b/gqlgen.yml
@@ -1,12 +1,12 @@
 # Where are all the schema files located? globs are supported eg  src/**/*.graphqls
 schema:
-  - graphql/schema/*.graphql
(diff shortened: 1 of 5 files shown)


jarrel-b (commented):
awesome

Robinnnnn (approved):
you're a wizard, harry

PR #826, opened 2023-02-24
Syncing token changes

jarrel-b:
### Changes
The indexer used to call `media.MakePreviewsForMetadata` after a `GET` request for a token. I decided to remove this step because I believe this was causing a race condition where the order of writes to GCP is undefined. I think this caused a bug where Gianna tried to load a token that had no backing media because it was deleted by the indexer after tokenprocessing handled it. This means that the indexer won't have media for a token anymore (only URI and token metadata). This should be simpler to reason about where media comes from because it can only come from either: OpenSea or tokenprocessing (in the case of eth).  For refreshes, only the metadata for a token is refreshed by the indexer. The multichain provider then uses the refreshed metadata to refresh the media.

### Misc changes
* Moved sentry functions with `auth` dependencies to the `auth` package to get around an import cycle
* Moved the fallback provider to a separate file because `multichain.go` was getting big


service/sentry/sentry.go
```diff
@@ -63,26 +60,6 @@ func ReportError(ctx context.Context, err error) {
 	ReportRemappedError(ctx, err, nil)
 }
 
-func ScrubEventCookies(event *sentry.Event, hint *sentry.EventHint) *sentry.Event {
```
jarrel-b, line 66: Moved this stuff to `auth/sentry.go`

service/multichain/fallback.go
```diff
@@ -0,0 +1,96 @@
+package multichain
```
jarrel-b, line 1: This was moved from `multichain.go`

service/auth/sentry.go
```diff
@@ -0,0 +1,52 @@
+package auth
```
jarrel-b, line 1: Moved from `sentry.go` to fix import cycle.

service/multichain/multichain.go
```diff
@@ -208,7 +209,6 @@ var chainValidation map[persist.Chain]validation = map[persist.Chain]validation{
 	},
 	persist.ChainTezos: {
 		tokensFetcher:         true,
-		tokenRefresher:        true,
```
jarrel-b, line 211: Removed this for tezos because we don't have a way to refresh tokens for it.

service/multichain/multichain.go
```diff
@@ -381,39 +381,118 @@ outer:
...
-			tokenIDsToProcess = append(tokenIDsToProcess, token.ID)
+	for i, token := range providerTokens {
+		// Add already existing media to the provider token if it exists so that
+		// we can display media for a token while it gets handled by tokenprocessing
+		if dbToken := tokenLookup[token.TokenIdentifiers()]; !token.Media.IsServable() && dbToken.Media.IsServable() {
+			providerTokens[i].Media = dbToken.Media
+		}
+		// There's no available media for the token at this point, so set the state to syncing
+		// so we can show the loading state instead of a broken token while tokenprocessing handles it.
+		if !providerTokens[i].Media.IsServable() && len(providerTokens[i].TokenMetadata) > 0 {
```
jarrel-b, line 417: We check if `TokenMetadata` is empty because tokenprocessing won't be able to do anything if there's no token metadata anyway and also because we no longer store token URI data for it to fall back to.

service/multichain/multichain.go
```diff
@@ -686,16 +777,16 @@ func (p *Provider) RefreshTokensForContract(ctx context.Context, ci persist.Cont
...
-	allTokens := []chainTokens{}
-	allContracts := []chainContracts{}
+	tokensFromProviders := []chainTokens{}
+	contractsFromProviders := []chainContracts{}
 	tokensReceive := make(chan chainTokens)
 	contractsReceive := make(chan chainContracts)
 	errChan := make(chan errWithPriority)
 	done := make(chan struct{})
 	wg := &sync.WaitGroup{}
-	wg.Add(len(providers))
```
jarrel-b, line 696: Think this could cause a bug where waiting would block forever if not all providers implemented `tokensFetcher`

service/multichain/multichain.go
```diff
@@ -1088,24 +1119,33 @@ func tokensToNewDedupedTokens(ctx context.Context, tokens []chainTokens, contrac
...
-					BlockNumber:          token.BlockNumber,
-					IsProviderMarkedSpam: token.IsSpam,
+			// If we've never seen the incoming token before, then add it.
+			if !seen {
+				seenTokens[ti] = candidateToken
+			} else {
+				// If the candidate token has usable media, then replace the existing token.
+				// We only use the token if there is a way to regenerate the media from the metadata it has.
+				hasMediaLinks := hasMediaURLs(ctx, chainToken.chain, token)
+				if !existingToken.Media.IsServable() && candidateToken.Media.IsServable() && hasMediaLinks {
```
jarrel-b, line 1147: This means we'll only use a token if it has valid media and valid metadata (i.e. way to get media from the metadata). The tradeoff is that we show "Loading..." while the token is being processed by tokenprocessing. It could also mean that we won't show tokens with good media but bad metadata if tokenprocessing fails to handle it.

service/multichain/multichain.go
```diff
@@ -381,39 +381,118 @@ outer:
...
+
+	// Invariant to make sure that its safe to index persistedTokens
+	if len(tokensToUpsert) != len(persistedTokens) {
+		panic("expected the length of tokens inserted to match the input length")
+	}
+
+	errors := make([]error, 0)
+	for userID, offset := range userTokenOffsets {
+		err = p.sendTokensToTokenProcessing(ctx, userID, persistedTokens[offset[0]:offset[1]])
+		errors = append(errors, err)
```
benny-conn, line 453: Should there be a nil check here? If we append every time without a nil check, even nil errors will increase the length of `errors`

benny-conn (approved):
Looks great! Love all of the naming changes, going to start using the word `candidate` in naming myself, great word!

PR #834, opened 2023-02-28
Skip remap for IPFS gateways

jarrel-b:
**Background**:
I noticed for MakersPlace that media is served from their IPFS gateway:
https://ipfsgateway.makersplace.com/ipfs/QmV7Zk9joyNpoi8Q8ZPZL6ywoTyRhbXevYUYvWEXLcovCt

The raw IPFS URI would be:
ipfs://QmV7Zk9joyNpoi8Q8ZPZL6ywoTyRhbXevYUYvWEXLcovCt

And mapped to our gateway would be:
https://gallery.infura-ipfs.io/ipfs/QmV7Zk9joyNpoi8Q8ZPZL6ywoTyRhbXevYUYvWEXLcovCt

Only the MakersPlace URL loads, but the IPFS URI and our gateway times out. I think it's safe to assume that if media doesn't load from the original gateway then its unlikely to load from our gateway unless their gateway is actually down. I think in those cases we would want to actually use our own gateway, but we don't have a mechanism to retry yet since we only have a single timeout per token.

### DIFF
--- a/service/media/media.go
+++ b/service/media/media.go
@@ -359,7 +359,7 @@ func getRawMedia(pCtx context.Context, mediaType persist.MediaType, name, vURL, 
 
 func remapPaths(mediaURL string) string {
 	switch persist.TokenURI(mediaURL).Type() {
-	case persist.URITypeIPFS, persist.URITypeIPFSAPI, persist.URITypeIPFSGateway:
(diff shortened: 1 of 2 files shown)


benny-conn (approved):
Looks great!

PR #836, opened 2023-03-01
Benny/image optim

benny-conn:
Changes:

- **Added** dimension detection to the final step of creating media using a combination of `ffprobe` as well as `iframe` and `svg` XML tags
- **Added** live render url creation for videos, caching a downsized and trimmed (to a height of 720 px and a time length of max 5 seconds) video to be used for live render

Considerations:

- We are going to have to write a script that goes through every token and updates them all with these new fields


cmd/data/farcaster/farcaster.go
```diff
@@ -50,9 +50,9 @@ type ResponseStruct struct {
 }
```
benny-conn, line 51: Disregard these changes on this file, just some minor adjustments for getting data from farcaster, we already ran the script.

service/media/media.go
```diff
@@ -575,6 +745,31 @@ func thumbnailAndCache(ctx context.Context, videoURL, bucket, name string, clien
 	return nil
 }
 
+func createLiveRenderAndCache(ctx context.Context, videoURL, bucket, name string, client *storage.Client) error {
```
jarrel-b, line 748: Can't comment on the relevant line in `MakePreviewsForMetadata`?, but do we also need to remove this file if a live render was created from a prior refresh, but nothing was `cached` on the current refresh?

jarrel-b (commented):
Looks awesome! Excited to see it in action

PR #843, opened 2023-03-06
metadata processing on token processing service

benny-conn:
Changes:

- **Removed** metadata processing logic from the indexer
- **Added** metadata processing logic to the token processing service
- **Added** some minor fixes to the indexer that were causing some traffic jams

The idea behind this change is that now the token processing does not need to worry about the state of the indexer. When tokens are indexed during the indexing process, they are finalized. They have all the necessary fields so that when the token processing server receives all the tokens for a refresh it has all the context it needs to get the media. 

Before this change, a sync could look like this:

1. User syncs
2. Indexer gives new tokens to user but only has token URI stored for the token (no metadata yet)
3. Indexer kicks off async process to find metadata
4. User sends tokens to token processing for media retrieval. (tokens syncing state on frontend)
5. Token processing sees new tokens, but tokens have no metadata, so marks them as invalid
6. User sees token failed to process state
7. User clicks refresh
8. Indexer now has metadata because async process finished
9. Token is valid

After this lands it will look like this:

1. User syncs
2. Indexer gives new tokens to user with only token URI
3. User sends tokens to token processing
4. Token processing uses URI for each token to find the metadata
5. Token processing uses metadata to find media
6. User has their tokens!

Considerations:

- We should keep an eye on how much longer token processing refreshes take because adding in metadata to the mix might increase the time. If it does increase the time noticeably, we could only retrieve metadata if it is empty or maybe if it hasn't been updated in a certain amount of time. I imagine it won't increase the time very much given that the metadata is almost always a short piece of JSON text stored at an IPFS link or http link.


indexer/nfts.go
```diff
@@ -696,8 +671,8 @@ func updateMetadataForTokens(pCtx context.Context, updateChan chan<- tokenFullUp
 	}
 }
 
-func getUpdateForToken(pCtx context.Context, uniqueHandlers uniqueMetadatas, tokenType persist.TokenType, chain persist.Chain, tokenID persist.TokenID, contractAddress persist.EthereumAddress, metadata persist.TokenMetadata, uri persist.TokenURI, ethClient *ethclient.Client, ipfsClient *shell.Shell, arweaveClient *goar.Client) (tokenFullUpdate, error) {
-	newMetadata := metadata
+func getUpdateForToken(pCtx context.Context, tokenType persist.TokenType, chain persist.Chain, tokenID persist.TokenID, contractAddress persist.EthereumAddress, metadata persist.TokenMetadata, uri persist.TokenURI, ethClient *ethclient.Client, ipfsClient *shell.Shell, arweaveClient *goar.Client) (tokenFullUpdate, error) {
```
jarrel-b, line 674: Would we be able to remove the `metadata` arg now too?

service/persist/postgres/postgres.go
```diff
@@ -83,7 +83,7 @@ func (c *connectionParams) toConnectionString() string {
 
 func newConnectionParamsFromEnv() connectionParams {
 	return connectionParams{
-		user:     viper.GetString("POSTGRES_USER"),
+		user:     "postgres",
```
jarrel-b, line 86: Left over from testing?

service/persist/postgres/token_gallery.go
```diff
@@ -83,7 +84,10 @@ func NewTokenGalleryRepository(db *sql.DB, queries *db.Queries) *TokenGalleryRep
 	updateInfoByTokenIdentifiersUnsafeStmt, err := db.PrepareContext(ctx, `UPDATE tokens SET COLLECTORS_NOTE = $1, LAST_UPDATED = $2 WHERE TOKEN_ID = $3 AND CONTRACT = $4 AND DELETED = false;`)
 	checkNoErr(err)
 
-	updateURIDerivedFieldsByTokenIdentifiersUnsafeStmt, err := db.PrepareContext(ctx, `UPDATE tokens SET MEDIA = $1, TOKEN_URI = '', TOKEN_METADATA = $2, NAME = $3, DESCRIPTION = $4, LAST_UPDATED = $5 WHERE TOKEN_ID = $6 AND CONTRACT = $7 AND DELETED = false;`)
+	updateURIDerivedFieldsByTokenIdentifiersUnsafeStmt, err := db.PrepareContext(ctx, `UPDATE tokens SET MEDIA = $1, TOKEN_URI = $2, TOKEN_METADATA = $3, NAME = $4, DESCRIPTION = $5, LAST_UPDATED = $6 WHERE TOKEN_ID = $7 AND CONTRACT = $8 AND DELETED = false;`)
```
jarrel-b, line 87: I think we'll still want to set `TOKEN_URI` to an empty string. The callers might already do that, but nice to be explicit about it!

service/persist/postgres/token_gallery.go
```diff
@@ -83,7 +84,10 @@ func NewTokenGalleryRepository(db *sql.DB, queries *db.Queries) *TokenGalleryRep
...
+	checkNoErr(err)
+
+	updateURITokenIdentifiersUnsafeStmt, err := db.PrepareContext(ctx, `UPDATE tokens SET TOKEN_URI = $1, LAST_UPDATED = $2 WHERE TOKEN_ID = $3 AND CONTRACT = $4 AND DELETED = false;`)
```
jarrel-b, line 90: Also `TOKEN_URI = ''` here too?

tokenprocessing/process.go
```diff
@@ -163,8 +184,8 @@ func processToken(c context.Context, key string, t persist.TokenGallery, contrac
 
 	up := persist.TokenUpdateAllURIDerivedFieldsInput{
 		Media:       newMedia,
-		Metadata:    t.TokenMetadata,
-		TokenURI:    t.TokenURI,
+		Metadata:    newMetadata,
+		TokenURI:    newURI,
```
jarrel-b, line 188: Should this be set?

jarrel-b (approved):
Looks awesome Benny, thanks for getting this done so quickly!

Robinnnnn (commented):
image

^ to clarify, in step 3, does the user (client) itself need to provide the token processing service with a list of tokens it'd like to be processed? i would hope that the indexer can communicate / hand this off directly to the token processor.

> We should keep an eye on how much longer token processing refreshes take because adding in metadata to the mix might increase the time. If it does increase the time noticeably, we could only retrieve metadata if it is empty or maybe if it hasn't been updated in a certain amount of time. I imagine it won't increase the time very much given that the metadata is almost always a short piece of JSON text stored at an IPFS link or http link.

hmm, are token refreshes _blocking_ until all metadata has been processed? or will the frontend be able to poll via Loading state, similar to tezos NFTs? let me know if this is what you're referring to.

benny-conn:
Yes @Robinnnnn, the client itself does nothing, token processing from indexed tokens all occurs on the backend as it is currently.

Yes this will be a part of the polling process, not the initial sync. So it will not block.

PR #849, opened 2023-03-13
find metadata anyway on indexer, fallback to opensea

benny-conn:
Changes:

- **Added** token metadata fetcher to opensea provider
- **Added** manual index step to indexer API metadata fetcher
- **Changed** indexer API metadata fetcher to not return when a token is not found, and instead continue the process of finding metadata using the blockchain


jarrel-b (approved):
lgtm!

PR #850, opened 2023-03-13
Use pg_cron to refresh search relevance tables

radazen:
Now that pg_cron is up and running, it's time to keep our search relevance tables up to date. Thanks @jarrel-b!

This PR refreshes the `user_relevance`, `contract_relevance`, and `gallery_relevance` tables every hour, and staggers the start times (:05, :10, :15) to avoid running a bunch of operations at the same time.


jarrel-b (approved):
whoohooo

PR #851, opened 2023-03-14
Fix missing OS tokens bug

jarrel-b:
There's an issue with the OpenSea API where it doesn't seem to return all tokens if only the owner address is used in the query. Including the owner address and the contract address seems to get around this issue. This PR also wraps calls to OpenSea with retries so that requests retry and backoff if we get rate-limited. It also refactors the pagination to use a loop vs recursion which makes it a little simpler to read (at least for me!)

### DIFF
--- a/indexer/refresh/refresh.go
+++ b/indexer/refresh/refresh.go
@@ -300,11 +300,11 @@ func loadFromFile(path string) (*bloom.BloomFilter, error) {
 	}
 
 	f, err := os.Open(path)
-	defer f.Close()
-
 	if err != nil {
 		return nil, err
 	}
+
+	defer f.Close()
 
 	var bf bloom.BloomFilter
 	bf.ReadFrom(f)
--- a/service/multichain/multichain.go
+++ b/service/multichain/multichain.go
@@ -11,7 +11,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/mikeydub/go-gallery/service/media"
 	"github.com/mikeydub/go-gallery/service/persist/postgres"
 	"github.com/mikeydub/go-gallery/service/redis"
 	"github.com/mikeydub/go-gallery/service/task"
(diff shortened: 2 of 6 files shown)


indexer/refresh/refresh.go
```diff
@@ -300,12 +300,12 @@ func loadFromFile(path string) (*bloom.BloomFilter, error) {
 	}
 
 	f, err := os.Open(path)
-	defer f.Close()
```
jarrel-b, line 303: This was causing deep refreshes to explode if it wasn't able to open the file.

service/multichain/multichain.go
```diff
@@ -1176,13 +1175,8 @@ func tokensToNewDedupedTokens(ctx context.Context, tokens []chainTokens, contrac
 			// If we've never seen the incoming token before, then add it.
 			if !seen {
 				seenTokens[ti] = candidateToken
-			} else {
-				// If the candidate token has usable media, then replace the existing token.
-				// We only use the token if there is a way to regenerate the media from the metadata it has.
-				hasMediaLinks := hasMediaURLs(ctx, chainToken.chain, token)
```
jarrel-b, line 1182: This was a check I added that confirm that we'd only use the token if the metadata is valid. Since we now refresh the metadata at sync time for each token, this check isn't useful anymore.

service/multichain/opensea/opensea.go
```diff
@@ -391,221 +383,182 @@ func verifySignature(pSignatureStr string,
...
+	err = util.UnmarshallBody(&contract, resp.Body)
 	if err != nil {
 		return Contract{}, err
 	}
 
-	return response, nil
+	return contract, nil
+}
+
+func streamAssetsForWallet(ctx context.Context, assetsChan chan<- assetsReceieved, address persist.EthereumAddress) {
```
Robinnnnn, line 524: 😍

Robinnnnn (approved):
the hero we don't deserve 🔥

PR #852, opened 2023-03-14
indexer checks at the end of each pipeline that tokens exist for each log

benny-conn:
Changes:

- **Added** a check at the end of the indexer pipelines that will go through every Ethereum log and ensure that there is a token for the log in the database, logging if not
- **Changed** media processing to not delete videos if thumbnailing fails
- **Changed** wallet hook to match the correct interface definition
- **Changed** polling process in a few ways to simplify things and make everything easier to reason about (and remove chances for race conditions because we were editing a shared struct field from within a goroutine without a mutex): 

1. Only poll in groupings of 50 blocks, just like the indexer does during the catchup process
2. Don't send to transfer processing in groupings of 4, send all together like the indexer does during the catchup process

### DIFF
--- a/indexer/indexer.go
+++ b/indexer/indexer.go
@@ -194,9 +194,6 @@ type indexer struct {
 
 	eventHashes []eventHash
 
-	polledLogs   []types.Log
-	lastSavedLog uint64
-
 	mostRecentBlock uint64  // Current height of the blockchain
 	lastSyncedChunk uint64  // Start block of the last chunk handled by the indexer
 	maxBlock        *uint64 // If provided, the indexer will only index up to maxBlock
@@ -225,8 +222,6 @@ func newIndexer(ethClient *ethclient.Client, ipfsClient *shell.Shell, arweaveCli
 		tokenBucket: viper.GetString("GCLOUD_TOKEN_CONTENT_BUCKET"),
 
 		chain: pChain,
-
-		polledLogs: []types.Log{},
 
 		maxBlock: maxBlock,
 
@@ -290,7 +285,6 @@ func (i *indexer) Start(ctx context.Context) {
 	logger.For(ctx).Info("Catching up to latest block")
 	i.isListening = false
 	i.catchUp(ctx, topics)
-	i.lastSavedLog = i.lastSyncedChunk
 
 	if !rpcEnabled {
 		logger.For(ctx).Info("Running in cached logs only mode, not listening for new logs")
@@ -362,6 +356,7 @@ func (i *indexer) startPipeline(ctx context.Context, start persist.BlockNumber, 
 	plugins := NewTransferPlugins(ctx, i.ethClient, i.tokenRepo, i.addressFilterRepo)
 	enabledPlugins := []chan<- PluginMsg{plugins.balances.in, plugins.owners.in, plugins.uris.in, plugins.refresh.in, plugins.previousOwners.in}
 
+	logsToCheckAgainst := make(chan []types.Log)
 	go func() {
 		ctx := sentryutil.NewSentryHubContext(ctx)
 		span, ctx := tracing.StartSpan(ctx, "indexer.logs", "processLogs")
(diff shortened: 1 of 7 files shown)


indexer/indexer.go
```diff
@@ -382,16 +381,79 @@ func (i *indexer) startNewBlocksPipeline(ctx context.Context, topics [][]common.
...
+	}
+
 }
 
 func (i *indexer) listenForNewBlocks(ctx context.Context) {
 	defer sentryutil.RecoverAndRaise(ctx)
 
 	for {
-		<-time.After(time.Minute * 2)
+		<-time.After(time.Minute * 12)
```
benny-conn, line 456: Increased time here because if we are only processing blocks in groupings of 50 now, we don't need to attempt to process blocks every 2

indexer/indexer.go
```diff
@@ -382,16 +381,79 @@ func (i *indexer) startNewBlocksPipeline(ctx context.Context, topics [][]common.
...
+	}
+
 }
 
 func (i *indexer) listenForNewBlocks(ctx context.Context) {
 	defer sentryutil.RecoverAndRaise(ctx)
 
 	for {
-		<-time.After(time.Minute * 2)
+		<-time.After(time.Second*12*time.Duration(blocksPerLogsCall) + time.Minute)
```
benny-conn, line 456: It takes about 12 seconds to process a block. Since we are only processing in groupings of `blocksPerLogsCall` we should wait that amount of time plus a buffer of a minute.

service/persist/postgres/token.go
```diff
@@ -106,7 +110,7 @@ func NewTokenRepository(db *sql.DB) *TokenRepository {
...
 	checkNoErr(err)
 
-	deleteStmt, err := db.PrepareContext(ctx, `DELETE FROM tokens WHERE TOKEN_ID = $1 AND CONTRACT_ADDRESS = $2 AND OWNER_ADDRESS = $3;`)
+	deleteStmt, err := db.PrepareContext(ctx, `DELETE FROM tokens WHERE TOKEN_ID = $1 AND CONTRACT_ADDRESS = $2 AND OWNER_ADDRESS = $3 and TOKEN_TYPE = $4;`)
```
benny-conn, line 113: This was causing an extremely rare edge case issue where a contract transferred both an ERC-1155 and ERC-721 token to the same owner with the same token ID for both and if the owner transferred away all of the ERC-1155 tokens (had a balance of 0) it would deleted the ERC-721 token as well even if they still owned it.

Why a contract would work this way? Who knows.

indexer/indexer.go
```diff
@@ -382,16 +381,79 @@ func (i *indexer) startNewBlocksPipeline(ctx context.Context, topics [][]common.
...
+					if err != nil {
+						logger.For(ctx).Errorf("error getting balance of ERC1155 token: %s", err)
+						return
+					}
+					if balance.Cmp(big.NewInt(0)) == 0 {
+						logger.For(ctx).Debugf("balance of ERC1155 token is 0, skipping check")
+						return
+					}
+				}
+				exists, err := i.tokenRepo.TokenExistsByTokenIdentifiers(ctx, ti.tokenID, ti.contractAddress)
```
jarrel-b, line 424: Would it make sense to switch the calls so that we first check if it exists and if it does exist we can skip the RPC call? Basically, wondering if this will impact our Alchemy utilization significantly and if it's worth it to try to save on additional calls.

indexer/indexer.go
jarrel-b: Makes sense to me! Is the `+ time.Minute` for some buffer? So we'll probably be relying on OS then for users that mint and want to sync immediately in this 11 min window?

jarrel-b (approved):

indexer/indexer.go
benny-conn: Wonderful idea!

benny-conn:
@jarrel-b most likely it will use opensea but it is also likely that they minted towards the end of the 11 minute window. And yes, the minute is just a buffer.

PR #857, opened 2023-03-15
Remove metadata check

jarrel-b:
Removes the metadata check for when a token is set to the syncing state. If a token has usable media, we'll choose to display that while it gets processed.

### DIFF
--- a/service/multichain/multichain.go
+++ b/service/multichain/multichain.go
@@ -424,7 +424,7 @@ func (p *Provider) prepTokensForTokenProcessing(ctx context.Context, tokensFromP
 		}
 		// There's no available media for the token at this point, so set the state to syncing
 		// so we can show the loading state instead of a broken token while tokenprocessing handles it.
-		if !providerTokens[i].Media.IsServable() && len(providerTokens[i].TokenMetadata) > 0 {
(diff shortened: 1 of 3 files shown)


service/multichain/multichain.go
```diff
@@ -424,7 +424,7 @@ func (p *Provider) prepTokensForTokenProcessing(ctx context.Context, tokensFromP
 		}
 		// There's no available media for the token at this point, so set the state to syncing
 		// so we can show the loading state instead of a broken token while tokenprocessing handles it.
-		if !providerTokens[i].Media.IsServable() && len(providerTokens[i].TokenMetadata) > 0 {
+		if !providerTokens[i].Media.IsServable() {
```
jarrel-b, line 427: Decided to keep this check as `IsServable()` since it's the strictest check we have for whether it should be displayed on the sidebar. Thinking it's better to show something as syncing rather than potentially broken and risk having it not get handled by tokenprocessing or have it flash as broken -> valid.

benny-conn: I'm with it, it doesn't cost much anyway to attempt to refresh the permanently invalid tokens given how often we expect syncs to occur.

benny-conn (approved):

PR #858, opened 2023-03-15
keywords but simplified

benny-conn:
Changes:

- **Changed** token processing handlers to not accept keywords and instead use shared code to determine the keywords for a given change (given that it is always just a few switch statements for what chain and contract it is)


Robinnnnn (commented):
yooo are we still keeping this around?

PR #859, opened 2023-03-16
Add additional details to owned_contracts view

radazen:
## What's new?

The `owned_contracts` view now contains additional columns for `user_created_at` and `displayed_count`. I kept the `displayed` bool because I think it's more convenient for indexing, and there's no danger of it getting out of sync with the `displayed_count` column since this is a materialized view and can't be edited manually.

@jarrel-b I had previously mentioned getting rid of the `tokens.is_user_marked_spam` check, but honestly, I don't think it'd make a meaningful difference for my purposes, so I left it in!

### DIFF
--- a/db/gen/coredb/models_gen.go
+++ b/db/gen/coredb/models_gen.go
@@ -223,11 +223,13 @@ type Notification struct {
 }
 
 type OwnedContract struct {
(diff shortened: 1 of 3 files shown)


jarrel-b (approved):
looks great!

PR #860, opened 2023-03-16
Revert OS changes

jarrel-b:
* Removes including the contract_address in the query param to OS
* Doesn't return if we get an unsupported token type from OS
* Tries to get the first valid metadata from the providers

### DIFF
--- a/service/multichain/opensea/opensea.go
+++ b/service/multichain/opensea/opensea.go
@@ -386,19 +386,9 @@ func verifySignature(pSignatureStr string,
 }
 
 func FetchAssetsForWallet(ctx context.Context, address persist.EthereumAddress) ([]Asset, error) {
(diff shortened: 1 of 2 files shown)


service/multichain/opensea/opensea.go
```diff
@@ -595,7 +585,7 @@ func assetsToTokens(ctx context.Context, ownerAddress persist.Address, assetsCha
 					case "ERC1155":
 						tokenType = persist.TokenTypeERC1155
 					default:
-						errChan <- fmt.Errorf("unknown token type: %s", nft.Contract.ContractSchemaName)
+						logger.For(ctx).Warnf("unknown token type: %s", nft.Contract.ContractSchemaName)
```
jarrel-b, line 588: Changed this to just log a message so that we can process the remaining tokens from OS

Robinnnnn (approved):

PR #862, opened 2023-03-16
possibly invalid refresh token

benny-conn:
Changes:

- **Added** a new error that signals when a user's refresh token is no longer valid

Considerations:

- Requires the frontend to support redirecting to the connect login flow again

### DIFF
--- a/graphql/schema/schema.graphql
+++ b/graphql/schema/schema.graphql
@@ -982,7 +982,7 @@ type SearchCommunitiesPayload {
 union SearchCommunitiesPayloadOrError = SearchCommunitiesPayload | ErrInvalidInput
 
 union SocialQueriesOrError =
-    SocialConnectionsConnection
+    SocialQueries
   | ErrNotAuthorized
   | ErrNeedsToReconnectSocial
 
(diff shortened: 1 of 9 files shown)


graphql/schema/schema.graphql
```diff
@@ -1058,6 +1063,19 @@ type Query {
     first: Int
     last: Int
   ): SocialConnectionsConnection @goField(forceResolver: true) @authRequired
+
+  socialQueries: SocialQueries @authRequired
```
jarrel-b, line 1067: Would this be `SocialQueriesOrError`?

publicapi/socials.go
```diff
@@ -287,6 +283,10 @@ func (api SocialAPI) newTwitterAPIForUser(ctx context.Context, userID persist.DB
 			return nil, fmt.Errorf("error updating social auth: %w", err)
 		}
 	}
+	if err != nil {
```
jarrel-b, line 286: jw why does the error check happen after now?

jarrel-b (approved):

graphql/schema/schema.graphql
benny-conn: Yes! Thanks for that catch

publicapi/socials.go
benny-conn: Because if we receive new social auth tokens, we should always update it to those even if something failed further down the pipeline. That way we can prevent as many occasions of user's having invalid tokens

PR #863, opened 2023-03-16
Benny/token processing improvements

benny-conn:
Changes:

- **Changed** token processing to return media type even when a reader cannot be retrieved, the media could still be valid and renderable by the frontend using the URL directly. It is better than having an unknown media.
- **Changed** metadata processing to check for unknown and invalid URIs so that we can possibly retrieve a better URI to replace it
- **Changed** live render google storage string to the correct one
- **Changed** svg url unescaping to continue even if unescaping fails (sometimes unescaping would fail when the svg doesn't have anything to unescape or includes an un-unescapable character that does not affect the ability for the svg to render)

TODO:

- [x] Figure out how to make token processing not leave any tokens as syncing
- [x] Figure out why token processing can fail to get media during the big sync and successfully get media during an individual sync
- [x] Figure out if imgix caching is the reason why some tokens appear broken on frontend but have completely valid media

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -163,7 +163,7 @@ func SetDefaults() {
 	viper.SetDefault("PORT", 4000)
 	viper.SetDefault("POSTGRES_HOST", "0.0.0.0")
 	viper.SetDefault("POSTGRES_PORT", 5432)
-	viper.SetDefault("POSTGRES_USER", "postgres")
+	viper.SetDefault("POSTGRES_USER", "gallery_backend")
 	viper.SetDefault("POSTGRES_PASSWORD", "")
 	viper.SetDefault("POSTGRES_DB", "postgres")
 	viper.SetDefault("IPFS_URL", "https://gallery.infura-ipfs.io")
--- a/server/server.go
+++ b/server/server.go
@@ -163,7 +163,7 @@ func SetDefaults() {
 	viper.SetDefault("PORT", 4000)
 	viper.SetDefault("POSTGRES_HOST", "0.0.0.0")
 	viper.SetDefault("POSTGRES_PORT", 5432)
-	viper.SetDefault("POSTGRES_USER", "gallery_backend")
+	viper.SetDefault("POSTGRES_USER", "postgres")
 	viper.SetDefault("POSTGRES_PASSWORD", "")
 	viper.SetDefault("POSTGRES_DB", "postgres")
 	viper.SetDefault("IPFS_URL", "https://gallery.infura-ipfs.io")
--- a/service/media/media.go
+++ b/service/media/media.go
@@ -58,11 +58,6 @@ type errUnsupportedMediaType struct {
 	mediaType persist.MediaType
 }
 
-type errGeneratingThumbnail struct {
-	err error
-	url string
-}
-
 type mediaWithContentType struct {
 	mediaType   persist.MediaType
 	contentType string
@@ -228,8 +223,6 @@ func downloadMediaFromURL(ctx context.Context, storageClient *storage.Client, ar
 			}
 		case *net.DNSError:
 			resultCh <- cacheResult{persist.MediaTypeInvalid, cached, err}
-		case errGeneratingThumbnail:
-			resultCh <- cacheResult{mediaType, cached, err}
 		case *googleapi.Error:
 			panic(fmt.Errorf("googleAPI error %s: %s", caught, err))
 		default:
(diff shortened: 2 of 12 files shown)


service/media/media.go
```diff
@@ -564,6 +557,9 @@ func remapPaths(mediaURL string) string {
 func remapMedia(media persist.Media) persist.Media {
 	media.MediaURL = persist.NullString(remapPaths(strings.TrimSpace(media.MediaURL.String())))
 	media.ThumbnailURL = persist.NullString(remapPaths(strings.TrimSpace(media.ThumbnailURL.String())))
+	if !persist.TokenURI(media.ThumbnailURL).IsRenderable() {
```
benny-conn, line 560: If we aren't going to be able to render whatever it is that is stored here, set it to empty so the frontend can at the least attempt to use the media URL instead.

service/multichain/multichain.go
```diff
@@ -1174,7 +1174,15 @@ func tokensToNewDedupedTokens(ctx context.Context, tokens []chainTokens, contrac
 			if !seen {
 				seenTokens[ti] = candidateToken
 			} else if !existingToken.Media.IsServable() && candidateToken.Media.IsServable() {
+				if persist.TokenURI(existingToken.Media.ThumbnailURL).IsRenderable() && !persist.TokenURI(candidateToken.Media.ThumbnailURL).IsRenderable() {
+					candidateToken.Media.ThumbnailURL = existingToken.Media.ThumbnailURL
+				}
```
benny-conn, line 1179: If the current token has a good thumbnail and the new token does not, use the current token's thumbnail.

tokenprocessing/process.go
```diff
@@ -172,6 +172,10 @@ func processToken(c context.Context, key string, t persist.TokenGallery, contrac
 		return nil
 	}
 
+	if !persist.TokenURI(newMedia.ThumbnailURL).IsRenderable() && persist.TokenURI(t.Media.ThumbnailURL).IsRenderable() {
+		newMedia.ThumbnailURL = t.Media.ThumbnailURL
```
benny-conn, line 176: Use the better thumbnail when we can.

server/server.go
```diff
@@ -163,7 +163,7 @@ func SetDefaults() {
 	viper.SetDefault("PORT", 4000)
 	viper.SetDefault("POSTGRES_HOST", "0.0.0.0")
 	viper.SetDefault("POSTGRES_PORT", 5432)
-	viper.SetDefault("POSTGRES_USER", "gallery_backend")
+	viper.SetDefault("POSTGRES_USER", "postgres")
```
jarrel-b, line 166: Should this be `gallery_backend` still so the server isn't run with superuser access?

service/media/media.go
```diff
@@ -985,15 +981,15 @@ func GuessMediaType(bs []byte) (persist.MediaType, string) {
 
 }
 
-func thumbnailVideoToWriter(url string, writer io.Writer) error {
-	c := exec.Command("ffmpeg", "-seekable", "1", "-i", url, "-ss", "00:00:00.000", "-vframes", "1", "-f", "mjpeg", "pipe:1")
+func thumbnailVideoToWriter(ctx context.Context, url string, writer io.Writer) error {
+	c := exec.CommandContext(ctx, "ffmpeg", "-hide_banner", "-loglevel", "error", "-i", url, "-ss", "00:00:00.000", "-vframes", "1", "-f", "mjpeg", "pipe:1")
```
jarrel-b, line 985: Ty for removing the banner 🙌

tokenprocessing/tokenprocessing.go
```diff
@@ -64,7 +64,7 @@ func setDefaults() {
 	viper.SetDefault("GCLOUD_TOKEN_CONTENT_BUCKET", "dev-token-content")
 	viper.SetDefault("POSTGRES_HOST", "0.0.0.0")
 	viper.SetDefault("POSTGRES_PORT", 5432)
-	viper.SetDefault("POSTGRES_USER", "gallery_backend")
+	viper.SetDefault("POSTGRES_USER", "postgres")
```
jarrel-b, line 67: Should this be `gallery_backend`?

tokenprocessing/process.go
jarrel-b: Nice call!

tokenprocessing/process.go
```diff
@@ -132,7 +132,7 @@ func processToken(c context.Context, key string, t persist.TokenGallery, contrac
 		"chain":           t.Chain,
 	})
 	totalTime := time.Now()
-	ctx, cancel := context.WithTimeout(ctx, time.Minute*10)
+	ctx, cancel := context.WithTimeout(ctx, time.Hour)
```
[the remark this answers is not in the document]
jarrel-b: Maybe we can try doubling the time at first? I'm a little worried that in the worse case, all the pool slots are hung waiting for replies from bad servers that will never respond back. Would also be interesting to see what tokens are still timing even after 20+ minutes

jarrel-b (commented):
🚢 just had a few comments/questions but looks great! Checking the thumbnail is such a good call too

PR #867, opened 2023-03-17
Benny/env reliability

benny-conn:
Changes:

- **Added** new `env.Get[T](context.Context, string)` func for retrieving environment variables (don't use `viper.GetString` anymore)
- **Added** new way to use `init()` funcs to establish runtime validation for environment variables

Examples:

## Getting a string Env var
```
func UpdateUser(ctx context.Context, userID persist.DBID) error {
    specialUserEnvField := env.Get[string](ctx, "USER_ENV_FIELD")
    ...
}
```

## Setting validation
```
func init() {
    env.RegisterEnvValidation("REQUIRED_ENV_VAR", []string{"required", "another_validation_tag_maybe"})
}


func UseRequiredEnvVar(ctx context.Context) error {
    weReallyNeedThisSetHere := env.Get[string](ctx, "REQUIRED_ENV_VAR")
    // if this was not set it will error log it using the logger.For(ctx)
    ...
}
```

### DIFF
--- a/admin/snapshot.go
+++ b/admin/snapshot.go
@@ -8,10 +8,14 @@ import (
 
 	storage "cloud.google.com/go/storage"
 	"github.com/gin-gonic/gin"
+	"github.com/mikeydub/go-gallery/env"
 	"github.com/mikeydub/go-gallery/util"
 	"github.com/sirupsen/logrus"
-	"github.com/spf13/viper"
 )
+
+func init() {
+	env.RegisterEnvValidation("SNAPSHOT_BUCKET", []string{"required"})
+}
 
 var rwMutex = &sync.RWMutex{}
 
@@ -55,14 +59,14 @@ func updateSnapshot(stg *storage.Client) gin.HandlerFunc {
 }
 
 func getSnapshotReader(c context.Context, stg *storage.Client) (*storage.Reader, error) {
-	r, err := stg.Bucket(viper.GetString("SNAPSHOT_BUCKET")).Object("snapshot.json").NewReader(c)
+	r, err := stg.Bucket(env.Get[string](c, "SNAPSHOT_BUCKET")).Object("snapshot.json").NewReader(c)
 	if err != nil {
 		return nil, err
 	}
 	return r, nil
 }
 func writeSnapshot(c context.Context, stg *storage.Client, snapshot []string) error {
-	obj := stg.Bucket(viper.GetString("SNAPSHOT_BUCKET")).Object("snapshot.json")
+	obj := stg.Bucket(env.Get[string](c, "SNAPSHOT_BUCKET")).Object("snapshot.json")
 	obj.Delete(c)
 	w := obj.NewWriter(c)
 	w.CacheControl = "no-store"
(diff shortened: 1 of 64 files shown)


env/env.go
```diff
@@ -0,0 +1,70 @@
...
+func init() {
+	v.RegisterValidation("required_for_env", RequiredForEnv)
+}
+
+func RegisterEnvValidation(name string, tags []string) {
+	validators[name] = dedupe(append(validators[name], tags...))
+}
+
+func Get[T any](ctx context.Context, name string) T {
+	for _, tag := range validators[name] {
```
radazen, line 25: I think we probably want a mutex on reads/writes to `validators`, since a new one could technically be registered at any time, and we wouldn't want to be looping/ranging over it on one thread while another is modifying it on another. I think that'd include the deduping on line 21, too, since you wouldn't want to allow the possibility to loop and modify at the same time!

graphql/resolver/handlers.go
```diff
@@ -240,10 +240,11 @@ func AuthRequiredDirectiveHandler() func(ctx context.Context, obj interface{}, n
 }
 
 func RestrictEnvironmentDirectiveHandler() func(ctx context.Context, obj interface{}, next gqlgen.Resolver, allowed []string) (res interface{}, err error) {
-	env := viper.GetString("ENV")
+
 	restrictionErr := errors.New("schema restriction: functionality not allowed in the current environment")
 
 	return func(ctx context.Context, obj interface{}, next gqlgen.Resolver, allowed []string) (res interface{}, err error) {
+		env := env.Get[string](ctx, "ENV")
```
radazen, line 247: I think it's more efficient for this to be in the outer function where it had been previously, since it's not expected to change, and putting it in the outer function means we only need to look it up once!

env/env.go
```diff
@@ -0,0 +1,70 @@
...
+
+var validators = map[string][]string{}
+
+var v = validator.New()
+
+func init() {
+	v.RegisterValidation("required_for_env", RequiredForEnv)
+}
+
+func RegisterEnvValidation(name string, tags []string) {
```
radazen, line 20: Could we rename this to `RegisterValidation`? I misunderstood the function at first -- I thought it was going to register a per-environment validation (e.g. "this value is required, but only when `ENV = production`).

Also, this could be a good case for variadic function parameters! I.e.

`RegisterValidation("MY_ENV_VAR", "required", "unique")`

instead of

`RegisterValidation("MY_ENV_VAR", []string{"required", "unique")}`

radazen (approved):
Looks great! I left a few requests/comments, but I think this is the right direction for our ENV stuff in general!

Honestly, it makes me want to add a linting step in the future, such that we check for every `env.Get` in a package and then make sure whatever `VARIABLE_NAME` it queries for has a validator registered in the package, even if that validator is just a no-op `optional` tag that doesn't do anything beyond documenting the fact that the variable itself is optional. I think we could make it work! Someday...

admin/snapshot.go
```diff
@@ -55,14 +59,14 @@ func updateSnapshot(stg *storage.Client) gin.HandlerFunc {
 }
 
 func getSnapshotReader(c context.Context, stg *storage.Client) (*storage.Reader, error) {
-	r, err := stg.Bucket(viper.GetString("SNAPSHOT_BUCKET")).Object("snapshot.json").NewReader(c)
+	r, err := stg.Bucket(env.Get[string](c, "SNAPSHOT_BUCKET")).Object("snapshot.json").NewReader(c)
```
jarrel-b, line 62: Would we able to add helpers so that we wouldn't need to specify the type hint?
```go
package env

func GetString(ctx context.Context, var string) string {
    return Get[string](c, var)
}
```
(Not actually sure if this will work) Also free to ignore if this is a PITA tho!

PR #870, opened 2023-03-21
Fix syncing bug and only process new tokens on syncs

jarrel-b:
This PR changes tokenprocessing to reply with a `429` status code when a sync is already in progress for a user. Replying with a non-200 status code will re-queue the message for retry so that tokens that are syncing will eventually get processed again.

This PR also ensures that only newly synced tokens are sent to tokenprocessing so that consecutive syncs don't send the same faulty tokens to tokenprocessing each time.

### DIFF
--- a/service/multichain/multichain.go
+++ b/service/multichain/multichain.go
@@ -417,7 +417,6 @@ func (p *Provider) prepTokensForTokenProcessing(ctx context.Context, tokensFromP
 	}
 
 	newTokens := make(map[persist.TokenIdentifiers]bool)
-	syncingTokens := make([]persist.DBID, 0, len(providerTokens))
 
 	for i, token := range providerTokens {
 		existingToken, exists := tokenLookup[token.TokenIdentifiers()]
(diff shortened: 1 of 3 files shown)


service/multichain/multichain.go
```diff
@@ -483,27 +502,25 @@ func (p *Provider) processTokensForUser(ctx context.Context, tokensFromProviders
...
-	err = p.sendTokensToTokenProcessing(ctx, user.ID, persistedTokens)
-	return persistedTokens, err
-}
-
-func (p *Provider) sendTokensToTokenProcessing(ctx context.Context, userID persist.DBID, tokens []persist.TokenGallery) error {
-	tokensToProcess := make([]persist.DBID, 0, len(tokens))
-	for _, token := range tokens {
-		// Process net new tokens based on the creation and update time so that new tokens are handled at least once by tokenprocessing.
-		// Also process tokens that are in a syncing state either from this sync or tokens that were left syncing for whatever reason.
-		if (token.CreationTime.Time() == token.LastUpdated.Time()) || token.Media.MediaType == persist.MediaTypeSyncing {
```
jarrel-b, line 495: Removed this logic up to the callers so that they can decide which tokens to send to tokenprocessing rather than having this method filter tokens.

PR #871, opened 2023-03-21
dummy metadata server

benny-conn:
The purpose of this PR is to create a source of live testing data for our token processing service to use to ensure we get the resulting media's expected from the processing pipeline.

The reason why we can't just use live data is because most live NFTs can change. These will not change. At the very least, we will have the control over whether they change.

Changes:

- **Added** a new dummy metadata server that returns media and metadata in the ways that many tokens might
- **Added** many test medias to the repo


dummymetadata/metadata.go
```diff
@@ -0,0 +1,219 @@
...
+
+func differentKeywordMetadataHandler(c *gin.Context) {
+	mediaURL := fmt.Sprintf("%s/media/image", c.Request.Host)
+
+	if !c.Request.URL.IsAbs() {
+		mediaURL = fmt.Sprintf("%s://%s", c.Request.URL.Scheme, mediaURL)
+	}
+
+	c.JSON(200, gin.H{
+		"image": mediaURL,
```
benny-conn, line 151: `image` instead of `image_url`

dummymetadata/metadata.go
```diff
@@ -0,0 +1,219 @@
...
+
+func wrongKeywordMetadataHandler(c *gin.Context) {
+	mediaURL := fmt.Sprintf("%s/media/video", c.Request.Host)
+
+	if !c.Request.URL.IsAbs() {
+		mediaURL = fmt.Sprintf("%s://%s", c.Request.URL.Scheme, mediaURL)
+	}
+
+	c.JSON(200, gin.H{
+		"image_url": mediaURL,
```
benny-conn, line 163: `image_url` when the media returned is a video (some from of `animation_url` is normally used here)

dummymetadata/media.go
```diff
@@ -0,0 +1,66 @@
...
+func pdfHandler(c *gin.Context) {
+	c.File("./static/test_pdf.pdf")
+}
+
+func textHandler(c *gin.Context) {
+	c.String(200, "I love Gallery!")
+}
+
+func badImageHandler(c *gin.Context) {
+	c.File("./static/test_bad_image.png")
```
benny-conn, line 65: this is a png file where I chopped the file randomly so it won't load correctly. A token could return something but what it returns could be bad

Makefile
```diff
@@ -106,6 +109,15 @@ $(DEPLOY)-%-tokenprocessing            : MEMORY         := $(TOKENPROCESSING_MEM
 $(DEPLOY)-%-tokenprocessing            : CONCURRENCY    := $(TOKENPROCESSING_CONCURRENCY)
 $(DEPLOY)-$(DEV)-tokenprocessing       : SERVICE        := tokenprocessing-dev
 $(DEPLOY)-$(PROD)-tokenprocessing      : SERVICE        := tokenprocessing-v2
+$(DEPLOY)-%-dummymetadata              : REPO           := tokenprocessing
```
jarrel-b, line 112: should be `dummymetadata`?

Makefile
```diff
@@ -82,6 +83,7 @@ $(DEPLOY)-$(PROD)-admin           : SERVICE_FILE := app-prod-admin.yaml
 $(DEPLOY)-$(PROD)-feed            : SERVICE_FILE := feed-env.yaml
 $(DEPLOY)-$(PROD)-feedbot         : SERVICE_FILE := feedbot-env.yaml
 $(DEPLOY)-$(PROD)-tokenprocessing : SERVICE_FILE := tokenprocessing-env.yaml
+$(DEPLOY)-$(PROD)-dummymetadata : SERVICE_FILE := dummymetadata-env.yaml
```
jarrel-b, line 86: Would we need a prod and dev instance since this would be primarily used for testing, or would it be sufficient to just have one instance running somewhere?

docker/dummymetadata/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# Install deps
+WORKDIR /app
+COPY go.mod go.sum /app/
+RUN go mod download
+
+# Install certs
+ADD root-certs /usr/local/share/ca-certificates/
+RUN update-ca-certificates
+
+COPY . /app
```
jarrel-b, line 14: We may need to add exclude `static/` and `dummymetadata/` in the `.dockerignore` file so that Docker will include it

dummymetadata/dummymetadata.go
jarrel-b, line 1: What do you think about moving this so it's not at the same level as the other services? Maybe something like `tests/integration/metadata_server` or something along those lines so it's clear that it's  test related?

jarrel-b (approved):
Looks good to me, just had a few questions! I should be able to set up a live test server during the tests like we do for the backend, so I wouldn't need it deployed somewhere at the moment, but doesn't hurt to have it ready for when we do 👍

PR #875, opened 2023-03-22
use dimensions when available

benny-conn:
Changes:

- **Changed** indexer to no longer return any media

We currently don't use the indexer for media but there is a possibility that the indexer could return some outdated media that would override media created by token processing.


Robinnnnn (approved):

jarrel-b (approved):
🚢

PR #879, opened 2023-03-27
first pass at removing token data logic

benny-conn:
Changes:

- **Removed** token plugins
- **Added** contracts plugin
- **Added** DB hooks that can be added to in the future
- **Added** a single contract DB hook for inserting the contracts into the DB

The `fillContractFields` func which is used in the new contracts plugin will be where we might retrieve the contract owner in the near future.


jarrel-b (approved):
Looks awesome!  Deleting code always feels nice.

PR #880, opened 2023-03-28
update graphql router config to forward cookies

tbezman:
I didn't know when I set this up that headers are not forwarded by default. This PR aims to make sure Cookies are forwarded both ways.

You can read all the docs for what I'm talking about [here](https://www.apollographql.com/docs/router/configuration/header-propagation/).

There's two pieces of configuration related to this. 
1. Request (Client => Server)
  i. This is the `headers` bit in the config. It just tells the router, forward all headers right on over. I'd rather lean towards the side of letting more through, I don't want to have to change this every time there's a new header.
2. Response(Server => Client) 
  i. `rhai`. This is a scripting language built for Rust. They have an example (thank the lord) on how to combine all the cookie headers from the subgraphs into a single one for the requester.


radazen (approved):
Makes sense / looks good! Do you foresee any circumstance where we'll need to update this, or is it on autopilot from here on out?

PR #884, opened 2023-03-28
first pass at alchemy provider

benny-conn:
Changes:

- **Added** alchemy provider with funcs for fetching tokens
- **Removed** indexer related fetch token funcs
- **Removed** media from the `multichain.ChainAgnosticToken` type now that it will never be used. This should speed up refreshes too because providers won't spend time trying to determine media wow!

Considerations:

- Alchemy is very naive with their approach to NFTs, they don't have to think much about how the token's will be used because that is not their responsibility. Where opensea thinks "oh, a lot of people love autoglyphs but their metadata is abnormal, let's manually support it", alchemy goes "this token URI is invalid". We need to re-implement some of our manual handlers for alchemy in a follow up PR probably. And ideally before this makes it into production. We could also have some sort of metadata conglomeration like we used to do with media where if a token from one provider has bad metadata but the other has good metadata, we use the good metadata. (I think combination of both is good)


Robinnnnn (commented):
That is so useful to hear about the alchemy API. Damn, no one does it as good as us!

I wonder how this compares to the Infura API? Curious what a diff would look like for a user like dcinvestor

service/multichain/alchemy/alchemy.go
```diff
@@ -0,0 +1,684 @@
...
+	IsSpam string `json:"isSpam"`
+}
+
+type Token struct {
+	Contract         Contract         `json:"contract"`
+	ID               TokenIdentifiers `json:"id"`
+	Balance          string           `json:"balance"`
+	Title            string           `json:"title"`
+	Description      string           `json:"description"`
+	TokenURI         TokenURI         `json:"owner"`
```
jarrel-b, line 116: Should this be "tokenURI"?

benny-conn:
> > Alchemy is very naive with their approach to NFTs, they don't have to think much about how the token's will be used because that is not their responsibility
> 
> Do you think it is worth having Alchemy as a metadata fetcher? Thinking we could use Alchemy only to tell us what tokens an address has and we can use Opensea and the indexer server as our metadata fetchers. Or is the metadata we fetch from alchemy essentially the same as what we would get from the indexer server anyway since they would just be making the same RPC calls in the end?

At the very worst, they will still return the raw token URI which we will use at the token processing level to determine metadata, and at best they will return pretty straight forward metadata so I'd say it would be fine to try and use their metadata and maybe use opensea's if they don't return any metadata at all.

service/multichain/multichain.go
```diff
@@ -345,17 +329,19 @@ func (p *Provider) SyncTokens(ctx context.Context, userID persist.DBID, chains [
 					return
 				}
 				subWg := &sync.WaitGroup{}
-				subWg.Add(len(providers))
```
benny-conn, line 348: This could have led to an endless wait given that the `wg.Done()` was only called for providers that matched a given interface, good thing all of the providers have matched the interface up until now.

PR #885, opened 2023-03-30
Benny/infura simple provider

benny-conn:
Changes:

- **Added** simple infura provider used for refreshing
- **Added** a `FailureFallbackProvider` type that will be used to default to infura but backup to alchemy on failure
- **Added** a `SubproviderProvider` interface used to allow the fallback providers to pass on their sub providers in the case that the fallback provider itself does not satisfy an interface but it's sub-providers do.
- **Changed** sync to only fail if all providers fail

The purpose of the failure fallback in our use case is to allow infura to be the primary provider for tokens, but if we use up our 100k free requests a day, we backup to alchemy.


service/multichain/infura/infura.go
```diff
@@ -0,0 +1,400 @@
...
+		}
+	} else {
+		if err := json.Unmarshal([]byte(s), &newM); err != nil {
+			return fmt.Errorf("failed to unmarshal metadata from string: %w (%s) (%s)", err, s, string(b))
+		}
+	}
+	*m = newM
+
+	return nil
+}
```
benny-conn, line 63: For some reason, sometimes the metadata is returned as a weird escaped string and sometimes it is correct JSON format, thanks inufra!

service/persist/token.go
```diff
@@ -767,7 +767,7 @@ func (a EthereumAddress) Value() (driver.Value, error) {
 }
 
 // MarshallJSON implements the json.Marshaller interface for the address type
-func (a EthereumAddress) MarshallJSON() ([]byte, error) {
+func (a EthereumAddress) MarshalJSON() ([]byte, error) {
```
benny-conn, line 770: Somehow this evaded us and didn't cause any problems until now, when json.Marshal(EthereumAddress) was called before it would not use this method because of this typo and would just marshal as a plain string.

service/persist/token.go
```diff
@@ -840,12 +840,13 @@ func (m *TokenMetadata) Scan(src interface{}) error {
...
 func (m TokenMetadata) Value() (driver.Value, error) {
-	return m.MarshallJSON()
+	return m.MarshalJSON()
 }
 
-// MarshallJSON implements the json.Marshaller interface for the TokenMetadata type
-func (m TokenMetadata) MarshallJSON() ([]byte, error) {
-	val, err := json.Marshal(m)
+// MarshalJSON implements the json.Marshaller interface for the TokenMetadata type
+func (m TokenMetadata) MarshalJSON() ([]byte, error) {
```
benny-conn, line 847: Same here.

radazen (approved):
Looks great! Thanks for talking through the fallback/subprovider stuff with me!

PR #886, opened 2023-03-31
get contract owner

benny-conn:
Changes:

- **Added** a function that will use 3 methods to determine contract owner
- **Added** functionality to indexer server to return contracts by owner
- **Changed** `creator_address` across the app (except graphql for now to avoid breaking changes) to be `owner_address`. We don't need the contract creator (although often the owner is the creator), we need the owner of the contract for the purpose of artist pages


indexer/contracts.go
```diff
@@ -87,12 +114,42 @@ func updateMetadataForContract(c context.Context, input UpdateContractMetadataIn
...
+	var cmeta alchemy.GetContractMetadataResponse
+	if err := json.NewDecoder(resp.Body).Decode(&cmeta); err != nil {
+		return "", err
+	}
+
+	if cmeta.ContractMetadata.ContractDeployer != "" {
+		return cmeta.ContractMetadata.ContractDeployer, nil
+	}
+	return rpc.GetContractCreator(ctx, address, ethClient)
+}
```
benny-conn, line 155: First use the `ownable` smart contract interface (most accurate owner), if that fails ask alchemy API for contract deployer and if for some reason that fails, try to get the creator the binary search way. We could potentially add in the etherscan API but that seems like overkill.

indexer/nfts.go
benny-conn, line 1: Lots of unused content here now that we don't call these APIs from the backend.

service/persist/token.go
```diff
@@ -250,47 +250,6 @@ func (m Media) IsServable() bool {
 	return m.MediaURL != "" && m.MediaType.IsValid()
 }
 
-// NFT represents an old nft throughout the application
-type NFT struct {
```
benny-conn, line 254: Saw this hanging around, thought I'd drop it 😎

indexer/indexer.go
```diff
@@ -807,6 +818,23 @@ func fillContractFields(ctx context.Context, ethClient *ethclient.Client, contra
 		c.Name = persist.NullString(cMetadata.Name)
 		c.Symbol = persist.NullString(cMetadata.Symbol)
 	}
+
+	cOwner, method, err := GetContractOwner(ctx, contractAddress, ethClient, httpClient)
+	if err != nil {
+		logger.For(ctx).WithError(err).WithFields(logrus.Fields{
```
radazen, line 824: Love how this logging makes use of all the features available to us. I need to use stuff like `WithError` more often!

radazen, line 824: Love how this logging makes use of all the features available to us. I need to use stuff like `WithError` more often!

indexer/nfts.go
radazen, line 1: Thanks for getting rid of all that unused stuff! You are the hero we need.

indexer/t__utils_test.go
```diff
@@ -78,7 +79,7 @@ func newMockIndexer(db *sql.DB, pool *pgxpool.Pool) *indexer {
 	storageClient := newStorageClient(context.Background())
 	bucket := storageClient.Bucket(env.GetString("GCLOUD_TOKEN_LOGS_BUCKET"))
 
-	i := newIndexer(ethClient, nil, nil, nil, postgres.NewTokenRepository(db), postgres.NewContractRepository(db), refresh.AddressFilterRepository{Bucket: bucket}, persist.ChainETH, defaultTransferEvents, func(ctx context.Context, curBlock, nextBlock *big.Int, topics [][]common.Hash) ([]types.Log, error) {
+	i := newIndexer(ethClient, http.DefaultClient, nil, nil, nil, postgres.NewTokenRepository(db), postgres.NewContractRepository(db), refresh.AddressFilterRepository{Bucket: bucket}, persist.ChainETH, defaultTransferEvents, func(ctx context.Context, curBlock, nextBlock *big.Int, topics [][]common.Hash) ([]types.Log, error) {
```
radazen, line 82: Any reason for this to be `http.DefaultClient` when the one we use outside of tests is instantiated with a 10-minute timeout? I'm assuming our tests probably don't have anything that would take 10 minutes, but still curious if there's any reason to deviate from what we do in production.

service/persist/postgres/contract.go
```diff
@@ -80,12 +83,34 @@ func (c *ContractRepository) BulkUpsert(pCtx context.Context, pContracts []persi
...
 // UpdateByAddress updates the given contract's metadata fields by its address field.
 func (c *ContractRepository) UpdateByAddress(ctx context.Context, addr persist.EthereumAddress, up persist.ContractUpdateInput) error {
-	if _, err := c.updateByAddressStmt.ExecContext(ctx, addr, up.Name, up.Symbol, up.CreatorAddress, up.LatestBlock, persist.LastUpdatedTime{}); err != nil {
+	if _, err := c.updateByAddressStmt.ExecContext(ctx, addr, up.Name, up.Symbol, up.OwnerAddress, up.LatestBlock, persist.LastUpdatedTime{}); err != nil {
 		return err
 	}
 	return nil
 }
 
+// GetContractsOwnedByAddress returns all contracts owned by the given address
```
radazen, line 92: Not a "right now" thing, but I'd love to start moving over to sqlc for indexer stuff too!

radazen (approved):
Looks great to me! I left some minor comments here and there.

One general question: is there any reason to keep `CreatorAddress` around instead of getting rid of it everywhere? Does it cost us anything (performance, resources, complexity, etc) to keep tracking it and add `OwnerAddress` as a separate field?

indexer/indexer.go
benny-conn: Thanks to Copilot for recommending me to use it haha

PR #891, opened 2023-04-04
[GAL-490] Add OpenSea created tokens and initial schema for sub-communities

jarrel-b:
#### DB schema changes
`parent_id` is added to `contracts` to refer to the parent contract of a contract, if applicable. Child contracts are inserted into the `contracts` table with reference to the `parent_id`. The `address` column for child contracts is a unique identifier that uniquely identifies the child contract within the parent contract. For Opensea, this is the "slug" of the collection, but it will differ per contract.

`child_contract_id` is added to `tokens` to refer to the child contract that a token may also belong to. I figure that a token can only belong to one sub-contract at most. Resolvers figure out which tokens to fetch based on a flag, `is_root_node`, which is determined by checking if the community being resolved has a parent node. If it is a parent node, queries will fetch tokens matching the `contract` column, and use `child_contract_id` otherwise. 

I didn't find any issues with using a non-ethereum address as the value for the `address` column in my testing, but will definitely keep testing / keep an eye on this.

#### GraphQL schema changes
```graphql
# New fields

type GalleryUser {
  createdCommunities(
    input: CreatedCommunitiesInput!
    before: String
    after: String
    first: Int
    last: Int
  ): CommunitiesConnection @goField(forceResolver: true)
}

type Community {
  creator: GalleryUser @goField(forceResolver: true)
  parentCommunity: CommunityEdge @goField(forceResolver: true)
  subCommunities(before: String, after: String, first: Int, last: Int): CommunitiesConnection
    @goField(forceResolver: true)
}

type Mutation {
  syncCreatedTokens(input: SyncCreatedTokensInput!): SyncCreatedTokensPayloadOrError @authRequired
}
```

#### Multichain changes
`childContractFetcher` is the new provider interface for fetching child contract-related data. The runtime validation was also improved to provide more human-readable errors on failing. Helper methods were also added to help fetch matching providers per chain.

#### OpenSea changes
Previously, all assets needed to be fetched from OpenSea before any processing was done on them. Now, assets are processed as they are received which should help performance. 

A new method, `GetChildContractsCreatedOnSharedContract`, is also added to support fetching child contracts from OpenSea shared contracts. An important thing to note is that the owner of tokens fetched via this method is the user (i.e. the "creator", even if they don't actually own the token). This is subject to change depending on how artist tokens are implemented.

#### TODOs
This PR doesn't handle artist tokens or processing artist tokens after they are synced.

#### Misc changes
* Updates contract bulk upsert to use arrays
* Removes old queries and dataloaders that are no longer used
* Removes deep refresh stuff


jarrel-b:
TODO: Add child contract address to `communityByAddress` query

db/migrations/core/000081_add_creator_contracts.up.sql
```diff
@@ -0,0 +1,8 @@
+alter table contracts add column if not exists parent_id varchar(255) references contracts(id);
+create index if not exists contracts_parent_idx on contracts(parent_id) where deleted = false;
+create unique index if not exists contracts_chain_address_idx on contracts(chain, address) where parent_id is null;
+create unique index if not exists contracts_chain_parent_address_idx on contracts(chain, parent_id, address) where parent_id is not null;
+drop index if exists contract_address_chain_idx;
+
+alter table tokens add column if not exists child_contract_id varchar(255) references contracts(id);
```
radazen, line 7: Following up on our previous discussion: I think `child_contract_id` feels a little redundant because it duplicates a relationship that can already be derived from the `contracts` table. If we can get away with having the existing `tokens.contract` column point to the sub-collection instead of the parent, then querying for all tokens in a top-level contract (assuming only one level of depth in the tree) is just querying for all tokens whose contract or contract's parent is `<someTopLevelContractID>`, and nothing feels redundant.

If we do end up needing to store the parent info in the tokens table -- for performance reasons, to reduce complexity in queries, or otherwise -- then another alternative might be to mirror what the `contracts` table does and store `parent_contract_id`, such that `tokens.contract` always points to the sub-collection, but `parent_contract_id` points to the parent. That way, the logic between the two tables is similar: `token.contract` refers to a contract, and `token.parent_contract_id` refers to the same thing as `token.contract -> parent_id`.

In all cases, I _feel_ like we should be able to internalize `IsRootNode` as a thing the query can account for instead of a thing callers have to determine, but I might be oversimplifying.

service/persist/postgres/contract_gallery.go
```diff
@@ -117,26 +117,65 @@ func (c *ContractGalleryRepository) UpsertByAddress(pCtx context.Context, pAddre
...
+
+	// addIDIfMissing is used because sqlc was unable to bind arrays of our own custom types
+	// i.e. an array of persist.DBIDs instead of an array of strings. A zero-valued persist.DBID
+	// generates a new ID on insert, but instead we need to generate an ID beforehand.
+	addIDIfMissing := func(c *persist.ContractGallery) {
+		if c.ID == persist.DBID("") {
+			(*c).ID = persist.GenerateID()
+		}
 	}
-	sqlStr = sqlStr[:len(sqlStr)-1]
```
radazen, line 132: Getting rid of dynamically generated SQL brings me such joy

service/multichain/multichain.go
```diff
@@ -203,129 +199,170 @@ func NewProvider(ctx context.Context, repos *postgres.Repositories, queries *cor
...
+	TokensContractFetcher bool `json:"tokensContractFetcher"`
+	TokenMetadataFetcher  bool `json:"tokenMetadataFetcher"`
+	ContractRefresher     bool `json:"contractRefresher"`
+	ChildContractFetcher  bool `json:"childContractFetcher"`
 }
 
+// validateProviders verifies that the input providers match the expected multichain spec and panics otherwise
 func validateProviders(ctx context.Context, providers []any) map[persist.Chain][]any {
-	chains := map[persist.Chain][]any{}
+	provided := make(map[persist.Chain][]any)
```
radazen, line 242: As someone who doesn't work with this file very often, I really appreciate the clarity provided by these changes

service/multichain/multichain.go
```diff
@@ -433,21 +456,138 @@ outer:
...
+
+	pResult, err := providerPool.Wait()
+	if err != nil {
+		return err
+	}
+
+	combinedResult := combinedProviderChildContractResults(pResult)
+
+	params := db.UpsertCreatedTokensParams{}
+	now := time.Now()
```
radazen, line 555: Out of curiosity: why `time.Now()` here as opposed to `now()` in the query? If you're worried about consistency, AFAIK `now()` reports the time at the beginning of a SQL transaction, so it should give the same value during all uses in `UpsertCreatedTokens`.

radazen (commented):
Looking really awesome, @jarrel-b! There's a lot of scope to this PR, and I really appreciate how you're not _just_ adding new features; you're also refactoring existing code to make everything easier to follow and easier to maintain. I left some comments with ideas and with some follow-ups to previous discussions, but in general, this looks very solid to me!

db/queries/core/contract_gallery.sql
```diff
@@ -0,0 +1,72 @@
...
+  , unnest(@version::int[])
+  , unnest(@created_at::timestamptz[])
+  , unnest(@address::varchar[])
+  , unnest(@symbol::varchar[])
+  , unnest(@name::varchar[])
+  , unnest(@owner_address::varchar[])
+  , unnest(@chain::int[])
+  , unnest(@description::varchar[])
+)
+on conflict (chain, address) where parent_id is null
```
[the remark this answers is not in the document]
jarrel-b: Yup that's correct, I renamed `UpsertCreatedTokens` -> `UpsertChildContracts` for clarity.

graphql/schema/schema.graphql
```diff
@@ -2121,6 +2155,9 @@ type Mutation {
   setSpamPreference(input: SetSpamPreferenceInput!): SetSpamPreferencePayloadOrError @authRequired
 
   syncTokens(chains: [Chain!]): SyncTokensPayloadOrError @authRequired
+  syncCreatedTokensByUser(
```
jarrel-b, line 2158: This is added as a separate mutation from `syncTokens`, I suppose the frontend could call both? Or we can consolidate to sync created tokens in `syncTokens`? They use separate locks so they can be run concurrently.

graphql/schema/schema.graphql
```diff
@@ -494,7 +511,11 @@ type TokenHoldersConnection {
   pageInfo: PageInfo!
 }
 
-type Community implements Node @goGqlId(fields: ["contractAddress", "chain"]) @goEmbedHelper {
+type CommunityLink {
+  node: Community
+}
+
+type Community implements Node @goEmbedHelper {
```
jarrel-b, line 518: Uses the `dbid` as the node ID

service/persist/postgres/contract_gallery.go
```diff
@@ -117,26 +117,47 @@ func (c *ContractGalleryRepository) UpsertByAddress(pCtx context.Context, pAddre
...
-		sqlStr += generateValuesPlaceholders(7, i*7, nil)
-		vals = append(vals, persist.GenerateID(), contract.Version, contract.Address, contract.Symbol, contract.Name, contract.OwnerAddress, contract.Chain)
-		sqlStr += ","
+
+	contracts := removeDuplicateContractsGallery(pContracts)
+	params := db.UpsertContractsParams{}
+
+	for i := range contracts {
+		c := &contracts[i]
+		params.Ids = append(params.Ids, persist.GenerateID().String())
```
jarrel-b, line 130: In the end I unfortunately wasn't able to get sqlc to handle `persist.DBIDList` correctly. With some indirection, I was able to write a query which worked:
```
insert into contracts(ids) (select unnest(id::varchar[]) from (select @ids id) params)
```

The weird thing is sqlc will:
 * Complain if `@ids` is not aliased to a column in the table that is being update
 * You need to cast `id` again in the outer query even though it should already be an array
 * You occasionally get the error: `ERROR: ON CONFLICT DO UPDATE command cannot affect row a second time` even if the rows being inserted are unique, you are inserting only one row, and even if the row doesn't already exist in the db

service/multichain/multichain.go
```diff
@@ -321,21 +334,135 @@ outer:
...
+				params.CreatorAddress = append(params.CreatorAddress, child.CreatorAddress.String())
+				params.OwnerAddress = append(params.OwnerAddress, child.OwnerAddress.String())
+				params.Chain = append(params.Chain, int32(result.Chain))
+				params.Description = append(params.Description, child.Description)
+				params.ParentIds = append(params.ParentIds, contractToDBID[persist.NewContractIdentifiers(edge.Parent.Address, result.Chain)].String())
+			}
+		}
+	}
+
+	_, err = p.Queries.UpsertChildContracts(ctx, params)
```
jarrel-b, line 459: Will need to upsert creator tokens here after the creator tokens PR is merged, and also submit the tokens to tokenprocessing to process their media.

service/persist/persist.go
```diff
@@ -38,12 +37,6 @@ func (l *DBIDList) Scan(value interface{}) error {
 	return pq.Array(l).Scan(value)
 }
 
-// CreationTime represents the time a record was created
-type CreationTime time.Time
```
jarrel-b, line 42: Removing these types from the codebase!

service/multichain/multichain.go
jarrel-b: Yup was for that reason, good to know!

db/migrations/core/000081_add_creator_contracts.up.sql
jarrel-b: Makes sense! I was avoiding an extra join to contracts this makes thing so much simpler

db/queries/core/contract_gallery.sql
```diff
@@ -0,0 +1,72 @@
...
+    , creator_address = excluded.creator_address
+    , description = excluded.description
+    , last_updated = now()
+  returning *
+)
+insert into contracts(id, deleted, created_at, name, address, creator_address, chain, description, parent_id)
+(
+  select child.id, child.deleted, child.created_at, child.name, child.address, child.creator_address, child.chain, child.description, insert_parent_contracts.id
+  from child_contracts_data child
+  join insert_parent_contracts on child.chain = insert_parent_contracts.chain and child.parent_address = insert_parent_contracts.address
```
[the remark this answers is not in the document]
jarrel-b: This has been updated so that providers can optionally supply the parent contract if its convenient for it to return it too

service/persist/postgres/contract_gallery.go
```diff
@@ -117,26 +117,65 @@ func (c *ContractGalleryRepository) UpsertByAddress(pCtx context.Context, pAddre
...
-		sqlStr += ","
+
+	contracts := removeDuplicateContractsGallery(pContracts)
+	params := db.UpsertContractsParams{}
+	now := time.Now()
+
+	// addIDIfMissing is used because sqlc was unable to bind arrays of our own custom types
+	// i.e. an array of persist.DBIDs instead of an array of strings. A zero-valued persist.DBID
+	// generates a new ID on insert, but instead we need to generate an ID beforehand.
+	addIDIfMissing := func(c *persist.ContractGallery) {
```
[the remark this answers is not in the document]
jarrel-b: In the end I unfortunately wasn't able to get sqlc to handle `persist.DBIDList` correctly. With some indirection, I was able to write a query which worked:
```
insert into contracts(ids) (select unnest(id::varchar[]) from (select @ids id) params)
```

The weird thing is sqlc will:
 * Complain if `@ids` is not aliased to a column in the table that is being update
 * You need to cast `id` again in the outer query even though it should already be an array
 * You occasionally get the error: `ERROR: ON CONFLICT DO UPDATE command cannot affect row a second time` even if the rows being inserted are unique, you are inserting only one row, and even if the row doesn't already exist in the db

publicapi/user.go
```diff
@@ -805,7 +806,71 @@ func (api UserAPI) SharedCommunities(ctx context.Context, userID persist.DBID, b
...
+	}
+
+	queryFunc := func(params timeIDPagingParams) ([]any, error) {
+		serializedChains := make([]string, len(includeChains))
+		for i, c := range includeChains {
+			serializedChains[i] = strconv.Itoa(int(c))
+		}
+		keys, err := api.loaders.ContractsLoaderByCreatorID.Load(db.GetCreatedContractsBatchPaginateParams{
+			UserID:           userID,
+			Chains:           strings.Join(serializedChains, ","),
```
jarrel-b, line 830: I realized I actually did get around the map hashable issue by serializing the chains before, is there a better way to do this in your PR?

db/queries/core/contract_gallery.sql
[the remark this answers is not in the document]
radazen: Awesome, thanks!

publicapi/user.go
radazen: Nope! it was an "I'll add this while I'm here" thing in my PR, but wasn't required functionality, so I just got rid of it when I ran into this issue. I'll use this approach when I end up adding it!

graphql/schema/schema.graphql
jarrel-b: makes sense!

PR #896, opened 2023-04-05
find and present fallback media for tokens when possible

benny-conn:
Changes:

- **Added** `fallback_media` column to gallery tokens
- **Added** `FallbackMedia` subtype to GQL schema
- **Added** necessary multi-chain provider code to each of the providers to create a valid fallback media
- **Added** step at resolver to return fallback media when regular media is not servable but fallback media is

Considerations:

- Should fallback media always be used when the regular media is not servable and the fallback is? Or should it only be used when the regular media is syncing, in which case it is just a syncing fallback.


service/persist/token_gallery.go
```diff
@@ -82,28 +83,6 @@ type ContractIdentifiers struct {
 	Chain           Chain   `json:"chain"`
 }
 
-// TokenInCollection represents a token within a collection
-type TokenInCollection struct {
-	ID           DBID         `json:"id" binding:"required"`
-	CreationTime CreationTime `json:"created_at"`
```
benny-conn, line 88: Got rid of this and everywhere it was being used... because it was really not being used wherever it was being used lol

graphql/resolver/schema.resolvers.helpers.go
```diff
@@ -1858,6 +1858,11 @@ func getUrlExtension(url string) string {
 
 func getMediaForToken(ctx context.Context, token db.Token) model.MediaSubtype {
 	med := token.Media
+
+	if !med.IsServable() && token.FallbackMedia.IsServable() {
+		return getFallbackMedia(ctx, token.FallbackMedia)
+	}
+
```
benny-conn, line 1865: the sus code in question 🤔

Robinnnnn (commented):
awesome work benny!! thanks a lot for handling this for all our multichain providers

re: only having FallbackMedia for syncing tokens – as opposed to any unservable media – agreed with your take, i think it'll make the logic cleaner ✅ 

that said, i don't see a huge problem with keeping FallbackMedia around post-sync

radazen (approved):
This is very cool! Seems like it'll really improve the user experience here.

My only design question: does `FallbackMedia` need its own type, or should it be a field on media types that might have/need a fallback? I could be misreading, but it looks to me like `UnknownMedia` could return a `FallbackMedia` if we have a servable URL for it. If that happened, would the frontend be able to differentiate between syncing and unknown media, given that both of them could return a `FallbackMedia`? Would it ever want to? Basically, I'm just wondering if there's value in saying "this media is syncing but you can check its fallback field to see if there's something you can display in the meantime" instead of a more generic "this is fallback media."

PR #901, opened 2023-04-06
Feat/metadata

jacohend:
Update contract metadata from contractURIs


service/rpc/rpc.go
```diff
@@ -344,6 +344,106 @@ func GetMetadataFromURI(ctx context.Context, turi persist.TokenURI, ipfsClient *
 
 }
 
+// GetMetadataFromURI parses and returns the NFT metadata for a given token URI
+func GetMetadataFromContractURI(ctx context.Context, curi persist.TokenURI, ipfsClient *shell.Shell, arweaveClient *goar.Client) (persist.ContractMetadata, error) {
```
benny-conn, line 348: If possible, since this logic mimics the logic for token URIs, could we either rename `persist.TokenMetadata` to `persist.Metadata` given that it is the same underlying type as `persist.ContractMetadata` and use the current `GetMetadataFromURI` functions or make the original `GetMetadataFromURI` generic so that it can take in different generic map types? That way we can share the logic.

benny-conn (commented):
Thanks so much for submitting a PR! Looks great.

Could I get a little more clarity on the purpose of this PR? I think I have heard about the `contractURI()` interface, not sure how many contracts currently implement it, but what information would we expect to get in the contract metadata that would be useful for displaying on the Gallery site? Would this metadata replace the current `name()` and `symbol()` for some contracts, include more information like a description, etc.?

jacohend:
Sorry, wrong repo!

PR #909, opened 2023-04-09
Make sure tokenprocessing (and other services) can find a Sentry hub

radazen:
The `tokenprocessing` service was failing to report some errors to Sentry because retrieving a `sentry.Hub` from the available `context` was failing. This happens because of an annoying context split: each request has a `gin.Context` and an `http.Request.Context`. When using GraphQL, gqlgen passes the `http.Request.Context` to resolvers. In other scenarios (like token processing), we may pass the `gin.Context` to handlers. The existing flow was only set up to find a Sentry hub from a `http.Request.Context`; this PR adds handling to retrieve the Sentry hub from a `gin.Context` as well.

**Notable change:** `util.GinContextFromContext` has been renamed to `util.MustGetGinContext`, which is a more idiomatic golang way to say "get this or panic." There's also a new variant, `util.GetGinContext`, which will return nil instead of panicking. We'll still want to use the panic version for most purposes, since the lack of a Gin context usually indicates that something is very wrong, but there are some scenarios (like our Sentry handling) where the existence of a Gin context should be optional (e.g. maybe we want to call Sentry from a service that doesn't handle HTTP requests and doesn't use Gin, and that shouldn't cause a panic).

### DIFF
--- a/service/sentry/sentry.go
+++ b/service/sentry/sentry.go
@@ -37,7 +37,7 @@ var logToSentryLevel = map[logrus.Level]sentry.Level{
 var sentryTrailLimit = 8
 
 func ReportRemappedError(ctx context.Context, originalErr error, remappedErr interface{}) {
-	hub := sentry.GetHubFromContext(ctx)
+	hub := SentryHubFromContext(ctx)
 	if hub == nil {
 		logger.For(ctx).Warnln("could not report error to Sentry because hub is nil")
 		return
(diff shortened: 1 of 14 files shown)


service/sentry/sentry.go
```diff
@@ -155,20 +155,23 @@ func NewSentryHubContext(ctx context.Context) context.Context {
 }
 
 // SentryHubFromContext gets a Hub from the supplied context, or from an underlying
-// gin.Context if one is available. NOTE: once gin 1.7.8 is released, this method can
```
radazen, line 158: I was wrong about this. Though gin 1.7.8 _did_ include support for pulling values from the request context, it also started falling back to the request context for other context stuff (deadline and done). That ended up being a breaking change (people complained that their contexts were ending sooner than expected since they were tied to the request), so this behavior became opt-in, and I'm reluctant to opt in and expose us to potential breaking changes atm!

benny-conn (approved):
Context Ginception! Looks great to me!

PR #912, opened 2023-04-12
Swap order of owner / contract address parameters

radazen:
These params were in the wrong order, which was causing some lookups to fail


kaitoo1 (approved):

Robinnnnn (approved):

benny-conn (approved):
Yep, that is correct!

PR #929, opened 2023-04-27
Benny/processing pipeline

benny-conn:
Changes:

- **Created** `pipeline.go` which holds the entire token processing pipeline
- **Created** a `tokenProcessor` type for holding clients necessary for token processing and `tokenProcessingJob` for holding the information necessary for processing a single token, pipeline metadata, and all the methods for running the pipeline
- **Created** two tables `token_processing_jobs` for the job data primarily for debug and `token_medias` for holding the medias for a given token
- **Created** new field on tokens for referencing the `token_medias` table
- **Created** `persist.PipelineMetadata` and necessary tracking functions to track the statuses of individual steps in the token processing pipeline (could be called `persist.TokenProcessingPipelineMetadata`?)
- **Created** `persist.TokenProperties` for tracking the final properties of a completely processed token

TODO:

- [x] Update tokens new media field when synced

### Summary by OpenAI

Release Notes:

- New Feature: Adds a new token processing pipeline with functions to extract, transform, and load token data. Also adds new API endpoints to retrieve token data and metadata.
- Refactor: Modifies the database schema to add new tables and fields for token processing and metadata tracking.

> "Tokens processed with ease,
> Metadata tracked with breeze,
> Our codebase now refined,
> Bugs and issues left behind."

@openai: ignore


benny-conn:
@openai: ignore

tokenprocessing/pipeline.go
```diff
@@ -0,0 +1,332 @@
...
+	})
+
+	return errGroup.Wait()
+}
+
+func (tpj *tokenProcessingJob) updateTokenMetadataDB(ctx context.Context, tmetadata coredb.TokenMedia) error {
+	defer persist.TrackStepStatus(&tpj.pipelineMetadata.UpdateTokenMetadataDB)()
+	err := func() error {
+		if !tmetadata.Active {
+			return tpj.tp.queries.InsertTokenMedia(ctx, coredb.InsertTokenMediaParams{
```
jarrel-b, line 239: Noticed that `InsertTokenMedia` hard-codes the `active` column to `true`, should that still be the case even if the incoming metadata isn't active?

db/queries/core/token_gallery.sql
```diff
@@ -78,12 +85,14 @@ insert into tokens
...
-      , unnest(@chain::int[]) as chain
-      , unnest(@contract::varchar[]) as contract
       , unnest(@is_user_marked_spam::bool[]) as is_user_marked_spam
       , unnest(@is_provider_marked_spam::bool[]) as is_provider_marked_spam
       , unnest(@last_synced::timestamptz[]) as last_synced
       , unnest(@token_uri::varchar[]) as token_uri
+      , unnest(@token_id::varchar[]) as token_id
+      , unnest(@contract::varchar[]) as contract
+      , unnest(@chain::int[]) as chain
+      , unnest(tms.id) as media_id from tms
```
jarrel-b, line 95: It's possible to have no metadata for a token at the point this query is run, right? If so, seems like `tms` can be a different length from the token/address/contract inputs causing the media ids to be offset from the original inputs. I think rather than using `unnest`, we'll want to do a join either in the `tms` cte so that missing media_ids are filled with `null`s or join between `bulk_upsert` and `tms` to ensure the correct `media_id` is used!

jarrel-b (commented):
Looking awesome so far! I still need to look at `tokenprocessing/media.go`, but everything looks sweet. I especially like how the pipeline statuses are handled!

tokenprocessing/pipeline.go
```diff
@@ -0,0 +1,332 @@
...
+
+	result.Name, result.Description = tpj.retrieveTokenInfo(ctx, result.Metadata)
+
+	cachedObjects, err := tpj.cacheMediaObjects(ctx, result.Metadata)
+	if err != nil {
+		isSpam := tpj.contract.IsProviderMarkedSpam || util.GetOptionalValue(tpj.token.IsProviderMarkedSpam, false) || util.GetOptionalValue(tpj.token.IsUserMarkedSpam, false)
+		reportTokenError(ctx, err, tpj.id, tpj.token.Chain, tpj.contract.Address, tpj.token.TokenID, isSpam)
+		switch err.(type) {
+		case errNotCacheable:
+			errNotCacheable := err.(errNotCacheable)
```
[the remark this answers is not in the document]
benny-conn: I wish it would auto assert here!

service/multichain/multichain.go
```diff
@@ -462,9 +462,11 @@ func (p *Provider) prepTokensForTokenProcessing(ctx context.Context, tokensFromP
 
 	for i, token := range providerTokens {
 		existingToken, exists := tokenLookup[token.TokenIdentifiers()]
+
 		// Add already existing media to the provider token if it exists so that
 		// we can display media for a token while it gets handled by tokenprocessing
 		if !token.Media.IsServable() && existingToken.Media.IsServable() {
+			// TODO remove
```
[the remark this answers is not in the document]
benny-conn: Added these for later since we won't be checking this field here when we completely move to the new table, but should stay for now for the sake of backwards compat

db/queries/core/token_gallery.sql
benny-conn: Wonderful! Thank you for helping me figure this one out!

tokenprocessing/media.go
```diff
@@ -0,0 +1,1167 @@
...
+	}
+
+	if retryOpensea {
+		defer persist.TrackStepStatus(subMeta.OpenseaFallback)()
+		logger.For(pCtx).Infof("failed to get data from uri '%s' for '%s' because of (err: %s <%T>), trying opensea", mediaURL, tids, err, err)
+		// if token is ETH, backup to asking opensea
+		assets, err := opensea.FetchAssetsForTokenIdentifiers(pCtx, persist.EthereumAddress(tids.ContractAddress), opensea.TokenID(tids.TokenID.Base10String()))
+		if err != nil || len(assets) == 0 {
+			// no data from opensea, return error
+			return nil, errNoDataFromReader{err: err, url: mediaURL}
```
jarrel-b, line 955: Would it be worth it to include in the error that we also tried OS as a fallback at this? I think it'll be helpful for debugging to know that even OS didn't have it.

tokenprocessing/media.go
```diff
@@ -0,0 +1,1167 @@
...
+	switch mediaType {
+	case persist.MediaTypeHTML:
+		return getHTMLMedia(pCtx, tids, tokenBucket, animURL, imgURL, objects)
+	default:
+		panic(fmt.Sprintf("media type %s should be cached", mediaType))
+	}
+
+}
+
+func createMediaFromCachedObjects(ctx context.Context, tokenBucket string, objects []CachedMediaObject) persist.Media {
```
jarrel-b, line 247: feel free to ignore: Rather than passing a slice of cached objects, I think passing in a map of object type -> object or a struct of object type fields -> object would improve readability on how a media object is built up from various cached objects. It might also be slightly performant too since we wouldn't need to iterate through the slice to find the appropriate object

tokenprocessing/pipeline.go
```diff
@@ -0,0 +1,337 @@
...
+	cachedObjects, err := tpj.cacheMediaObjects(ctx, result.Metadata)
+	if err != nil {
+		isSpam := tpj.contract.IsProviderMarkedSpam || util.GetOptionalValue(tpj.token.IsProviderMarkedSpam, false) || util.GetOptionalValue(tpj.token.IsUserMarkedSpam, false)
+		switch err.(type) {
+		case errNotCacheable:
+			errNotCacheable := err.(errNotCacheable)
+			first, _ := util.FindFirst(cachedObjects, func(c CachedMediaObject) bool {
+				return c.StorageURL(tpj.tp.tokenBucket) != "" && (c.ObjectType == ObjectTypeImage || c.ObjectType == ObjectTypeSVG || c.ObjectType == ObjectTypeThumbnail)
+			})
+			result.Media = tpj.createRawMedia(ctx, errNotCacheable.MediaType, errNotCacheable.URL, first.StorageURL(tpj.tp.tokenBucket), cachedObjects)
```
jarrel-b, line 137: Just wanted to double-check my understanding, if we catch `errNotCacheable`, is `cachedObjects` expected to be an empty slice every time? Is there a difference between "raw" media created via `createRawMedia` and media that gets created via `createMediaFromCachedObjects`? I'm wondering if we could eliminate the concept of "raw" if there's not a significant difference between the two scenarios

db/queries/core/query.sql
```diff
@@ -1009,6 +1009,23 @@ insert into users (id, username, username_idempotent, bio, wallets, universal, e
...
+
+-- name: InsertTokenMedia :exec
+insert into token_medias (id, contract, token_id, chain, metadata, media, name, description, processing_job_id, active) values ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10);
+
+-- name: UpdateTokenTokenMediaByTokenIdentifiers :exec
+update tokens set token_media = $1 where tokens.contract = $2 and tokens.token_id = $3 and tokens.chain = $4 and deleted = false;
+
+-- name: UpdateActiveTokenMediaByTokenIdentifiers :exec
+with old as (
+  insert into token_medias (id, contract, token_id, chain, metadata, media, name, description, processing_job_id, active, created_at, last_updated) (select $1, contract_address, token_id, chain, metadata, media, name, description, processing_job_id, false, created_at, last_updated from token_medias where token_medias.contract = $2 and token_medias.token_id = $3 and token_medias.chain = $4 and active = true and deleted = false)
```
radazen, line 1023: Huh! `contract_address` isn't a column on `token_medias`. It looks like sqlc did compile this and generate a statement for it, but I'm very surprised that it did; I would have expected it to error out on this!

tokenprocessing/pipeline.go
benny-conn: We could technically cache something on either side of the two caching steps (image, animation) so if we encounter this error we just know that we specifically encountered something not to be cached (on purpose) but something else like a thumbnail could have been. Will make this more clear in a comment

radazen (approved):
This is really great! I think everything is structured well and sets us up for a much cleaner token processing pipeline. Awesome work, @benny-conn!

I left some comments -- mostly minor, with an idea to combine some SQL queries -- but @jarrel-b already got the issue I mentioned during standup (unnesting the `tms` CTE), so I don't have any real functional concerns at this point!

db/queries/core/token_gallery.sql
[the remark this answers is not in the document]
benny-conn: Can't comment on the next comment for some reason, but I asked chat GPT about how to limit `tms` to one just in case things change in the future and it came up with just another sub-query to solve it. Here is the final CTE group:

```sql
with tids as (
  select unnest(@token_id::varchar[]) as token_id, unnest(@contract::varchar[]) as contract, unnest(@chain::int[]) as chain
)
, limited_tms as (
  select
    tids.token_id,
    tids.contract,
    tids.chain,
    token_medias.id as media_id,
    ROW_NUMBER() OVER (PARTITION BY token_medias.id ORDER BY token_medias.last_updated) AS row_num
  from
    tids
    left join token_medias on (
      token_medias.token_id = tids.token_id
      and token_medias.contract = tids.contract
      and token_medias.chain = tids.chain
      and token_medias.active = true
      and token_medias.deleted = false
    )
)
, tms as (
  select
    token_id,
    contract,
    chain,
    media_id
  from
    limited_tms
  where
    row_num = 1
)
-- query here
```

PR #931, opened 2023-04-28
Push Notifications

radazen:


feed/handler.go
```diff
@@ -35,7 +36,7 @@ func taskRequired() gin.HandlerFunc {
 }
 
 func handlersInit(router *gin.Engine, queries *db.Queries, taskClient *cloudtasks.Client) *gin.Engine {
-	router.GET("/ping", ping())
+	router.GET("/ping", util.HealthCheckHandler())
```
radazen, line 39: Figured I'd standardize on the util method Benny added for health checks!

feedbot/handler.go
```diff
@@ -2,48 +2,30 @@ package feedbot
...
-		}
-
 		creds := c.Request.Header.Get("Authorization")
 		if creds != "Basic "+env.GetString("FEEDBOT_SECRET") {
+			// Return 200 on auth failures to prevent task retries
 			c.AbortWithError(http.StatusOK, errors.New("unauthorized request"))
 			return
 		}
 	}
 }
```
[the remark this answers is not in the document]
radazen: @openai: ignore

feedbot/handler.go
```diff
@@ -2,48 +2,30 @@ package feedbot
...
-}
-
-// TaskRequired checks that the request comes from Cloud Tasks and does a basic auth check.
-// Returns a 200 status to remove the message from the queue if it is a bad request.
-func TaskRequired() gin.HandlerFunc {
+// TODO: Use middleware.BasicHeaderAuthRequired. Requires reworking the sender to use
+// the appropriate header/encoding.
+func authRequired() gin.HandlerFunc {
 	return func(c *gin.Context) {
-		taskName := c.Request.Header.Get("X-CloudTasks-TaskName")
```
radazen, line 25: I moved the task-handling checks into the `middleware`, since I wanted to use it for my tasks too! The auth header isn't quite the same as my generic `basicauth` middleware, though, so I kept that part here for now.

middleware/middleware.go
```diff
@@ -36,6 +40,72 @@ func AdminRequired() gin.HandlerFunc {
 	}
 }
 
+type BasicAuthOptionBuilder struct{}
```
radazen, line 43: Figured I'd experiment with a pattern here. I like functional options, but I don't like how the options themselves pollute the package namespace and don't feel scoped to the functions they're intended for. So what if we scope them to an "option builder" struct for the given type? It might be a dumb idea, but it does feel more organized than the alternative. Feedback appreciated!

middleware/middleware.go
```diff
@@ -15,14 +16,17 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/mikeydub/go-gallery/service/auth"
-	"github.com/mikeydub/go-gallery/service/persist"
 	"github.com/mikeydub/go-gallery/util"
 )
 
 var mixpanelDistinctIDs = map[string]string{}
 
-type errUserDoesNotHaveRequiredNFT struct {
```
radazen, line 24: We don't use this error anymore

docker/pushnotifications/Dockerfile
jarrel-b, line 1: You'll want to add the `pushnotifications` dir to `.dockerignore` because we exclude all files except the ones specified in the file in the build step.

secrets/prod/makevars.yaml
```diff
@@ -42,6 +42,11 @@ GRAPHQL_GATEWAY_TIMEOUT: ENC[AES256_GCM,data:q3hQ3rA=,iv:E9XMCyK4GvAln5XxS7NYM6m
 GRAPHQL_GATEWAY_CPU: ENC[AES256_GCM,data:f1r1hTM=,iv:w1dVsIHPOgIZp75GQdVNRsII6mQ8d5w0qdDMUEG4UKs=,tag:tSKb563Rffvr5DZxQpo26g==,type:str]
 GRAPHQL_GATEWAY_MEMORY: ENC[AES256_GCM,data:tsU9,iv:jSRQ21J179wbFa09SlOyF34UAl7vLb/PRPzOM7NbjlY=,tag:EU6oMzcBxtOoIqr+iezt7g==,type:str]
 GRAPHQL_GATEWAY_CONCURRENCY: ENC[AES256_GCM,data:DXwHzt9JXA==,iv:xE/s5s4pvM6ztT9NPnu3Mn8PdJGN+PzAzk6fUluqlDA=,tag:q3KGpz3xJxVkPFTfT3/agQ==,type:str]
+PUSH_NOTIFICATIONS_TIMEOUT: ENC[AES256_GCM,data:M4UuJ+w=,iv:rfBvxEMY6DXZdEqcMVmy+/cg4ePNZVRbWARsLSFn24I=,tag:WQrrF2dvfqlnK9kfzSxYsg==,type:str]
+PUSH_NOTIFICATIONS_CPU: ENC[AES256_GCM,data:RGpmt+U=,iv:1s6pW9XcAsNaS8ooPRf9FpBeWeW9h4lFgVtk0ELvT4U=,tag:ZhtNxSmD27Y82tl5AUFJNw==,type:str]
+PUSH_NOTIFICATIONS_MEMORY: ENC[AES256_GCM,data:QYxM,iv:B0Z0hyc91C2EvqNN4uKJh+gtsqb708YPcyA7Qs4Tr2o=,tag:k8XaoxyhIruOGRXcJH+kNQ==,type:str]
+PUSH_NOTIFICATIONS_CONCURRENCY: ENC[AES256_GCM,data:7b69Zt6lzQ==,iv:0N7pitqQoaPPefz+vI67yUz4gA2gtJEzcvSkM0AIRFo=,tag:YqUbZ9UuF5+teOFjYmfLTQ==,type:str]
+PUSH_NOTIFICATIONS_SECRET: ENC[AES256_GCM,data:5JdGV8SiDnAV3kwJX3G3N5ogYpaqyeenpKroCX4bvMs=,iv:MucE1ANX0hq8AOAARmNjqomR0IC5Ng8B79+Xv9LWQF8=,tag:7x1oVYx3nMJqKGSNcZsVYQ==,type:str]
```
jarrel-b, line 49: Does this need to be in `makevars.yaml` and `backend-env.yaml`?

pushnotifications/expo/handler.go
```diff
@@ -0,0 +1,339 @@
...
+}
+
+func (h *PushNotificationHandler) CheckPushTickets() error {
+	return h.checkPushTicketsRecursive(h.ctx, 0)
+}
+
+func (h *PushNotificationHandler) checkPushTicketsRecursive(ctx context.Context, recursionDepth int) error {
+	if recursionDepth >= maxCheckPushTicketsRecursions {
+		err := fmt.Errorf("checkPushTicketsRecursive: exceeded recursion depth (%d iterations)", maxCheckPushTicketsRecursions)
+		reportError(ctx, err)
```
jarrel-b, line 234: Should we return here?

middleware/middleware.go
jarrel-b: I'm all for using structs to namespace functions!

publicapi/user.go
```diff
@@ -1167,3 +1168,86 @@ func (api UserAPI) RecommendUsers(ctx context.Context, before, after *string, fi
...
+		return db.PushNotificationToken{}, err
+	}
+
+	userID, err := getAuthenticatedUserID(ctx)
+	if err != nil {
+		return db.PushNotificationToken{}, err
+	}
+
+	// Does the token already exist?
+	token, err := api.queries.GetPushTokenByPushToken(ctx, pushToken)
```
jarrel-b, line 1188: Is it common for tokens to get re-used, or is this more of an edge case that we should definitely check for? I think there's an unlikely race condition where an instance sees that a token isn't used yet, but another instance writes the token to the db before it does. I think in that case the client can just retry with a different token, but I wonder if its possible to do the check and write in the same query

jarrel-b (approved):
This is so good, excited to see this in action soon!

db/migrations/core/000080_add_push_notifications.up.sql
```diff
@@ -0,0 +1,23 @@
+create table if not exists push_notification_tokens (
+    id varchar(255) primary key,
+    user_id varchar(255) not null references users(id),
+    push_token varchar(255) not null,
```
jarrel-b, line 4: Would we want to store the message body somewhere eventually, or it's not really useful for us?

docker/pushnotifications/Dockerfile
radazen: Ah, thanks! I'll add that.

secrets/prod/makevars.yaml
radazen: That's something I was actually hoping to follow up with you about! Long story short, I'd like to have my Cloud Scheduler job send auth headers to the `pushnotifications` service, and if I'm deploying the job via makefile, I'll actually need to specify the correct auth stuff in a `--headers` flag. Before doing any of that, though, I was hoping to ask a few questions about deploying Cloud Scheduler jobs from our makefile in general. I'll follow up offline!

pushnotifications/expo/handler.go
radazen: Oof, good catch. I think I replaced `return error` with `reportError` instead of adding it on a new line. Thanks!

db/migrations/core/000080_add_push_notifications.up.sql
radazen: We could feasibly store any of the message details in the `push_notification_tickets` table if we wanted to, since there's a 1:1 mapping from tickets to sent messages. I'm not sure if there's any benefit to storing it, though, as opposed to logging it (if we're looking to keep it around for debugging purposes).

publicapi/user.go
radazen: Yeah, this is an interesting one. My original query actually did exactly what you're describing, but it felt like it made the query harder to read and harder to work with, so I opted for a less "correct" approach that felt easier to reason about.

To your original question: it should be exceedingly rare for a token to be registered twice, and it should never be the case that a token is registered by more than one user. While one writer could indeed see that the token doesn't exist, and then try to write it after it's already been written...there's a uniqueness constraint on the `push_token` column that will cause the second writer to error. The worst case scenario should be that, in the unlikely case where a race condition is actually triggered, there's no data corruption and the second writer just gets an unexpected error and has to try again.

Given that the constraint handles the "correctness" of the whole thing, the real issue is just figuring out what to communicate to the user when adding a token doesn't go according to plan. While I think the current approach here is fine for our use case (very very small chance of an unexpected error, no real downside if it happens), I would like a better pattern overall. Something I was testing on my storage branch: when we encounter a constraint violation, pgx actually returns an error with the name of the constraint that was violated, and we can match against that constraint name to reduce this kind of stuff to single step where we try to insert, fail, and check whether the failure was due to a constraint issue. I'm not sure if it's a good approach -- matching against strings feels brittle -- but it is definitely a possibility!

jarrel-b: Makes sense to me! Seems super rare that this will actually happen and not a big deal if we drop a notification

PR #937, opened 2023-05-02
Add optional flags to cloud scheduler deployments

radazen:
This PR adds support for base64-encoded auth and `attempt deadline` flags when deploying cron jobs.

I wasn't sure if there's a more correct way to set optional flags in make recipes. I added a callable function that appends to a variable if a value exists, but if there's something else I should be doing instead, I'm all for it!

### Summary by OpenAI

Release Notes:
- New Feature: Added support for base64-encoded auth and `attempt deadline` flags when deploying cron jobs.
- New Feature: Introduced a helper function to add optional flags to a command and a callable function that appends to a variable if a value exists.

> "Flags and deadlines, oh my!
> Cron jobs now deploy with ease,
> Thanks to this PR's new features."

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,9 @@ DOCKER_CONTEXT       := .
 DOCKER_REGISTRY      := us-east1-docker.pkg.dev
 FFMPEG_VERSION       = 7:4.3.6-0+deb11u1
 
+# Helper function to add optional flags to a command
+OPTIONAL_FLAG = $(if $(1),$(eval $(2) += $(3)))
+
 # Environments
 DEV     := dev
 PROD    := prod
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -290,7 +302,9 @@ _$(DOCKER)-$(DEPLOY)-%:
 	@if [ $$? -eq 0 ] && echo $(DEPLOY_FLAGS) | grep -e "--no-traffic" > /dev/null; then echo "\n\tVERSION: '$(CURRENT_COMMIT_HASH)' was deployed but is not currently receiving traffic.\n\tRun 'make promote-$(ENV)-$* version=$(CURRENT_COMMIT_HASH)' to promote it!\n"; else echo "\n\tVERSION: '$(CURRENT_COMMIT_HASH)' was deployed!\n"; fi
 
 _$(CRON)-$(DEPLOY)-%:
-	@$(SCHEDULER_DEPLOY)
+	$(call OPTIONAL_FLAG, $(CRON_PASSWORD), FLAGS, --headers='Authorization=Basic $(shell printf ":$(CRON_PASSWORD)" | base64)')
+	$(call OPTIONAL_FLAG, $(CRON_DEADLINE), FLAGS, --attempt-deadline=$(CRON_DEADLINE))
```
jarrel-b, line 306: Very cool!  `FLAGS` doesn't need to be defined before?

jarrel-b (approved):
Very neat! I would've done something like this:
```make
COMMAND_TO_RUN  = do something $(REQUIRED_FLAGS) $(OPTIONAL_FLAGS)
OPTIONAL_FLAGS  =

# target-specific vars
deploy-something: OPTIONAL_FLAGS = --foo=bar --bar=foo
```

radazen:
> Very neat! I would've done something like this:
> 
> ```makefile
> COMMAND_TO_RUN  = do something $(REQUIRED_FLAGS) $(OPTIONAL_FLAGS)
> OPTIONAL_FLAGS  =
> 
> # target-specific vars
> deploy-something: OPTIONAL_FLAGS = --foo=bar --bar=foo
> ```

That makes sense! I was thinking target-specific variables wouldn't work with stuff like `$(shell)` (for the base64 encoding) for some reason, but obviously they do -- the `CRON_URI` variable uses `$(shell)` and works fine. I'll try that approach. Thanks!

Makefile
radazen: It doesn't! Pretty convenient, as it turns out.

jarrel-b (approved):
🚀

PR #939, opened 2023-05-04
QR code login

radazen:
## What's new?
- Updated to the latest golang-jwt library
- Refactored JWT generation and parsing
  - Should be clearer and more reusable now
  - Rolled `emails` token handling into the regular `auth/jwt.go` file
  - Renamed a bunch of methods and parameters -- no need to call our auth tokens `jwt` everywhere
- Renamed JWT_SECRET env variables
  - Makes it more obvious what key is used for what
  - These should be functionally equivalent to what they were before, just renamed
- Added `OneTimeLoginToken` auth token type
  - This is used by the new `GenerateQRCodeLoginToken` GraphQL mutation
- Added `OneTimeLoginTokenAuthenticator` to consume `OneTimeLogin` tokens

## What's missing?
- "Consuming" used `OneTimeLogin` tokens via Redis cache. This will be in a follow-up PR (that also includes some Redis refactoring)


service/auth/auth.go
```diff
@@ -23,13 +23,6 @@ import (
 	"github.com/mikeydub/go-gallery/util"
 )
 
-// TODO: ExternalWallet? Something to denote "wallet, but not part of our ecosystem"
```
radazen, line 26: I appreciate that everyone left my brainstorming comment around for a year, but it was time to let it go

service/auth/auth.go
```diff
@@ -265,11 +253,7 @@ func (e MagicLinkAuthenticator) Authenticate(pCtx context.Context) (*AuthResult,
 
 	user, err := e.UserRepo.GetByEmail(pCtx, persist.Email(info.Email))
 	if err != nil {
-		if _, ok := err.(persist.ErrUserNotFound); !ok {
```
radazen, line 268: @benny-conn I changed this to return `persist.ErrUserNotFound` instead of a custom `auth.ErrUserNotFound` type, but I wanted to check with you: is there a reason we wanted to return a different type here instead? For example, are we showing this message directly to clients and using the `auth` variant of the error to produce a more appropriate user-facing message?

service/auth/jwt.go
```diff
@@ -2,88 +2,138 @@ package auth
...
+}
+
+type authClaims struct {
+	UserID persist.DBID `json:"user_id"`
+	GalleryClaims
+}
+
+type oneTimeLoginClaims struct {
 	UserID persist.DBID `json:"user_id"`
-	jwt.StandardClaims
```
radazen, line 16: `jwt.StandardClaims` was replaced by `jwt.RegisteredClaims` in the latest version of golang-jwt

Robinnnnn (approved):
cleannn

PR #941, opened 2023-05-05
Add token v3 feature flag

jarrel-b:
#### Changes
* Adds user roles to context via auth middleware
* Users with admin role now uses v3 tokens
    * This works by adding a resolver specifically for media that checks if the user is an admin
    * If the user is an admin, it will fetch the media from the new v3 media table instead
* Upgrades sqlc to 1.18
    * For context, I was thinking to use the embed feature but realized that sqlc generated new structs which would prevent us from re-using dataloaders so I went another way
* Adds `role` package mainly for fetching roles


db/queries/core/query.sql
```diff
@@ -706,21 +706,20 @@ on conflict (user_id, role) do update set deleted = false, last_updated = now();
 update user_roles set deleted = true, last_updated = now() where user_id = $1 and role = any(@roles);
 
 -- name: GetUserRolesByUserId :many
-select role from user_roles where user_id = $1 and deleted = false
+with membership_roles(role) as (
```
jarrel-b, line 709: Had to refactor this query because the new version of sqlc weirdly couldn't parse it

graphql/schema/schema.graphql
```diff
@@ -366,13 +366,13 @@ enum InteractionType {
   Comment
 }
 
-type Token implements Node {
+type Token implements Node @goEmbedHelper {
```
jarrel-b, line 369: Added this helper so that we can use the fetched fallback media from the db token when resolving media that coming from the v3 table.

middleware/middleware.go
```diff
@@ -130,6 +131,9 @@ func AddAuthToContext() gin.HandlerFunc {
 		userID, err := auth.JWTParse(jwt, env.GetString("JWT_SECRET"))
 		auth.SetAuthStateForCtx(c, userID, err)
 
+		roles, err := auth.RolesByUserID(c, queries, userID)
```
radazen, line 134: This might be something that changes in the near future with refresh tokens. As I understand it, one of the goals behind a two-token system (one auth token, one refresh token) is to eliminate hitting the database to authenticate every user action. In the future, we'll probably bake the `roles` right into the auth token, parse them the same way that we parse the userID, and assume that they're true. And then whenever we use the refresh token to generate a new auth token (which could be like every 5 minutes or something), we hit the database one time to figure out the user's roles for the newly generated auth token.

publicapi/user.go
```diff
@@ -208,27 +206,22 @@ func (api UserAPI) GetUsersWithTrait(ctx context.Context, trait string) ([]db.Us
...
-		GrantedMembershipRole: persist.RoleEarlyAccess, // Role granted if user carries a matching token
-		Chain:                 persist.ChainETH,
-	})
+	gc := util.MustGetGinContext(ctx)
+
+	if auth.GetUserRolesExistFromCtx(gc) {
+		return auth.GetRolesFromCtx(gc), nil
+	}
+
+	return auth.RolesByUserID(ctx, api.queries, userID)
```
radazen, line 215: It seems like this function is doing two different things: if the _currently logged in user_ has any roles, it returns them (regardless of the `userID` supplied). If there are no roles there, then we query and return the roles for the _specified user_.

radazen (approved):
Looks good! I love the `Media` resolver approach to feature flagging the token changes. I left a few comments, but I also have one general question: do we actually set `RoleAdmin` anywhere? I searched the existing code and the changes in this PR, and I can't find where an admin user would actually end up with `persist.RoleAdmin` in their list of roles.

jarrel-b:
> Looks good! I love the `Media` resolver approach to feature flagging the token changes. I left a few comments, but I also have one general question: do we actually set `RoleAdmin` anywhere? I searched the existing code and the changes in this PR, and I can't find where an admin user would actually end up with `persist.RoleAdmin` in their list of roles.

Yea good question! This is set manually atm or via the `addRolesToUser` mutation

publicapi/user.go
jarrel-b: Ohh good call, I was thinking we could save time by checking the context, but makes sense to always refetch the roles.

graphql/resolver/schema.resolvers.helpers.go
```diff
@@ -1860,38 +1861,43 @@ func getUrlExtension(url string) string {
 	return strings.ToLower(strings.TrimPrefix(filepath.Ext(url), "."))
 }
 
-func getMediaForToken(ctx context.Context, token db.Token) model.MediaSubtype {
-	med := token.Media
+func mediaToModel(ctx context.Context, media persist.Media, fallback persist.FallbackMedia) model.MediaSubtype {
+	var fallbackMedia *model.FallbackMedia
 
-	fallbackMedia := getFallbackMedia(ctx, token.FallbackMedia)
+	if !media.IsServable() && fallback.IsServable() {
```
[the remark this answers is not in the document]
jarrel-b: Good catch! This was a bad merge

middleware/middleware.go
radazen: No need to do it until we have refresh tokens in place! Current auth tokens last for 4 weeks, which makes mutable stuff like roles a bad fit. Once we have refresh tokens (should be within the next 2 weeks!), then auth tokens will last for a short duration (5m?) and refresh tokens will be used to re-issue auth tokens frequently.

jarrel-b: ahhh gotcha gotcha makes snese

PR #943, opened 2023-05-08
Redis refactor

radazen:
## What's new?
- Our redis `Cache` abstraction now allows namespacing within a redis database via key prefix
- We always use `Caches` instead of `Clients` to make sure that namespacing is consistent and enforced
- Existing uses of `redis.Client` have been updated to use the `Cache` type instead
  - These include distributed locks via `redislock` and rate limiters
- Fixed a bug where the email service `/verification` and `/preverify` handlers were sharing the same rate limiters
- Refactored rate limiters to make it easier to understand how they work
- OneTimeLogin (e.g. QR code login) now uses redis to consume a token so it can't be used to log in twice

Also of note: I moved our notification locks into a different redis DB and gave them a prefix. We don't necessarily _have_ to do this right now, but it felt like a good time to make a `locks` DB. That being said, there could be some weirdness here, since existing notification locks could have been started in the old DB and will be queried in the new DB after this is deployed. I'm assuming we don't hold locks very long and the risk is negligible, but I'd love input from @benny-conn or @jarrel-b!

### DIFF
--- a/middleware/middleware.go
+++ b/middleware/middleware.go
@@ -142,8 +142,8 @@ func AddAuthToContext() gin.HandlerFunc {
 	}
 }
 
-// RateLimited is a middleware that rate limits requests by IP address
-func RateLimited(lim *KeyRateLimiter) gin.HandlerFunc {
+// IPRateLimited is a middleware that rate limits requests by IP address
+func IPRateLimited(lim *KeyRateLimiter) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		canContinue, tryAgainAfter, err := lim.ForKey(c, c.ClientIP())
 		if err != nil {
--- a/middleware/rate.go
+++ b/middleware/rate.go
@@ -3,56 +3,136 @@ package middleware
 import (
 	"context"
 	"fmt"
-	"log"
+	"github.com/bsm/redislock"
+	"github.com/mikeydub/go-gallery/service/logger"
+	"github.com/mikeydub/go-gallery/service/redis"
+	"github.com/sirupsen/logrus"
 	"time"
 
 	"github.com/benny-conn/limiters"
-	"github.com/go-redis/redis/v8"
-	gredis "github.com/mikeydub/go-gallery/service/redis"
 )
 
-// KeyRateLimiter .
 type KeyRateLimiter struct {
-	rateDuration time.Duration
-	rateAmount   int64
+	cache        *redis.Cache
+	name         string
+	capacity     int64
+	refillRate   time.Duration
+	timeToRefill time.Duration
 	reg          *limiters.Registry
-	red          *redis.Client
 	clock        *limiters.SystemClock
-	logger       *limiters.StdLogger
-	lock         *gredis.GlobalLock
+	logger       limiters.Logger
+	lock         *distributedLock
 }
 
-// NewKeyRateLimiter creates a new rate limiter that will limit to `rateAmount` requests `every` duration
-// uses redis in the backend
-func NewKeyRateLimiter(rateAmount int64, every time.Duration, red *redis.Client) *KeyRateLimiter {
+// NewKeyRateLimiter creates a new rate limiter that will limit to `amount` operations `every` duration.
+// The name will be used to uniquely identify this limiter in the specified redis cache. Consequently,
+// two different limiters in the same cache should NOT have the same name. It is safe to share a single
+// KeyRateLimiter object among multiple consumers to share a limit.
+func NewKeyRateLimiter(ctx context.Context, cache *redis.Cache, name string, amount int64, every time.Duration) *KeyRateLimiter {
+	registry := limiters.NewRegistry()
+	clock := limiters.NewSystemClock()
 
-	i := &KeyRateLimiter{
-		rateDuration: every,
-		rateAmount:   rateAmount,
-		reg:          limiters.NewRegistry(),
-		clock:        limiters.NewSystemClock(),
-		logger:       limiters.NewStdLogger(),
-		red:          red,
-		lock:         gredis.NewGlobalLock(gredis.NewLockClient(red.Options().DB), every*time.Duration(rateAmount)),
+	// Refill rate is per token, so we have to divide to get the correct rate
+	refillRate := time.Duration(float64(every) / float64(amount))
+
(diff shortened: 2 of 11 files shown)


middleware/middleware.go
```diff
@@ -142,8 +142,8 @@ func AddAuthToContext() gin.HandlerFunc {
 	}
 }
 
-// RateLimited is a middleware that rate limits requests by IP address
-func RateLimited(lim *KeyRateLimiter) gin.HandlerFunc {
+// IPRateLimited is a middleware that rate limits requests by IP address
+func IPRateLimited(lim *KeyRateLimiter) gin.HandlerFunc {
```
radazen, line 146: Renamed this to make it clear that it rate limits by IP address, since we could limit by API key etc at some point in the future

middleware/rate.go
```diff
@@ -3,56 +3,136 @@ package middleware
...
-	logger       *limiters.StdLogger
-	lock         *gredis.GlobalLock
-}
-
-// NewKeyRateLimiter creates a new rate limiter that will limit to `rateAmount` requests `every` duration
-// uses redis in the backend
-func NewKeyRateLimiter(rateAmount int64, every time.Duration, red *redis.Client) *KeyRateLimiter {
-
-	i := &KeyRateLimiter{
-		rateDuration: every,
```
radazen, line 30: I believe we were historically getting this wrong: the bucket refills _one_ token every `rateDuration`, but we were treating it as the time required to refill _all_ tokens. It hasn't mattered, though, since our only rate limiters have a bucket capacity of 1 (which means refilling one token and refilling all tokens are the same thing).

service/redis/redis.go
```diff
@@ -17,87 +17,105 @@ type ErrKeyNotFound struct {
...
 
-// NewCache creates a new redis cache
-func NewCache(db int) *Cache {
-	return &Cache{client: NewClient(db)}
+func (c *Cache) Client() *redis.Client {
+	return c.client
 }
 
-// ClearCache deletes the entire cache
-func ClearCache(db int) error {
```
radazen, line 86: This is trickier to implement with prefixed caches (since there's no easy way to flush only the keys in a database that start with a particular prefix), but we also weren't using it anywhere, so I got rid of it. We can add a prefix-compatible version in the future if necessary.

service/redis/redis.go
```diff
@@ -109,49 +127,63 @@ func (c *Cache) Get(pCtx context.Context, key string) ([]byte, error) {
...
+		return key
+	}
+
+	return c.keyPrefix + ":" + key
 }
 
-// GlobalLock is a distributed lock that does not depend on a key.
-// It can store its own context and lock because we can assume that there never be more than one global lock in use at a time
-// and therefore no concurrency issues with overwriting the context and lock.
-type GlobalLock struct {
```
radazen, line 132: Our uses of `GlobalLock` were very specific to rate limiters, so I refactored this into a private implementation used by rate limiters.

service/redis/redis.go
```diff
@@ -17,87 +17,105 @@ type ErrKeyNotFound struct {
...
-	case FeedDB:
-		return "FeedDB"
-	}
-
-	return fmt.Sprintf("db %d", databaseId)
-}
+// Every cache is uniquely defined by its database and key prefix. Display names are used for tracing.
+
+var (
+	NotificationLockCache        = CacheConfig{database: locks, keyPrefix: "notif", displayName: "notificationLock"}
```
jarrel-b, line 45: I may be misunderstanding, but wouldn't each config create prefixed keys in separate databases?  Since the Redis client is initted to use the DB provided by `database`?

middleware/rate.go
```diff
@@ -3,56 +3,136 @@ package middleware
...
+// NewKeyRateLimiter creates a new rate limiter that will limit to `amount` operations `every` duration.
+// The name will be used to uniquely identify this limiter in the specified redis cache. Consequently,
+// two different limiters in the same cache should NOT have the same name. It is safe to share a single
+// KeyRateLimiter object among multiple consumers to share a limit.
+func NewKeyRateLimiter(ctx context.Context, cache *redis.Cache, name string, amount int64, every time.Duration) *KeyRateLimiter {
+	registry := limiters.NewRegistry()
+	clock := limiters.NewSystemClock()
+
+	// Refill rate is per token, so we have to divide to get the correct rate
+	refillRate := time.Duration(float64(every) / float64(amount))
```
jarrel-b, line 36: To check my understanding, does this mean that one token is replenished each `refillRate` so with 10 * refillRate we expect 10 slots to be free?

jarrel-b (approved):
Looks awesome!

jarrel-b:
> I'm assuming we don't hold locks very long and the risk is negligible
Agreed! I think this is pretty low risk and probably won't happen with our traffic

service/redis/redis.go
radazen: Totally, yeah. For the most part I'm leaving things as they are (separate database, no prefix) for the sake of backwards compatibility and not breaking a bunch of stuff. The only thing I'd likely be breaking is the notification lock, since I'm moving it to a new database and giving it a prefix.

middleware/rate.go
radazen: Yeah, exactly. `amount` is actually what the token bucket algorithm calls `capacity`, and it refills at a rate of one token per `refillRate`, up to a maximum of `amount` tokens.

radazen: (also, I kept our existing pattern of `amount` requests `every` duration and mapped it to the `capacity` and `refillRate` concepts in the token bucket algorithm, but I'm tempted to use the correct terms and have us specify `capacity` and `refillRate` since they're more accurate and might give callers a better sense of what's going to happen)

service/redis/redis.go
jarrel-b: Ah makes sense!

middleware/rate.go
jarrel-b: makes sense to me! rate makes me think that the units should be `token / time` rather than `time / token` but if that's the standard terms then all for it

radazen: I legitimately have no idea if those are the standard terms for what we're describing, they're just the terms used by the package we're importing. 😅

PR #945, opened 2023-05-09
Sort interactions by comments, then admires

radazen:
The `Interactions` resolver has historically returned a list of intermingled `Comments` and `Admires`, sorted by time. This PR updates the resolver to return `Comments` first, and then `Admires` (each still sorted by time within its own category).

Doing this requires encoding the interaction type in the cursor, so I added an `intTimeIDPaginator` type to handle these cursors. Existing cursors will be invalid once this is deployed, but reloading a browser should fix any issues.

I also removed the `typeFilter` parameter, since we weren't using it and it doesn't make as much sense with this sort order.


jarrel-b (approved):
🚢 🚢 🚢

PR #946, opened 2023-05-10
Send push notifications

radazen:
## What's new?
- Admires, comments, and follows now trigger push notifications (if the user has any devices registered to receive push notifications)
- The `push_notification_tickets` table now has a status column so we know whether a notification was ultimately delivered or not
- The `logout` mutation takes an optional `pushTokenToUnregister` parameter to allow logging out and disabling push notifications in a single mutation

### DIFF
--- a/db/migrations/core/000082_add_push_ticket_status.up.sql
+++ b/db/migrations/core/000082_add_push_ticket_status.up.sql
@@ -1,0 +1,3 @@
+alter table push_notification_tickets add status text;
+update push_notification_tickets set status = 'unknown' where status is null;
+alter table push_notification_tickets alter column status set not null;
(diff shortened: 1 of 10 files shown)


db/migrations/core/000082_add_push_ticket_status.up.sql
```diff
@@ -0,0 +1,3 @@
+alter table push_notification_tickets add status text;
+update push_notification_tickets set status = 'unknown' where status is null;
```
radazen, line 2: This part probably isn't necessary -- I don't think we actually have anything in the `push_notifications_table` on dev or prod. Still, better safe than sorry!

jarrel-b (approved):
Looks awesome sir, excited to see this live!

PR #947, opened 2023-05-10
Add sentry pipeline tracing

jarrel-b:
This PR adds performance tracing to the pipeline: [example trace](https://usegallery.sentry.io/performance/tokenprocessing:a5bfc7adf1a74c4cb73fe95f99fb5a8a/?environment=local&project=6771759&query=http.method%3APOST&referrer=performance-transaction-summary&showTransactions=recent&statsPeriod=24h&transaction=POST+%2Fmedia%2Fprocess%2Ftoken&unselectedSeries=p100%28%29)

### DIFF
--- a/tokenprocessing/handler.go
+++ b/tokenprocessing/handler.go
@@ -1,15 +1,23 @@
 package tokenprocessing
 
 import (
+	"time"
+
 	"github.com/gin-gonic/gin"
 	"github.com/mikeydub/go-gallery/service/multichain"
 	"github.com/mikeydub/go-gallery/service/persist/postgres"
+	"github.com/mikeydub/go-gallery/service/sentry"
 	"github.com/mikeydub/go-gallery/service/throttle"
 )
 
 func handlersInitServer(router *gin.Engine, tp *tokenProcessor, mc *multichain.Provider, repos *postgres.Repositories, throttler *throttle.Locker) *gin.Engine {
 	mediaGroup := router.Group("/media")
-	mediaGroup.POST("/process", processMediaForUsersTokens(tp, repos.TokenRepository, repos.ContractRepository, throttler))
+	mediaGroup.POST("/process", func(c *gin.Context) {
+		if hub := sentryutil.SentryHubFromContext(c); hub != nil {
+			hub.Scope().AddEventProcessor(sentryutil.SpanFilterEventProcessor(c, 1000, 1*time.Millisecond, 8, true))
+		}
+		processMediaForUsersTokens(tp, repos.TokenRepository, repos.ContractRepository, throttler)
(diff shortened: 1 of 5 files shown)


tokenprocessing/handler.go
```diff
@@ -1,15 +1,23 @@
...
 )
 
 func handlersInitServer(router *gin.Engine, tp *tokenProcessor, mc *multichain.Provider, repos *postgres.Repositories, throttler *throttle.Locker) *gin.Engine {
 	mediaGroup := router.Group("/media")
-	mediaGroup.POST("/process", processMediaForUsersTokens(tp, repos.TokenRepository, repos.ContractRepository, throttler))
+	mediaGroup.POST("/process", func(c *gin.Context) {
+		if hub := sentryutil.SentryHubFromContext(c); hub != nil {
+			hub.Scope().AddEventProcessor(sentryutil.SpanFilterEventProcessor(c, 1000, 1*time.Millisecond, 8, true))
+		}
+		processMediaForUsersTokens(tp, repos.TokenRepository, repos.ContractRepository, throttler)
```
benny-conn, line 19: Does this work as expected? `processMediaForUsersTokens` returns a func that takes in a gin.Context so maybe we have to do `processMediaForUsersTokens(tp, repos.TokenRepository, repos.ContractRepository, throttler)(c)` ?

tokenprocessing/media.go
```diff
@@ -202,7 +202,8 @@ func cacheObjectsForMetadata(pCtx context.Context, metadata persist.TokenMetadat
 
 	// neither download worked, unexpectedly
 	if (animCh == nil || (animResult.err != nil && len(animResult.cachedObjects) == 0)) && (imgCh == nil || (imgResult.err != nil && len(imgResult.cachedObjects) == 0)) {
-		defer persist.TrackStepStatus(pCtx, &pMeta.NothingCachedWithErrors, "NothingCachedWithErrors")()
+		persist.TrackStepStatus(pCtx, &pMeta.NothingCachedWithErrors, "NothingCachedWithErrors")
```
benny-conn, line 205: Is there a reason why this one and the next one don't do `traceCallback, ctx :=` and `defer traceCallback()`?

benny-conn (approved):
Looks great, just a few small comments!

tokenprocessing/handler.go
jarrel-b: Ah good catch! You're right. I originally had this factored out but decided to do it inline

tokenprocessing/media.go
jarrel-b: Whoops, you're right this should call `traceCallback()`.

PR #948, opened 2023-05-11
Merch store fix

jarrel-b:
**Changes**
* Adds check to catch when the user owns no merch tokens
* Replaces the underlying API endpoint of the Alchemy provider's `GetTokensByContractAddressAndOwner` method from /getNFTsForCollection to /getNFTs. The former requires fetching the entire collection and then fetching the owner of each token to see if it matches the user's address which was causing slow load times
* Adds `RPC_URL` to local env

### DIFF
--- a/service/persist/postgres/user.go
+++ b/service/persist/postgres/user.go
@@ -375,6 +375,18 @@ func (u *UserRepository) GetByWalletID(pCtx context.Context, pWalletID persist.D
 		return persist.User{}, err
 	}
 
+	wallets := make([]persist.Wallet, len(user.Wallets))
+
+	for i, wallet := range user.Wallets {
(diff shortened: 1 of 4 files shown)


service/persist/postgres/user.go
```diff
@@ -375,6 +375,18 @@ func (u *UserRepository) GetByWalletID(pCtx context.Context, pWalletID persist.D
 		return persist.User{}, err
 	}
 
+	wallets := make([]persist.Wallet, len(user.Wallets))
+
+	for i, wallet := range user.Wallets {
```
jarrel-b, line 380: Copied from the other user repo get methods. We need to hydrate the user object with their wallet addresses since they get used in the token dedupe logic.

kaitoo1 (approved):

Robinnnnn (approved):

PR #954, opened 2023-05-12
Add metrics to token pipeline

jarrel-b:
#### Changes
* Adds `metric` package. The metrics reporter used should be easy to switch out if/when we want to record metrics in some other way like custom metrics
* Adds log-based metrics to the pipeline, to start it will track:
    * `pipeline_completed`: pipeline completed
    * `pipeline_duration`: how long it took to run the pipeline
    * `pipeline_failed`: error in executing the pipeline
    * `pipeline_timedout`: pipeline either timed out or was canceled
* Metrics are tagged by media type and chain so that we can plot different time series

### DIFF
--- a/service/metric/metric.go
+++ b/service/metric/metric.go
@@ -18,7 +18,7 @@ type MetricReporter struct {
 
 var LogOptions = LogOptionBulder{}
 
-func NewGCPLogMetricReporter() MetricReporter {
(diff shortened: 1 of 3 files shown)


Robinnnnn (commented):
thanks so much for the quick research + implementation!

Robinnnnn (approved):

PR #963, opened 2023-05-16
Rate limit push notifications

radazen:
This PR uses our existing Redis-based rate limiters to limit the number of push notifications a user can receive. The current limits are:

1 admire per `sender:receiver:feedEvent` every 10 minutes
1 follow per `sender:receiver` every 10 minutes
5 comments per `sender:receiver:feedEvent` every minute

In practical terms, this means you'll get notifications every time someone admires one of your feed events, but you won't get a notification if they un-admire and re-admire the same event (unless 10 minutes has passed since the last time they admired it). Same with follows: you won't get spammed by someone unfollowing and reflowing you, though you could get notified once every 10 minutes.

Comments are more spammable: the limit is 5 per feed event every minute, and in practice, I find it difficult to even write comments quickly enough to surpass this limit. My goal here was to make it permissive, since we're not really trying to limit legitimate activity, but we may want to tweak this. Our notifications currently say "ezra commented on your activity" and don't give any additional information, which means they can start to feel redundant when you get 5 of them in a row and they all say the same thing.

### DIFF
--- a/emails/handlers.go
+++ b/emails/handlers.go
@@ -2,6 +2,7 @@ package emails
 
 import (
 	"context"
+	"github.com/mikeydub/go-gallery/service/limiters"
 	"github.com/mikeydub/go-gallery/service/redis"
 	"time"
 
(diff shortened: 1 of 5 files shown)


Robinnnnn (approved):
😍 🔥

PR #967, opened 2023-05-18
Add dashboard indexes and tables

jarrel-b:
Changes
* Unschedules the migration view because we're done with the migration 🚀 
* Adds indexes to `token_medias`
* Adds validation table for the dashboard

### DIFF
--- a/db/migrations/core/000086_unschedule_migration_refresh.up.sql
+++ b/db/migrations/core/000086_unschedule_migration_refresh.up.sql
@@ -1,0 +1,1 @@
+select cron.unschedule((select jobid from cron.job where jobname = 'tokenprocessing-migration-dashboard'));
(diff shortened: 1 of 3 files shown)


db/migrations/core/000086_unschedule_migration_refresh.up.sql
```diff
@@ -0,0 +1 @@
+select cron.unschedule((select jobid from cron.job where jobname = 'tokenprocessing-migration-dashboard'));
```
radazen, line 1: FYI, I think you can unschedule via job name directly, i.e.

`select cron.unschedule('tokenprocessing-migration-dashboard');`

https://github.com/citusdata/pg_cron#what-is-pg_cron

radazen (approved):
👍👍

db/migrations/core/000086_unschedule_migration_refresh.up.sql
jarrel-b: ah good to know!

jarrel-b: Looks like the first method is idempotent if `tokenprocessing-migration-dashboard` was already unscheduled or doesn't exist.

PR #968, opened 2023-05-18
minor fixes

benny-conn:
Changes:

- **Changed** poap API urls to new correct ones
- **Changed** ENS to use opensea image
- **Changed** media pre-determine to not panic
- **Changed** raw media determination to use live renders if available and svgs for thumbnails
- **Changed** IPFS response to only fallback to the ipfs.io api in the worst case
- **Added** support for utf8 svgs


jarrel-b (approved):
🚢

PR #970, opened 2023-05-18
use opensea at further up errors as well

benny-conn:
Changes:

- **Added** opensea token processing fallback step to more top level failure points in the caching of objects


jarrel-b (approved):
Looks great!

PR #972, opened 2023-05-19
new context just for persist

benny-conn:
Changes:

- **Changed** persist token processing step to use its own context just in case the rest of the job failed with a timeout

### DIFF
--- a/tokenprocessing/pipeline.go
+++ b/tokenprocessing/pipeline.go
@@ -3,6 +3,7 @@ package tokenprocessing
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	"cloud.google.com/go/storage"
@@ -57,6 +58,8 @@ type tokenProcessingJob struct {
 
 	cause            persist.ProcessingCause
 	pipelineMetadata *persist.PipelineMetadata
+
+	lock *sync.Mutex
 }
 
 func (tp *tokenProcessor) ProcessTokenPipeline(c context.Context, t persist.TokenGallery, contract persist.ContractGallery, ownerAddress persist.Address, cause persist.ProcessingCause) error {
@@ -99,6 +102,8 @@ func (tp *tokenProcessor) ProcessTokenPipeline(c context.Context, t persist.Toke
 		recordPipelineEndState(ctx, tp.mr, &runResult, time.Since(totalTime))
 		return runResult.Err
 	case <-ctx.Done():
+		job.lock.Lock()
+		defer job.lock.Unlock()
 		recordPipelineEndState(ctx, tp.mr, nil, time.Since(totalTime))
 		return ctx.Err()
 	}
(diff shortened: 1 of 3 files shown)


tokenprocessing/pipeline.go
```diff
@@ -99,6 +102,8 @@ func (tp *tokenProcessor) ProcessTokenPipeline(c context.Context, t persist.Toke
 		recordPipelineEndState(ctx, tp.mr, &runResult, time.Since(totalTime))
 		return runResult.Err
 	case <-ctx.Done():
+		job.lock.Lock()
+		defer job.lock.Unlock()
```
benny-conn, line 106: This ensures that the job finishes inserting the results even if the context times out

tokenprocessing/pipeline.go
```diff
@@ -315,7 +325,11 @@ func (tpj *tokenProcessingJob) upsertDB(ctx context.Context, tmetadata coredb.To
 		HasDescription:  tmetadata.Description != "",
 	}
 
-	job, err := tpj.tp.queries.InsertJob(ctx, coredb.InsertJobParams{
+	// new context just for insert so that if the rest of the job failed because of a context cancellation, we still insert
+	insertContext, cancel := context.WithTimeout(ctx, 2*time.Minute)
```
jarrel-b, line 329: If `ctx` is already canceled would `insertContext` already be canceled, or `insertContext` has its own timeout independent of `ctx`?

benny-conn: Woops! You are right

benny-conn: Copilot did not get what I was going for here 🤦‍♂️

tokenprocessing/pipeline.go
```diff
@@ -112,9 +117,14 @@ type runResult struct {
 }
 
 func (tpj *tokenProcessingJob) run(ctx context.Context) runResult {
+	tpj.lock.Lock()
```
jarrel-b, line 120: Would we want to move this to `persistResults`? Correct me if I'm wrong, but if `ctx` times out and we hit that `ctx.Done()` select branch, could it get stuck waiting for media that may never come?

tokenprocessing/pipeline.go
```diff
@@ -115,12 +104,37 @@ func (tpj *tokenProcessingJob) run(ctx context.Context) runResult {
...
 	}
 
-	err = tpj.persistResults(ctx, media)
+	persistCtx, cancel := context.WithTimeout(ctx, time.Minute*10)
+	defer cancel()
+
+	persistErr := tpj.persistResults(persistCtx, media)
+
+	var err error
+	if persistErr != nil || mediaErr != nil {
```
jarrel-b, line 135: Do you think we can wrap the errors only when both errors are not `nil` just so we can get the actual error if only of them isn't null?

jarrel-b (approved):

PR #975, opened 2023-05-22
use chains in lock

benny-conn:
Changes:

- **Added** chains to throttle lock on token processor

This was previously not causing any obvious issues in v2 because tokens would just end up not getting reprocessed, it is a lot more obvious in v3 when a new token would not receive a token media ID.

What would happen is a user would sync tokens of one chain, then do another right after, and because we throttled on user ID, all of the second chain's tokens would get throttled and not get processed by the token processor.


jarrel-b (approved):
🚢

PR #976, opened 2023-05-22
Insert pipeline results in single query

jarrel-b:


tokenprocessing/pipeline.go
```diff
@@ -308,9 +308,8 @@ func (tpj *tokenProcessingJob) isNewMediaPreferable(ctx context.Context, media p
 	traceCallback, ctx := persist.TrackStepStatus(ctx, &tpj.pipelineMetadata.MediaResultComparison, "MediaResultComparison")
 	defer traceCallback()
 
-	if media.IsServable() || (!media.IsServable() && (tpj.token.TokenMediaID == "" || !tpj.token.TokenMedia.IsServable())) {
+	if media.IsServable() {
```
jarrel-b, line 311: I removed the check for v2 media and moved the check if a token media id exists to the insert query.

PR #981, opened 2023-05-24
Benny/token lock

benny-conn:
Changes:

- **Added** lock on individual tokens in sync


tokenprocessing/process.go
```diff
@@ -80,6 +73,8 @@ func processMediaForUsersTokens(tp *tokenProcessor, tokenRepo *postgres.TokenGal
 			}
 
 			wp.Go(func() error {
+				defer throttler.Unlock(reqCtx, lockID)
```
jarrel-b, line 76: I was thinking that the lock gets acquired here instead, that way whatever goroutine is handling this request doesn't hold a a token lock until it starts working on it.

jarrel-b (approved):

PR #986, opened 2023-05-25
Fix invalid type assertion

radazen:
Caught this error via Sentry. `UpdateFeedEventCaptionByGroup` returns a `FeedEvent`, not a `*FeedEvent`, which causes a type assertion error later on when we try to assert `*FeedEvent`.

This PR fixes the line in question to return `*FeedEvent`. It seems we always expect `handleImmediate` and `handleGroup` to return `*FeedEvent` types, so I also changed the signature of feed handler methods to use `*FeedEvent` instead of `interface{}` so the compiler will catch similar errors in the future.

### DIFF
--- a/event/event.go
+++ b/event/event.go
@@ -287,11 +287,11 @@ type delayedHandler interface {
 }
 
 type immediateHandler interface {
-	handleImmediate(context.Context, db.Event) (interface{}, error)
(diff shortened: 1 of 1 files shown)


Robinnnnn (approved):

jarrel-b (approved):

kaitoo1:
awesome use of sentry 💯

PR #991, opened 2023-05-30
Fix token name

jarrel-b:
**Changes**
* Fixes checks for filling out token descriptors
* The order of providers passed to multichain matter: the fallback eth provider was first meaning that its results are prioritized. The fallback provider was returning an inaccurate token name, and wasn't being replaced with later providers data

In the future, we can make the priority of providers more explicit


radazen (approved):
🎉🎉🎉

PR #994, opened 2023-06-01
Add wire to multichain and remove deep refresh

jarrel-b:
This PR introduces wire https://github.com/google/wire which allows us to have compile-time validation of our multichain configuration. 

You can install wire by running:
```
go install github.com/google/wire/cmd/wire@latest
```
which will add the `wire` to your `GOPATH`.  Running `wire` in a dir with build tags:
```
// +build !wireinject
```

will generate injectors for constructing structs. You can also use `go generate ./...` to generate injectors. Comments are left in `inject.go` which should be self explanatory, but lmk if more clarification is needed!

This PR also removes the deep refresh stuff everywhere since its no longer used.


server/inject.go
```diff
@@ -0,0 +1,330 @@
...
+	"github.com/mikeydub/go-gallery/service/task"
+	"github.com/mikeydub/go-gallery/util"
+)
+
+// envInit is a type returned after setting up the environment
+// Adding envInit as a dependency to a provider will ensure that the environment is set up prior
+// to calling the provider
+type envInit struct{}
+
+type ethProviderList []any
```
jarrel-b, line 37: If a class requires inputs of the same type, you'll need to define separate types so that wire nows which arg to use in initialization.

server/inject.go
```diff
@@ -0,0 +1,330 @@
...
+		ethFallbackProvider,
+		opensea.NewProvider,
+	)
+	return ethProviderList{}
+}
+
+// ethProvidersConfig is a wire injector that binds multichain interfaces to their concrete Ethereum implementations
+func ethProvidersConfig(indexerProvider *eth.Provider, openseaProvider *opensea.Provider, fallbackProvider multichain.SyncFailureFallbackProvider) ethProviderList {
+	wire.Build(
+		wire.Bind(new(multichain.NameResolver), util.ToPointer(indexerProvider)),
```
jarrel-b, line 108: Here is where you bind an interface to a concrete implementation. You can't bind two types to the same interface since wire won't know which one to use. For now, each provider maps to at least one distinct interface, but if we have a new provider that only implements `NameResolver` for example, we'd want to add an additional interface e.g. `NameResolverSecondary`, and bind the resolver to that interface.

server/inject.go
```diff
@@ -0,0 +1,354 @@
...
+func ethRequirements(
+	nr multichain.NameResolver,
+	v multichain.Verifier,
+	tof multichain.TokensOwnerFetcher,
+	toc multichain.TokensContractFetcher,
+	cr multichain.ContractRefresher,
+	tmf multichain.TokenMetadataFetcher,
+	tdf multichain.TokenDescriptorsFetcher,
+) ethProviderList {
+	return ethProviderList{nr, v, tof, toc, cr, tmf, tdf}
```
jarrel-b, line 134: I originally wanted to avoid having to manually build the provider list by defining a type like:
```go
type ethProviders struct {
  NameResolvers []multichain.NameResolvers
  Verifiers     []multichain.Verifiers
  // other interfaces ...
}
```
and let wire handle the construction by doing something like:
```go
func newEthProviders() ethProviders {
    wire.Build(
        // eth dependencies...
       wire.Struct(new(ethProviders), "*")
    )
   return ethProviders{}
}
```
but the way we access providers by chain and interface in `multichain` would require a larger refactor which I think would require using reflection so I'm leaving this alone for now

radazen (approved):
This is great. After I read through the `inject.go` file, I was like "hmm, seems like a lot of setup, is this really worthwhile?" And then I got to all the stuff you removed -- especially the runtime provider validation -- and it became immediately obvious that this is a real win. Thanks for the helpful comments, too. No notes!

PR #1000, opened 2023-06-06
use ID

benny-conn:
Changes:

- **Added** ID to statistics

### DIFF
--- a/db/gen/indexerdb/query.sql.go
+++ b/db/gen/indexerdb/query.sql.go
@@ -100,16 +100,17 @@ func (q *Queries) GetReprocessJobRangeByID(ctx context.Context, id int) (Reproce
 }
 
 const insertStatistic = `-- name: InsertStatistic :one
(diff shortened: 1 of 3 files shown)


tokenprocessing/pipeline.go
```diff
@@ -250,10 +250,10 @@ func (tpj *tokenProcessingJob) cacheMediaObjects(ctx context.Context, metadata p
 	)
 
 	if animURL != "" {
-		animCh = cacheAnimationObjects(ctx, imgURL, metadata, tpj)
+		animCh = cacheAnimationObjects(ctx, animURL, metadata, tpj)
 	}
 	if imgURL != "" {
-		imgCh = cacheImageObjects(ctx, animURL, metadata, tpj)
+		imgCh = cacheImageObjects(ctx, imgURL, metadata, tpj)
```
jarrel-b, line 256: another very good catch!!

PR #1005, opened 2023-06-07
Pipeline fixes part 4

jarrel-b:
**Changes**
* Adds separate types for animationURL and imageURL to prevent using the wrong one
* Better errors from exec commands
* More logging around failed copies
* Removes unused thumbnail func


jarrel-b:
Not seeing my last commit in this PR for some reason, closing and remaking

PR #1016, opened 2023-06-08
GAL-2875 Add top collection query

jarrel-b:
Adds the `topCollectionsForCommunity` query which ranks collections based on the number of tokens displayed in the collection. The ranking is cached in redis, and remains valid for two hours at which point a new ranking is calculated on the next request. 

```graphql
topCollectionsForCommunity(
  input: topCollectionsForCommunityInput!
  before: String
  after: String
  first: Int
  last: Int
): CollectionsConnection

type CollectionEdge {
  node: Collection
  cursor: String
}

type CollectionsConnection {
  edges: [CollectionEdge]
  pageInfo: PageInfo!
}
```

Example query:
```graphql
query {
	topCollectionsForCommunity(
		input: {chainAddress: {
			address: "0x4e1f41613c9084fdb9e34e11fae9412427480e56"
			chain: Ethereum
		}}
		first: 10
	) {
		edges {
			node {
				id
				name
			}
		}
		pageInfo {
			total
			startCursor
			endCursor
			hasNextPage
			hasPreviousPage
			size
		}
	}
}
```


publicapi/collection.go
```diff
@@ -99,6 +103,85 @@ func (api CollectionAPI) GetCollectionsByGalleryId(ctx context.Context, galleryI
...
+		}
+	} else if after != nil {
+		if _, collectionIDs, err = paginator.decodeCursor(*after); err != nil {
+			return nil, pageInfo, err
+		}
+	} else {
+		// No cursor provided, need to access the cache
+		r := ranker{
+			Cache: redis.NewCache(redis.SearchCache),
+			Key:   fmt.Sprintf("top_collections_by_address:%d:%s", chainAddress.Chain(), chainAddress.Address()),
```
Robinnnnn, line 133: not sure if this is customary but would be cool if we consolidated all the cache keys in a single file, so we know at a high level what kind of stuff we cache

Robinnnnn (approved):

kaitoo1 (approved):

PR #1018, opened 2023-06-09
pfps start

benny-conn:
#### Changes
Adds initial support for PFPs by allowing a user to set a PFP with a token that they own. There's also a few changes related to ENS that Benny started that I'll start work on after this PR is merged.

##### Schema Changes
* Adds `profileImage` to `GalleryUser`
* Adds mutations for setting and removing a PFP: `setProfileImage` and `removeProfileImage`

##### DB Changes
* Adds a table `profile_images` for handling PFPs from various sources
* Adds column `profile_image_id` to `users`


publicapi/user.go
```diff
@@ -1321,3 +1328,70 @@ func (api UserAPI) DeletePushTokenByPushToken(ctx context.Context, pushToken str
...
+		return db.ProfileImage{}, err
+	}
+
+	user, err := For(ctx).User.GetUserById(ctx, userID)
+	if err != nil {
+		return db.ProfileImage{}, err
+	}
+
+	if user.ProfileImageID == "" {
+		return db.ProfileImage{}, ErrProfileImageNotSet
```
jarrel-b, line 1393: I'm not sure if it's useful to have an error to denote the pfp isn't set vs. not being able to fetch it. Probably the frontend wouldn't handle the two errors any differently and it would save us a query in the db to fetch.

publicapi/user.go
```diff
@@ -1321,3 +1327,67 @@ func (api UserAPI) DeletePushTokenByPushToken(ctx context.Context, pushToken str
...
+		return err
+	}
+
+	if util.OnlyNils(tokenID) {
+		return ErrProfileImageNoSources
+	}
+
+	if util.ManyNotNils(tokenID) {
+		return ErrProfileImageTooManySources
+	}
```
benny-conn, line 1345: Is it currently not possible for this error to be returned? Just for my own understanding, how could a profile image have too many sources?

benny-conn (commented):
Lovely! I can't approve because I created the PR but feel free to merge because I approve!

publicapi/user.go
jarrel-b: Yup, that's right! I was thinking in the future the call would look something like:
```
if util.ManyNotNils(tokenID, ens, image...) {
   ...
}
```

PR #1024, opened 2023-06-16
Adds timestamp query param to preview urls

jarrel-b:
#### Changes
* Add `ts` query param to preview URLs for assets that we host
* Adds `GCLOUD_TOKEN_CONTENT_BUCKET` to the backend env since it gets used to determine if we host an asset 
* Refactors a few functions to accept `db.TokenMedia` instead of `persist.Media` so that the last updated time can get used

### DIFF
--- a/db/gen/coredb/query.sql.go
+++ b/db/gen/coredb/query.sql.go
@@ -1904,7 +1904,7 @@ func (q *Queries) GetGalleryIDByCollectionID(ctx context.Context, id persist.DBI
 }
 
 const getGalleryTokenMediasByGalleryID = `-- name: GetGalleryTokenMediasByGalleryID :many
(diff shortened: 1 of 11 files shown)


Robinnnnn (approved):

radazen (approved):
Wow, that was quick! Looks awesome! Two thoughts:

- Can we give the `ts` parameter a more Gallery-specific name? Given how short it is, I feel like there's a (very small) chance imgix will actually add their own `ts` parameter at some point, and all of our previews will break cause they're passing an unrecognized value.

- I was paranoid about passing made-up parameters to arbitrary third-party hosts earlier, but if the parameters are always being passed to imgix, is there a downside to attaching them to all previews, even if we aren't hosting their media? On one hand, it seems unlikely it'd do anything: if the image is hosted with a third party, updated metadata will probably point to:
  1. a new image URL, or
  2. the same image URL with the same image

But on the other hand, if someone's trying to refresh the metadata, maybe we should bust the cache just in case?

jarrel-b:
> * I was paranoid about passing made-up parameters to arbitrary third-party hosts earlier, but if the parameters are always being passed to imgix, is there a downside to attaching them to all previews, even if we aren't hosting their media? On one hand, it seems unlikely it'd do anything: if the image is hosted with a third party, updated metadata will probably point to:
>   
>   1. a new image URL, or
>   2. the same image URL with the same image
> 
> But on the other hand, if someone's trying to refresh the metadata, maybe we should bust the cache just in case?

That makes sense! I was thinking that the made-up parameters get forwarded to the hosting server which could reject the request since it doesn't recognize the params, but I was confusing imgix params with the original source params which is the portion that gets signed and URL encoded.

PR #1027, opened 2023-06-20
fetch contracts by owner

benny-conn:
Changes:

- **Added** multichain interfaces and implementations for fetching all contracts owned by a given owner address from the indexer server


radazen (approved):
Looks good to me! Thanks for getting this together so quickly!

PR #1033, opened 2023-06-21
combine sync for parent/child with just parents

benny-conn:
Changes:

- **Changed** get contracts owned by address to a sync that first checks for providers that return contracts owned by an address (just indexer), then uses the already in place sync for getting contracts with parent/child relationships (opensea)

Considerations:

- Right now the function for getting parent/child relationships is called `SyncTokensCreatedOnSharedContracts` but it currently does not do any syncing for tokens. Will it in the future?
- I opted for not implementing a `GetContractsOwnedByAddress` for the `*opensea.Provider` because that functionality is already encompassed in `GetChildContractsCreatedOnSharedContract`. I feel like it makes sense to do this to separate which providers only know the creator of a contract and which providers are capable of determining children

### DIFF
--- a/service/multichain/multichain.go
+++ b/service/multichain/multichain.go
@@ -983,51 +983,65 @@ outer:
 	return p.processTokensForOwnersOfContract(ctx, contract.ID, ci.Chain, users, chainTokensForUsers, persistedContracts)
 }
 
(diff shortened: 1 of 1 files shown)


radazen:
🙌 going to merge this and use it in my PR. Thanks Benny!

PR #1035, opened 2023-06-22
Creator token syncing

radazen:
## What's new?
- `SyncCreatorTokens` will:
  - Query providers to see what contracts a user owns
  - Query the database for all of the user's contracts
    - Because ownership can be manually overridden, this set may be different from what providers return
  - Query providers for all tokens in the contracts owned by this user
  - Bulk upsert all the resulting creator tokens

## Implementation notes
- The `BulkUpsert` SQL statement needed to be slightly adjusted to account for creator tokens. When upserting creator tokens, we don't want to add/update any values for `owned_by_wallets` or `ownership_history`, since those are specific to tokens that are being held.
- The bulk upsert deletion step needed to be modified to keep certain types of tokens. If we're doing a creator token upsert, the final deletion step _should not delete holder tokens_. Similarly, a regular non-creator sync _should not delete creator tokens_.

## Future improvements
- In this PR, we query all created tokens and then upsert them at once, using the same strategy we apply to held tokens. I think a better way would be to scope by contract: instead of waiting until we receive _all created tokens_ before upserting, we can upsert on a per-contract basis. We can upsert all tokens from a particular contract and then delete any tokens from that contract prior to the `last_synced` timestamp, with a final deletion step that removes creator tokens that don't belong to the user's currently owned set of contracts.


service/multichain/multichain.go
```diff
@@ -344,7 +344,119 @@ outer:
 		return err
 	}
 
-	_, err = p.ReplaceTokensForUser(ctx, tokensFromProviders, persistedContracts, user, chains)
+	_, err = p.ReplaceHolderTokensForUser(ctx, tokensFromProviders, persistedContracts, user, chains)
```
radazen, line 347: I've started calling them `Creator` tokens and `Holder` tokens, since those are the two ways someone can own a token

service/multichain/multichain.go
```diff
@@ -1075,7 +1204,8 @@ func (p *Provider) SyncContractsOwnedByUser(ctx context.Context, userID persist.
 		return err
 	}
 
-	return p.SyncTokensCreatedOnSharedContracts(ctx, userID, chains)
+	return nil
+	//return p.SyncTokensCreatedOnSharedContracts(ctx, userID, chains)
```
radazen, line 1208: This can be uncommented after we tweak some OpenSea stuff

service/persist/postgres/token_gallery.go
```diff
@@ -113,7 +113,26 @@ func NewTokenGalleryRepository(db *sql.DB, queries *db.Queries) *TokenGalleryRep
 	deleteTokensOfContractBeforeTimeStampStmt, err := db.PrepareContext(ctx, `UPDATE tokens SET DELETED = true WHERE CONTRACT = $1 AND LAST_SYNCED < $2 AND DELETED = false;`)
 	checkNoErr(err)
 
-	deleteTokensOfOwnerBeforeTimeStampStmt, err := db.PrepareContext(ctx, `UPDATE tokens SET DELETED = true WHERE OWNER_USER_ID = $1 AND CHAIN = ANY($2) AND LAST_SYNCED < $3 AND DELETED = false;`)
+	// user_id = $1, chains = $2, can_delete_holder_tokens = $3, can_delete_creator_tokens = $4, last_synced = $5
+	deleteTokensOfOwnerBeforeTimeStampStmt, err := db.PrepareContext(ctx, `with can_delete as (
+    select tokens.id, tokens.contract from tokens
+        left join token_ownership on tokens.id = token_ownership.token_id and token_ownership.owner_user_id = tokens.owner_user_id
+    where tokens.owner_user_id = $1
+      and (cardinality($2::int[]) = 0 or tokens.chain = any($2))
```
radazen, line 121: I might take the `cardinality` check out of this; upon reflection, I think the GraphQL or `publicapi` layers should be responsible for turning an empty parameter into a set of chains (i.e. treat chains as a filter, where an empty set means "all chains"). By the time we reach the database, I don't think we should be responsible for turning an empty set into "all chains" anymore.

service/persist/postgres/token_gallery.go
```diff
@@ -322,18 +346,27 @@ func (t *TokenGalleryRepository) BulkUpsertTokensOfContract(pCtx context.Context
...
 	tokens, err := t.excludeZeroQuantityTokens(pCtx, pTokens)
 	if err != nil {
 		return time.Time{}, nil, err
 	}
 
+	// If we're not upserting anything, we still need to return the current database time
+	// since it may be used by the caller and is assumed valid if err == nil
 	if len(tokens) == 0 {
-		return time.Time{}, []persist.TokenGallery{}, nil
+		currentTime, err := t.queries.GetCurrentTime(pCtx)
```
radazen, line 358: I added this because I had issues in a scenario where I removed a wallet and had 0 creator tokens to add but many creator tokens to delete. Since I had 0 to add, we returned an empty `time.Time{}`, which was then used in the deletion step (which attempted to delete tokens with `last_synced` prior to `0001-01-01`).

service/persist/postgres/token_gallery.go
```diff
@@ -381,7 +414,7 @@ func (t *TokenGalleryRepository) bulkUpsert(pCtx context.Context, pTokens []pers
 		(*t).LastSynced = upserted[i].LastSynced
 	}
 
-	return upserted[0].LastUpdated, tokens, nil
+	return upserted[0].LastSynced, tokens, nil
```
radazen, line 417: Pretty sure we want to return LastSynced here instead of LastUpdated. They're almost equivalent, but `LastSynced` can't be overwritten by an earlier timestamp, and `LastUpdated` can.

service/tracing/http.go
```diff
@@ -46,7 +46,7 @@ func (t *tracingTransport) RoundTrip(req *http.Request) (*http.Response, error)
 	span, _ := StartSpan(req.Context(), "http."+strings.ToLower(req.Method), fmt.Sprintf("HTTP %s %s", req.Method, req.URL.String()), t.opts...)
 
 	// Send sentry-trace header in case the receiving service can continue our trace
-	req.Header.Add("sentry-trace", span.TraceID.String())
+	req.Header.Set("sentry-trace", span.TraceID.String())
```
radazen, line 49: This resolves an issue where OpenSea would return a `431 - Header too large` error while we were paginating. Our pagination reuses a request object, and `req.Header.Add` would add an additional copy of the header every time we sent the request.

service/multichain/multichain.go
```diff
@@ -560,25 +675,39 @@ func (p *Provider) processTokensForOwnersOfContract(ctx context.Context, contrac
 	return nil
 }
 
-// AddTokensToUser will append to a user's existing tokens
-func (p *Provider) AddTokensToUser(ctx context.Context, tokensFromProviders []chainTokens, contracts []persist.ContractGallery, user persist.User, chains []persist.Chain) ([]persist.TokenGallery, error) {
-	return p.processTokensForUser(ctx, tokensFromProviders, contracts, user, chains, true)
+// AddHolderTokensToUser will append to a user's existing holder tokens
+func (p *Provider) AddHolderTokensToUser(ctx context.Context, tokensFromProviders []chainTokens, contracts []persist.ContractGallery, user persist.User, chains []persist.Chain) ([]persist.TokenGallery, error) {
+	return p.processTokensForUser(ctx, tokensFromProviders, contracts, user, chains, postgres.UpsertOptions{
+		UpsertingCreatorTokens: false,
```
jarrel-b, line 686: Love how much more explicit this is

jarrel-b (approved):
Looks awesome!

PR #1039, opened 2023-06-23
Add ENS PFP support

jarrel-b:
### Changes

This PR adds support for using an ENS PFP as your Gallery PFP. Clients can retrieve the PFP by querying `ensProfileImageByUserId` for the PFP and confirming it by calling the `setProfileImage` mutation with the wallet address returned by the query.

The backend handles standardizes the possible URIs from ENS to an HTTP URL (rewrites IPFS to HTTP, and finds image URLs from NFT avatars) and validates that the user is the owner of the wallet and also a current holder of the token before setting it. The standard technically supports PFPs from other chains, but I left supporting that for another day. We also won't support data URIs (e.g. `data:...`). I checked out the ENS subgraph https://thegraph.com/hosted-service/subgraph/ensdomains/ens and it looks like data URIs and PFP from other chains are hardly ever used (queried for the last 1000 blocks)

The backend doesn't check if a URI points to an actual image (users are able to set it to anything, ENS doesn't do any validation in the UI and users can just call the contract directly), is it frontend able to handle that @Robinnnnn?

We currently cache the PFP indefinitely. I considered re-fetching the PFP at some rate like every 2 days but decided against it for simplicity and since PFPs aren't likely to change frequently. And if a user updates their ENS PFP, they probably don't want to wait 2 days to see it reflected on Gallery. Technically, a user could "refresh" their PFP by calling `setProfileImage` again with the same inputs, which will re-fetch the PFP. At some point, we may want to support manual refreshes of PFPs in the future.

Similar to token PFPs, if a user disconnects the wallet they used for their PFP, we won't return the PFP anymore. One caveat is that we only check if a user owns the token when setting the PFP, and not at request time. This means that a user could still use a PFP from a token they no longer own, but I don't think that's a big deal. We could check for this, but it would mean making calls to Alchemy each time which might be costly.

#### Graphql schema changes
* Adds types: `HTTPSProfileImage` and `EnsProfileImage`
* Adds query `ensProfileImageByUserId`
* Adds support for setting an ENS pfp in `setProfileImage` by providing a wallet address as input
```graphql
type HTTPSProfileImage {
  previewURLs: PreviewURLSet
}

type EnsProfileImage {
  chainAddress: ChainAddress!
  profileImage: ProfileImage
}

union ProfileImage = TokenProfileImage | HTTPSProfileImage

type Query {
  ensProfileImageByUserId(userId: DBID!): EnsProfileImage
}

input SetProfileImageInput {
  # Only one of the following fields should be provided
  tokenId: DBID
  walletAddress: ChainAddressInput
}
```

#### DB changes
* Adds `wallet_id` and `ens_avatar_uri` to `profile_images` table


db/queries/core/query.sql
```diff
@@ -1281,14 +1281,37 @@ with new_image as (
...
+    insert into profile_images (id, user_id, source_type, wallet_id, ens_avatar_uri, deleted, last_updated)
+    values (@profile_id, @user_id, @ens_source_type, @wallet_id, @ens_avatar_uri, false, now())
+    on conflict (user_id) do update set wallet_id = excluded.wallet_id
+        , ens_avatar_uri = excluded.ens_avatar_uri
+        , source_type = excluded.source_type
+        , deleted = excluded.deleted
+        , last_updated = excluded.last_updated
+    returning *
+)
+update users set profile_image_id = profile_images.id from profile_images where users.id = @user_id and not users.deleted returning sqlc.embed(profile_images);
```
jarrel-b, line 1295: This is a bit of a workaround, but naming the CTE the same as the table (`profile_images`) allows for sqlc to embed the ProfileImage model in the generated struct

db/queries/core/query.sql
```diff
@@ -1281,14 +1281,37 @@ with new_image as (
...
+select * from profile_images pfp
+where pfp.id = @id
+	and not deleted
+	and case
+		when source_type = @ens_source_type
+		then exists(select 1 from wallets w where w.id = wallet_id and not w.deleted)
+		when source_type = @token_source_type
+		then exists(select 1 from tokens t where t.id = token_id and not t.deleted)
+		else
+		0 = 1
```
jarrel-b, line 1313: Added this so that the query will return no rows when there's an unknown source type in the table

graphql/resolver/schema.resolvers.helpers.go
```diff
@@ -1934,39 +1933,54 @@ func profileImageToModel(ctx context.Context, pfp db.ProfileImage) (model.Profil
...
+		}
+		urls := previewURLs(ctx, pfp.EnsAvatarUri.String, nil)
+		return &model.HTTPSProfileImage{PreviewURLs: urls}, nil
 	default:
 		return nil, publicapi.ErrProfileImageUnknownSource
 	}
 }
 
-func getPreviewUrls(ctx context.Context, tokenMedia db.TokenMedia, options ...mediamapper.Option) *model.PreviewURLSet {
+func previewURLsFromTokenMedia(ctx context.Context, tokenMedia db.TokenMedia, options ...mediamapper.Option) *model.PreviewURLSet {
```
jarrel-b, line 1951: Refactored this so that we can generate preview URLs from pfp URLs too

tokenprocessing/pipeline.go
```diff
@@ -231,7 +232,7 @@ func (tpj *tokenProcessingJob) retrieveTokenInfo(ctx context.Context, metadata p
 }
 
 func (tpj *tokenProcessingJob) cacheMediaObjects(ctx context.Context, metadata persist.TokenMetadata) (persist.Media, error) {
-	imgURL, animURL, err := findImageAndAnimationURLs(ctx, tpj.token.TokenID, tpj.contract.Address, tpj.token.Chain, metadata, tpj.token.TokenURI, tpj.token.Chain != persist.ChainETH, tpj.pipelineMetadata)
+	imgURL, animURL, err := findImageAndAnimationURLs(ctx, tpj.contract.Address, tpj.token.Chain, metadata, tpj.pipelineMetadata)
```
jarrel-b, line 235: Removed args like the tokenID and the tokenURI since they aren't actually used

tokenprocessing/media.go
jarrel-b, line 1: The bulk of these changes are moving functions and types to media.go so that they're usable from publicapi

service/media/media.go
```diff
@@ -247,3 +254,90 @@ func MediaFromContentType(contentType string) persist.MediaType {
...
+		}
+	}
+
+	if imgURL == "" && vURL == "" {
+		return "", "", ErrNoMediaURLs
+	}
+
+	// Ethereum has a consistent naming standard for metadata fields
+	// but other chains like Tezos do not
+	if chain != persist.ChainETH {
```
jarrel-b, line 301: This used to be an arg that was passed to this function, but moved it to inside the function since we currently only run this if the chain isn't ETH

benny-conn (approved):
So beautiful! The `AvatarRecord` code is so cool!

Robinnnnn:
image

totally!! 👍

PR #1040, opened 2023-06-23
just wallet overrides

benny-conn:
Changes:

- **Added** `WalletOverrides` which is a mapping of chain to many other chains that will be used to allow wallets to be reused across chains they are shared (EVM chains all share same wallet address)
- **Removed** current `ChainOverrides`, it was being used in two places, one in place of the where the new wallet overrides is and another in a place that did not functionally do anything. When syncing contracts it would check whether the contract addresses had overrides, looked like this:

```
    // let's say Arbitrum and ETH are chains here
    for _, chain := range chains {
                 // if chain was Arbitrum this would return ETH
                 // no override for ETH though
		override := p.ChainAddressOverrides[chain]
		for _, contract := range contracts {
		     // because the override is ETH for arbitrum, the ETH contracts would be added to chainsToAddresses under the arbitrum chain, even though contracts are not shared across EVM chains
		     // eth contracts would then get processed unnecessarily by arbitrum providers, potentially returning bad data if the contract does exist on that EVM.
			if contract.Chain == chain || (override != nil && *override == contract.Chain) {
			if contract.Chain == chain {
				chainsToAddresses[chain] = append(chainsToAddresses[chain], contract.Address)
			}
		}
	}
```

### DIFF
--- a/server/inject.go
+++ b/server/inject.go
@@ -51,7 +51,7 @@ func NewMultichainProvider(ctx context.Context, envFunc func()) (*multichain.Pro
 	wire.Build(
 		setEnv,
 		wire.Value(&http.Client{Timeout: 0}), // HTTP client shared between providers
-		defaultChainOverrides,
+		defaultWalletOverrides,
 		task.NewClient,
 		newCommunitiesCache,
 		postgres.NewRepositories,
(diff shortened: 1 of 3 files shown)


jarrel-b (approved):
Love it!!

PR #1049, opened 2023-06-28
Don't overwrite existing contract owner addresses by default

radazen:
This PR changes our contract upsert methods to preserve existing `owner_address` fields by default, with an optional `canOverwriteOwnerAddress` parameter that allows the upsert to modify a non-empty `owner_address`.

**Old behavior**: every contract upsert would overwrite `owner_address` with a new value

**New behavior**: every contract upsert will overwrite `owner_address` if the existing value is null or empty. Otherwise, the existing value will only be overwritten if the new value is not null/empty _and_ the `canOverwriteOwnerAddress` flag is specified.

The practical implications of this change: syncing/refreshing tokens will update contracts, but won't change an existing contract's owner address unless the existing contract has an empty owner address. Syncing creator tokens, however, will always update a contract's owner.

### DIFF
--- a/service/persist/postgres/contract_gallery.go
+++ b/service/persist/postgres/contract_gallery.go
@@ -40,7 +40,15 @@ func NewContractGalleryRepository(db *sql.DB, queries *db.Queries) *ContractGall
 	getByAddressesStmt, err := db.PrepareContext(ctx, `SELECT ID,VERSION,CREATED_AT,LAST_UPDATED,ADDRESS,SYMBOL,NAME,OWNER_ADDRESS,CHAIN,IS_PROVIDER_MARKED_SPAM FROM contracts WHERE ADDRESS = ANY($1) AND CHAIN = $2 AND DELETED = false;`)
 	checkNoErr(err)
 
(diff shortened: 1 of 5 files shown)


service/persist/postgres/contract_gallery.go
```diff
@@ -40,7 +40,15 @@ func NewContractGalleryRepository(db *sql.DB, queries *db.Queries) *ContractGall
...
 	checkNoErr(err)
 
-	upsertByAddressStmt, err := db.PrepareContext(ctx, `INSERT INTO contracts (ID,VERSION,ADDRESS,SYMBOL,NAME,OWNER_ADDRESS,CHAIN) VALUES ($1,$2,$3,$4,$5,$6,$7) ON CONFLICT (ADDRESS,CHAIN) WHERE parent_id IS NULL DO UPDATE SET VERSION = $2, ADDRESS = $3, SYMBOL = $4, OWNER_ADDRESS = $6, CHAIN = $7;`)
+	upsertByAddressStmt, err := db.PrepareContext(ctx, `
+		insert into contracts (id,version,address,symbol,name,owner_address,chain) values ($1,$2,$3,$4,$5,$6,$7)
+			on conflict (address,chain) where parent_id is null do update set
+			version = $2,
+			address = $3,
+			symbol = $4,
+			owner_address = case when nullif(contracts.owner_address, '') is null then $6 else contracts.owner_address end,
```
radazen, line 49: This statement didn't need an override flag like the sqlc upsert did, since we never use it in contexts where we _would_ want to overwrite an existing non-empty value.

Robinnnnn (approved):

jarrel-b (approved):
lgtm!

PR #1051, opened 2023-06-29
refresh and fetch description and profile image

benny-conn:
Changes:

- **Changed** contract dedupe to check for all metadata fields across all contracts received
- **Added** collection image url support to opensea and objkt providers
- **Added** script for refreshing contracts

### DIFF
--- a/service/multichain/multichain.go
+++ b/service/multichain/multichain.go
@@ -1574,6 +1574,7 @@ func contractsToNewDedupedContracts(contracts []chainContracts) []persist.Contra
 			if contract.Descriptors.ProfileImageURL != "" {
 				meta.ProfileImageURL = contract.Descriptors.ProfileImageURL
 			}
+			contractMetadatas[persist.NewChainAddress(contract.Address, chainContract.chain)] = meta
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


service/multichain/multichain.go
```diff
@@ -1539,35 +1540,53 @@ func tokensToNewDedupedTokens(tokens []chainTokens, contracts []persist.Contract
...
+		res = append(res, persist.ContractGallery{
+			Chain:        address.Chain(),
+			Address:      address.Address(),
+			Symbol:       persist.NullString(meta.Symbol),
+			Name:         persist.NullString(meta.Name),
+			OwnerAddress: meta.OwnerAddress,
+			Description:  persist.NullString(meta.Description),
+		})
 	}
 	return res
```
benny-conn, line 1591: This is the most major change in this PR

Robinnnnn (commented):
this is fantastic, benny! super pumped!

@radazen any concern with writing the `CreatorAddress` on the contract? guessing no because this is within a separate `Descriptors` field

radazen:
> this is fantastic, benny! super pumped!
> 
> @radazen any concern with writing the `CreatorAddress` on the contract? guessing no because this is within a separate `Descriptors` field

Shouldn't be a problem! We already map the contract's `CreatorAddress` to the database's `owner_address` field; this just does a better job processing that data before we write it.

benny-conn:
Closing remark for this PR, we should probably run the profile images through the media pipeline at some point instead of using opensea's profile images directly.

PR #1052, opened 2023-06-30
ENS pfp pipeline

jarrel-b:
#### GraphQL schema changes
* Adds `token` to `EnsProfileImage`. Clients should still use `profileImage` as the image for PFPs though, `token` is to reference a token in a detail page. The backend handles the appropriate `profileImage` to use (either from the token or from a URI cached when setting the PFP).
```graphql
type EnsProfileImage @goEmbedHelper {
  ...
  token: Token @goField(forceResolver: true)
}
```

#### DB schema changes
* Adds `ens_domain` to `profile_images` table, this column gets used to derive the token ID for lookups

#### Pipeline changes
* The pipeline now supports runtime options
* Added the option `WithProfileImageKey` to configure runs of ENS tokens to process and download profile images

#### Optimizations
* Profile images are now  served from Gallery for ENS pfps, if we don't have one processed for some reason then we fallback to the URL cached when initially setting the PFP
* We now check to see if we have ens profile pictures already processed from syncs before querying ENS for it
* Changes ethclient connections from `https://` to `wss://` - I was finding responses over websockets to take half the time vs http
* Removes redundant validation calls the `ens-go` library would make

#### Notes
@benny-conn I decided to not store the profile images in the fallback media. I realized the URL could be invalid if no avatar record is set for the token, and the only way to check is to ask ENS if an avatar is set or to try to download the URL and check. At that point, it made sense to me to just ask ENS for the avatar. I think this really only matters for onboarding before a sync has completed, and since a user would only have one wallet at that point it's still pretty quick.


benny-conn (approved):
Looks amazing!

PR #1053, opened 2023-06-30
Chore: Fix "composite literal uses unkeyed fields"

jarrel-b:
My editor allows me to jump to the next error in a file, but it would jump to these lines because of the warning: "composite literal uses unkeyed fields". I could've updated my config, but I thought I'd just update it in the codebase.

This also lets us run `go vet ./...` with no more errors!


radazen (approved):
🧹 🧹 🧹

PR #1063, opened 2023-07-06
Posts

benny-conn:
Changes:

- **Added** a new `posts` table
- **Added** columns to `comments` and `admires` table for referencing posts
- **Added** sql queries for paginating comments and admires and interactions for posts
- **Changed** sql queries for paginating the feed to union posts into the feed event results
- **Added** mutation for creating posts

I had to create a type in postgres called `feed_entity` that is what we use to generate sqlc results from the modified feed queries into a unified struct with the data from both feed events and posts, including both overlapping and non-overlapping data.

```sql
create type feed_entity AS (
    id varchar(255),
    token_ids varchar(255)[],
    caption varchar,
    event_time timestampz,
    source varchar,
    version int,
    owner_id varchar(255),
    group_id varchar(255),
    action varchar(255),
    data jsonb,
    event_ids varchar(255)[],
    deleted boolean,
    last_updated timestamptz,
    created_at timestamptz
);
```

Below is the modified global feed query that serves as a base for the other feed queries:

```sql
SELECT subquery.*::feed_entity
FROM (
    (
        SELECT id, null::varchar(255)[], caption, event_time, 'feed_event'::varchar, version, owner_id, group_id, action, data, event_ids, deleted, last_updated, event_time
        FROM feed_events 
        WHERE deleted = false
        AND (event_time, id) < (sqlc.arg('cur_before_time'), sqlc.arg('cur_before_id'))
        AND (event_time, id) > (sqlc.arg('cur_after_time'), sqlc.arg('cur_after_id'))
    )
    UNION ALL
    (
        SELECT id, token_ids, caption, created_at, 'post'::varchar, version, owner_id, null::varchar(255), null::varchar, null::jsonb, null::varchar(255)[], deleted, last_updated, created_at
        FROM posts 
        WHERE deleted = false
        AND (created_at, id) < (sqlc.arg('cur_before_time'), sqlc.arg('cur_before_id'))
        AND (created_at, id) > (sqlc.arg('cur_after_time'), sqlc.arg('cur_after_id'))
    )
) subquery
ORDER BY 
    CASE WHEN sqlc.arg('paging_forward')::bool THEN (event_time, id) END ASC,
    CASE WHEN NOT sqlc.arg('paging_forward')::bool THEN (event_time, id) END DESC
LIMIT sqlc.arg('limit');
```

Also, the scan for the `FeedEntity` type is a bit wonky, maybe someone knows something about sqlc that I don't that can make this easier? It returns a raw row as a string currently.

```golang
func (f *FeedEntity) Scan(src interface{}) error {
        ...
	source = strings.TrimLeft(source, "(")
	source = strings.TrimRight(source, ")")

	reader := csv.NewReader(strings.NewReader(source))

	// Assuming that the CSV file has only a single row, we can call reader.Read() to get the first row
	fields, err := reader.Read()
	...

	dataPresentStatus := pgtype.Present
	if fields[9] == "" {
		dataPresentStatus = pgtype.Null
	}
	// Parse the fields into the struct values, this sucks
	tokenIDsStrings := strings.Split(fields[1], ",")
	tokenIDsDBIDs, _ := util.Map(tokenIDsStrings, func(i string) (DBID, error) {
		return DBID(i), nil
	})
	eventIDsStrings := strings.Split(fields[10], ",")
	eventIDsDBIDs, _ := util.Map(eventIDsStrings, func(i string) (DBID, error) {
		return DBID(i), nil
	})
	new := FeedEntity{
		ID:          DBID(fields[0]),
		TokenIDs:    DBIDList(tokenIDsDBIDs),
		Caption:     sql.NullString{String: fields[2], Valid: fields[2] != ""},
		EventTime:   sql.NullTime{Time: parseTime(fields[3]), Valid: fields[3] != ""},
		Source:      FeedEntitySource(fields[4]),
		Version:     sql.NullInt32{Int32: parseInt32(fields[5]), Valid: fields[5] != ""},
		OwnerID:     DBID(fields[6]),
		GroupID:     DBID(fields[7]),
		Action:      sql.NullString{String: fields[8], Valid: fields[8] != ""},
		Data:        pgtype.JSONB{Bytes: []byte(fields[9]), Status: dataPresentStatus},
		EventIDs:    DBIDList(eventIDsDBIDs),
		Deleted:     sql.NullBool{Bool: fields[11] == "t", Valid: fields[11] != ""},
		LastUpdated: sql.NullTime{Time: parseTime(fields[12]), Valid: fields[12] != ""},
		CreatedAt:   sql.NullTime{Time: parseTime(fields[13]), Valid: fields[13] != ""},
	}

	*f = new
         return nil
 }
```

Considerations:

- Is it alright that a lot of the post public API logic lies in the `FeedAPI`? 
- I named the poster's id in the DB `actor_id`, is `poster_id` or `owner_id` better? `actor_id` matches events (not feed events) and I feel like a poster is an actor but we could also be more specific and have it be `poster_id`. We could also go less specific and do `user_id` and probably avoid `creator_id` so we don't add any more meaning to the word creator.

TODO: 
- [x] Tests
- [x] Admiring and commenting on posts


sqlc.yaml
benny-conn, line 1: Why it switched the quotes to single quotes? I have no idea.

radazen:
> Also, the scan for the FeedEntity type is a bit wonky, maybe someone knows something about sqlc that I don't that can make this easier? It returns a raw row as a string currently.

I ran into this some time ago, too! As it turns out, it's a Postgres thing (and maybe a general SQL thing?). Postgres makes a type for every table. Similar to the way you've created a `feed_entity` type, there's actually a named type for every table in our database. And you can cast things to those types or query for them, essentially turning an entire table row into a single value (as you've seen with `feed_entity`). I tried experimenting with it as a way to return multiple entity types from a single query (the way `sqlc.embed` does now, but it didn't exist at the time). Ultimately, I ran into some performance issues, and as you've seen with the output format...having to write and maintain a custom scan for the type would be pretty annoying to maintain. Same goes for maintaining the custom type itself: if we add new fields to feed events or posts, we need to update the custom type to reflect that.

Fortunately, I think there's a much easier-to-maintain alternative. A while back, I implemented something similar for retrieving all comments + admires for an event in chronological order. I eventually took it out because we didn't need the functionality anymore, but the gist went like this:

- Do a paginated union query for the _ids_ of the entities we'll want to retrieve:
```
SELECT interactions.created_at, interactions.id, interactions.tag FROM (
    SELECT t.created_at, t.id, @admire_tag::int as tag FROM admires t WHERE @admire_tag != 0 AND t.feed_event_id = $1 AND t.deleted = false
        AND (t.created_at, t.id) < (@cur_before_time, @cur_before_id) AND (t.created_at, t.id) > (@cur_after_time, @cur_after_id)
                                                                    UNION
    SELECT t.created_at, t.id, @comment_tag::int as tag FROM comments t WHERE @comment_tag != 0 AND t.feed_event_id = $1 AND t.deleted = false
        AND (t.created_at, t.id) < (@cur_before_time, @cur_before_id) AND (t.created_at, t.id) > (@cur_after_time, @cur_after_id)
) as interactions
ORDER BY CASE WHEN @paging_forward::bool THEN (created_at, id) END ASC,
         CASE WHEN NOT @paging_forward::bool THEN (created_at, id) END DESC
LIMIT $2;
```

- Loop through the results to figure out which IDs belong to which entity types (admires are tagged with `@admire_tag`, comments are tagged with `@comment_tag` -- arbitrary ints I passed in to identify types, and I was allowing 0 to mean "don't query for this type".

- Fire off two parallel queries: one to fetch comments by their IDs, one to fetch admires by their IDs. Use the original ordering from the paginated query to return them as a GraphQL `Interaction` type, which itself was just a union of possible return types (admires and comments).

The drawback is that you're running multiple queries: one to get paginated IDs/ordering, and then several in parallel to fetch the entities for those IDs. But I think it's a minor drawback, since the queries are all very fast and the performance penalty is minor. And from a maintenance standpoint, it's a big win: the types continue to be independent, and there's very little to keep up-to-date beyond the usual sqlc stuff.

---

As an aside, I was hoping sqlc's new `embed` syntax could take my suggested approach and turn it into a single database call:

```
with ids as (
  get paginated IDs here
)
select sqlc.embed(admires), sqlc.embed(comments) from ids
  left join admires on ids.type = 'admire' and ids.id = admires.id
  left join comments on ids.type = 'comment' and ids.id = comments.id;
```

It _almost_ works, but we run into null scanning issues because one of `admire` or `comment` is going to be completely null for any given row, and sqlc is trying to map stuff to a `coredb.Admire` or `coredb.Comment`, which have non-nullable fields. So for a row that corresponds to an admire, pgx will still try to scan the left-joined comment columns into a `Comment`, but the columns are null, and `Comment` has regular non-nullable fields, and the whole thing falls apart. Ah well.

radazen (commented):
Looking good, Benny! I've been trying to think through the overall architecture of adding a new type of object to the feed, what the implications are, how it scales, etc.

It's tricky, since it could be argued that posts are actually just another `feed_event` -- an event that gets surfaced on the feed. I don't think that's quite right, though; feed events are the aggregation of regular events, and I don't really think of posts as being an event type.

But the alternative also feels kind of messy: if `posts` are their own things that live in their own table, then everything we've built around `feed events` needs to branch to apply to posts. We end up accommodating posts in a way that seems like it'll become cumbersome: we have to query `posts` _and_ `feed_events` for any feed-related query; we have to allow an admire to reference a `post_id` _or_ a `feed_event_id`; calculating the trending feed has to check for `AdmiredFeedEvent` _or_ `AdmiredPost`, etc. Every feed entity type we add in the future will require an additional `and/or` in these cases.

I suggested a view in one of my review comments, and I think maybe we should double down on that as a solution. Assuming they perform sufficiently well and make use of indexes the same way raw queries would, views could be the proper abstraction to unite various "feed entity" types. I'm imagining a two-step process:
- `feed_entity` views (`feed_entities`, maybe `feed_entities_with_interactions` for calculating trending stuff without making the regular `feed_entities` view slower, etc) which provide a common set of columns that all feed entities have: IDs, timestamps, a type, etc
- a utility method that turns a set of entity IDs and types into an `[]any` that we can type assert on and convert to GraphQL models.

graphql/resolver/schema.resolvers.go
```diff
@@ -538,10 +538,11 @@ func (r *galleryUserResolver) Following(ctx context.Context, obj *model.GalleryU
 func (r *galleryUserResolver) Feed(ctx context.Context, obj *model.GalleryUser, before *string, after *string, first *int, last *int) (*model.FeedConnection, error) {
 	events, pageInfo, err := publicapi.For(ctx).Feed.PaginateUserFeed(ctx, obj.Dbid, before, after, first, last)
 	if err != nil {
+		panic("failed to paginate user feed: " + err.Error())
```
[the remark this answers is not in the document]
benny-conn: Left over from dev, thank you!

graphql/schema/schema.graphql
```diff
@@ -932,7 +998,16 @@ type ErrFeedEventNotFound implements Error {
   message: String!
 }
 
-union FeedEventOrError = FeedEvent | ErrFeedEventNotFound | ErrUnknownAction
```
benny-conn, line 935: `FeedEventOrError` getting changed to `FeedEntityOrError` is the only breaking change as far as the frontend is concerned

PR #1065, opened 2023-07-07
Remove uri check in ens resolver

jarrel-b:


graphql/resolver/schema.resolvers.helpers.go
```diff
@@ -1791,27 +1791,30 @@ func profileImageToModel(ctx context.Context, pfp db.ProfileImage) (model.Profil
 		}
 		return &model.TokenProfileImage{Token: token}, nil
 	case persist.ProfileImageSourceENS:
-		if pfp.EnsAvatarUri.String == "" {
```
jarrel-b, line 1794: An empty `EnsAvatarUri` means that either the avatar record from ENS is empty, or it's a URI format the backend doesn't support. However, tokenprocessing supports more URI formats so we should also check if tokenprocessing has a profile image for the token instead of exiting early.

radazen (approved):

PR #1066, opened 2023-07-07
Don't delete tokens

radazen:
## What's new?
In any context where we would have deleted a token, we'll set its `owned_by_wallets` to `{}` instead. For most contexts, this is enough to stop the token from being displayed: it won't show up in galleries or collections, and it won't be visible in the sidebar when editing a gallery.

Some contexts (like looking up a token by its ID) haven't historically filtered by current wallet ownership or contract creator status, and they still don't. This means there are places where we might show a token now where we previously would have considered it deleted. For example: on a community page.

I think letting old tokens hang around on community pages for a little while is okay for now. Given that my next priority will be to rework token ownership / filtering altogether, I don't think it's worth fixing little corner cases like this in the short term.

One corner case that I _did_ fix: filtering unowned tokens out of gallery previews. The editor interface would show my token in the gallery preview, but not in the actual gallery, and that felt a little weird. I updated the `GetGalleryTokenMediasByGalleryID` query to filter tokens based on ownership.

## Why shouldn't we delete tokens?
By _not_ deleting tokens, we can reuse them when someone regains ownership of a token. Disconnecting a wallet and syncing is no longer irreversible: you can reconnect the wallet and sync again, and the original tokens will be restored with all their metadata and positions within collections.

## Anything else?
I also deleted some unused token-related queries.


db/queries/core/query.sql
```diff
@@ -750,14 +736,19 @@ update galleries set collections = @collections, last_updated = now() where gall
 update users set featured_gallery = @gallery_id, last_updated = now() from galleries where users.id = @user_id and galleries.id = @gallery_id and galleries.owner_user_id = @user_id and galleries.deleted = false;
 
 -- name: GetGalleryTokenMediasByGalleryID :many
-select m.* from tokens t, collections c, galleries g, token_medias m where g.id = $1 and c.id = any(g.collections) and t.id = any(c.nfts) and t.deleted = false and g.deleted = false and c.deleted = false and (length(m.media->>'thumbnail_url'::varchar) > 0 or length(m.media->>'media_url'::varchar) > 0) and t.token_media_id = m.id and m.deleted = false and m.active order by array_position(g.collections, c.id),array_position(c.nfts, t.id) limit $2;
+select m.* from collections c, galleries g, token_medias m, users u, tokens t
+    left join contract_creators cc on t.contract = cc.contract_id
```
radazen, line 740: This was a weird one. I couldn't get the `token_ownership` view to join in any sort of performant way -- it always made the query take seconds instead of milliseconds. After trying different approaches for a while, I gave up and used the `contract_creators` view directly, and that gave very fast results. This shouldn't be an issue after we start recording ownership status in the `tokens` table.

jarrel-b (approved):
Looks great!

db/queries/core/query.sql
jarrel-b: I was going to ask why `token_ownership` wasn't used until I refreshed the page, that makes sense! There's no difference in the returned rows using either view?

radazen: As far as I can tell, they should be functionally equivalent! I think I'm souring on views a bit. The abstraction is nice, but it seems a little too easy to cause unintentional performance issues when using them.

Robinnnnn (approved):

PR #1068, opened 2023-07-10
posts admire comment

benny-conn:
Changes:

- **Added** constraints to `admires` and `comments`table to ensure that admires and comments never attach themselves to both a post and a feed event
- **Added** resolvers for creating admires and comments for posts
- **Added** Posts to the return of removing comments and admires

Important changes:

Constraints ensuring mutual exclusivity of `feed_event_id` and `post_id`:
```sql
alter table admires
add constraint post_feed_event_admire_check check (
    (post_id is not null and feed_event_id is null) or 
    (post_id is null and feed_event_id is not null) or
    (post_id is null and feed_event_id is null)
);

alter table comments
add constraint post_feed_event_comment_check check (
    (post_id is not null and feed_event_id is null) or 
    (post_id is null and feed_event_id is not null) or
    (post_id is null and feed_event_id is null)
);
```

Example query using `sqlc.narg` to ensure nullness of either `feed_event_id` or `post_id`:

```sql
-- name: CreateAdmireEvent :one
INSERT INTO events (id, actor_id, action, resource_type_id, admire_id, feed_event_id, post_id, subject_id, data, group_id, caption) VALUES ($1, $2, $3, $4, $5, sqlc.narg('feed_event'), sqlc.narg('post'), $5, $6, $7, $8) RETURNING *;
```

Considerations:

- Would we prefer to have separate mutations for removing comments and admires on posts like we do for creating them? In the same vein, do we want to consolidate the create mutations to be singular and include optional input values? Either way, very easy changes. I feel like as far as removing comments and admire that it is more comfortable to have them be one query and will return null values in the payload for either `FeedEvent` or `Post`, and I can go either way for the creation mutations. The current mutations are called `commentOnFeedEvent` and `admireFeedEvent` so my instincts told me to follow that pattern for posts but we could of course be less explicit and have the backend handle the logic for null input values and switch them to just `comment` and `admire` (or something similarly generic). This would of course break backwards compat.


publicapi/interaction.go
```diff
@@ -758,20 +801,61 @@ func (api InteractionAPI) CommentOnFeedEvent(ctx context.Context, feedEventID pe
 	return commentID, nil
 }
 
-func (api InteractionAPI) RemoveComment(ctx context.Context, commentID persist.DBID) (persist.DBID, error) {
+func (api InteractionAPI) CommentOnPost(ctx context.Context, postID persist.DBID, replyToID *persist.DBID, comment string) (persist.DBID, error) {
```
radazen, line 804: It kind of feels like this shares enough code with commenting on feed events, that maybe it should be a private `comment` method that gets called with appropriate parameters by wrapper methods `CommentOnPost` and `CommentOnFeedEvent`. I may be missing something, but it looks like the only differences would be validation (which we'd probably move to the wrapper methods, at least for validating `postID` and `commentID`) and some fields on the `db.Event`. The same is likely true of admiring posts and admiring feed events!

db/queries/core/query.sql
```diff
@@ -606,10 +608,10 @@ SELECT * FROM notifications
     ORDER BY created_at DESC;
 
 -- name: CreateAdmireNotification :one
-INSERT INTO notifications (id, owner_id, action, data, event_ids, feed_event_id) VALUES ($1, $2, $3, $4, $5, $6) RETURNING *;
+INSERT INTO notifications (id, owner_id, action, data, event_ids, feed_event_id, post_id) VALUES ($1, $2, $3, $4, $5, sqlc.narg('feed_event'), sqlc.narg('post')) RETURNING *;
```
radazen, line 611: Nit: I feel like we should name the parameters `feed_event_id` and `post_id` to keep them consistent with other ID names (and in `CreateCommentNotifications`, too). Ooh! Wait, if you name them with `_id` at the end, does that map them to `persist.DBID` and cancel out the `sqlc.narg`? That'd make sense -- just one more reason for us to turn on the "use pointers for nullable types" sqlc config flag someday.

radazen (approved):
This is great! I agree with your assessment re: whether or not we should have a single mutation for removal, multiple mutations for adding stuff, etc. The implementation looks good to me, and if the frontend folks would prefer something different, we can always make that change! 🚢 🚢 🚢

db/queries/core/query.sql
benny-conn: Yep, it was because of the narg unfortunately, I love the idea of using pointers with sqlc if that is an option in the near future!

PR #1069, opened 2023-07-11
base64 gif support

benny-conn:
Changes:

- **Added** new URIType and detection for base64 gifs


service/rpc/rpc.go
```diff
@@ -535,13 +558,32 @@ func GetDataFromURIAsReader(ctx context.Context, turi persist.TokenURI, ipfsClie
...
+		}
+		uriType := persist.TokenURI(h).Type()
+		if recurseRawReturns && uriType.IsRaw() {
+			full := &bytes.Buffer{}
+			_, err := io.Copy(full, reader)
+			if err != nil {
+				return nil, mediaType, err
+			}
+
+			newRetrieveTimeout := retrieveTimeout - time.Since(timeStart)
```
jarrel-b, line 577: Is this still used? Also makes sense to me just to use the same duration for every call.

jarrel-b (approved):

PR #1071, opened 2023-07-11
posts community page

benny-conn:
Changes:

- **Added** a resolver for posts on the `Community` type so that you can retrieve all of the posts that refer to any token in a given community

SQL query at hand:
```sql
-- name: PaginatePostsByContractID :batchmany
WITH unnest_post_ids AS (
    SELECT posts.id, unnest(token_ids) AS post_token_id
    FROM posts
    WHERE posts.deleted = false
    AND (posts.created_at, posts.id) < (sqlc.arg('cur_before_time'), sqlc.arg('cur_before_id'))
    AND (posts.created_at, posts.id) > (sqlc.arg('cur_after_time'), sqlc.arg('cur_after_id'))
)
SELECT posts.*
FROM unnest_post_ids
JOIN posts ON unnest_post_ids.id = posts.id
JOIN tokens ON tokens.id = unnest_post_ids.post_token_id
WHERE tokens.contract = sqlc.arg('contract')
ORDER BY 
    CASE WHEN sqlc.arg('paging_forward')::bool THEN (posts.created_at, posts.id) END ASC,
    CASE WHEN NOT sqlc.arg('paging_forward')::bool THEN (posts.created_at, posts.id) END DESC
LIMIT sqlc.arg('limit');
```


db/queries/core/query.sql
```diff
@@ -421,6 +421,33 @@ select * from feed_entities join unnest(@feed_entity_ids::text[]) with ordinalit
           case when not @paging_forward::bool then t.pos end asc
   limit sqlc.arg('limit');
 
+-- name: PaginatePostsByContractID :batchmany
```
radazen, line 424: I'm a little concerned with speed here. Given that posts only know `token_id` and not `contract_id`, I think a worst-case performance scenario looks like this: we have to unnest _every post in the database_ and look up its tokens to see if they're part of the contract we're querying for.

It's possible -- though I'm iffy on whether it'd work well in practice -- that we could do a gin index on `token_ids`, and then arrange the query as something like: find all tokens in the database that belong to this contract (fast, indexed), and then find posts where `token_ids` contains at least one of those token IDs (gin index, probably fast, still has to check a potentially large/unbounded number of tokens in the worst case).

I think the best idea may just be to include `contract_ids` in the `posts` table itself. It feels redundant -- we can determine the contract_ids from the token_ids, so why store redundant information? -- but I think the performance is worth the trade-off. My usual concern with redundant storage is something like "when the data changes, we have to update it in both places!". But in this case, we know that a token's `contract_id` will never change, and it's probably fine to store something redundant in the name of a significant performance boost.

radazen (commented):
🔥 🔥 🔥 I left a few comments!

PR #1072, opened 2023-07-12
Add missing condition to CountOwnersByContractId where clause

radazen:
We were missing a condition in the `CountOwnersByContractId` query, causing community load times to be significantly longer than necessary (and returning an inaccurate count, too!).

Adding the appropriate constraint takes query time from ~5 seconds to ~50ms.


Robinnnnn (approved):

jarrel-b:
Massive!

PR #1077, opened 2023-07-18
optimistically sync indexed eth tokens

benny-conn:
Changes:

- **Added** endpoint to token processing service for syncing specific tokens for a user
- **Added** multichain func for syncing specific tokens for a user
- **Added** hook to indexer that will organize tokens found by gallery user and wallet and send them to the token processing service

In my head, there are 3 ways to go about the optimistic sync with the indexer:

1. Every token that is associated with a gallery user gets its own task, token processing processes tokens one by one (with concurrency)
2. Tokens are grouped by user and processed concurrently by user (current solution)
3. Tokens are grouped all together (potentially batched in groups in a future where there are thousands of transfers to gallery users

I went with option 2 after some consideration and discussion with @jarrel-b because it seems to be a good middle ground given that this may not be (and probably won't be based on past discussions about optimistic syncing) the solution for optimistic syncing that we end up with.

This solution works great while users still have the option to manually sync because it is highly likely that the indexer will catch every token that comes in but theoretically a token could be missed if for example the indexer went down and there was a bug in the catchup process. A solution that robin proposed that could be a bit higher lift would be to keep track of what tokens a user received chronologically so that we can sync progressively backwards in chronological order until we encounter a token the user already has so that we don't have to sync all tokens every time. Alchemy and Opensea (not infura, infura doesn't even return transfer time 😢 ) allow you to order tokens chronologically. I believe this is along the lines of what we discussed in our big optimistic sync brainstorm a while back about just syncing all a user's tokens when we see any activity involving a user's wallets rather than trying to keep track of individual tokens as they come in.

As far as costs go, we aren't calling alchemy any more than we already are with this solution (other than the cost of getting metadata for the tokens which we would be doing anyway if the token's were manually synced). If we wanted to run this indexer for other EVM chains it would double our current alchemy usage every time. Still looking into how possible optimistic syncing would be for other chains like Tezos and POAP.


Robinnnnn (commented):
benny, this is magnificent :) i can't believe you whipped this up so quickly. some thoughts on your thoughts.

> As far as costs go, we aren't calling alchemy any more than we already are with this solution (other than the cost of getting metadata for the tokens which we would be doing anyway if the token's were manually synced)

while i agree that theoretically the costs should be ~same, not every user is constantly syncing on gallery, most users aren't sync'd up-to-date at any given time. in other words, 1) all users constantly syncing vs. 2) some users syncing occasionally map to different velocities of tokens being processed / being written to the DB each day (and therefore cost).

_that said_, maybe the difference isn't very big. i would be curious to see how this plays out in practice!

> If we wanted to run this indexer for other EVM chains it would double our current alchemy usage every time.

makes total sense

> Still looking into how possible optimistic syncing would be for other chains like Tezos and POAP.

totally- in the short run, i think it's 100% fair to label on the product that only certain chains support optimistic syncing while others require manual sync – similar to how alchemy's API docs feature a table that compares different features provided for different chains

indexer/hook.go
```diff
@@ -35,6 +39,58 @@ func newContractHooks(queries *indexerdb.Queries, repo persist.ContractRepositor
 	}
 }
 
-func newTokenHooks() []DBHook[persist.Token] {
-	return []DBHook[persist.Token]{}
+func newTokenHooks(tasks *gcptasks.Client, bQueries *coredb.Queries) []DBHook[persist.Token] {
```
Robinnnnn, line 42: roughly how many tokens (or blocks) are processed at a time? is it every poll cycle?

benny-conn: Roughly 2k-4k in a 50 block cycle, each poll is around 10-20 blocks. So around 400-1000?

indexer/hook.go
```diff
@@ -35,6 +39,58 @@ func newContractHooks(queries *indexerdb.Queries, repo persist.ContractRepositor
...
+						ContractAddress: persist.Address(t.ContractAddress),
+						TokenID:         t.TokenID,
+						OwnerAddress:    persist.Address(t.OwnerAddress.String()),
+					})
+				}
+			}
+
+			for userID, tids := range tokensForUser {
+				// send each token grouped by user ID to the task queue
+				err = task.CreateTaskForUserTokenProcessing(ctx, task.TokenProcessingUserTokensMessage{
```
jarrel-b, line 84: This is awesome. Thinking out loud for later, I wonder if there are ways we could group tokens to reduce the amount of processing on our side. One way I can think of is if a token is an ERC1155, we can group that by contract -> owners so that we only need to process the media once, and update the media for those owner's tokens to use that media.

service/multichain/multichain.go
```diff
@@ -295,6 +295,110 @@ func (p *Provider) SyncTokens(ctx context.Context, userID persist.DBID, chains [
...
+
+	}
+
+	go func() {
+		defer close(incomingTokens)
+		defer close(incomingContracts)
+		wg.Wait()
+	}()
+
+	return p.receiveSyncedTokensForUser(ctx, user, chains, incomingTokens, incomingContracts, errChan)
```
jarrel-b, line 398: Noticed that `receivedSyncedTokensForUser` calls `ReplaceHolderTokensForUser`, should it call `AddHolderTokensForUser` instead if optimistic syncing is append onky?

indexer/indexer.go
```diff
@@ -747,14 +785,20 @@ func (i *indexer) runDBHooks(ctx context.Context, contracts []persist.Contract,
 	for _, hook := range i.contractDBHooks {
 		hook := hook
 		wp.Submit(func() {
-			hook(ctx, contracts, statsID)
+			err := hook(ctx, contracts, statsID)
+			if err != nil {
+				logger.For(ctx).WithError(err).Error("Failed to run contract db hook")
```
jarrel-b, line 790: nit: Could we also add the error to the log message? I noticed that with `WithError` it makes it hard to find the actual error in cloud logging

jarrel-b (approved):
This is awesome Benny! Just had one question. Also, is it easy to add a new processing cause for optimistic syncs? I think it could be helpful in debugging.

indexer/hook.go
benny-conn: Very true, I feel like this would work will in multiple parts of the sync process -> token processing processes to cut down on processing! Love it!

PR #1078, opened 2023-07-19
Add viewToken mutation

Rohan-cp:
Screenshot 2023-07-19 at 11 48 38 AM


jarrel-b:
This is awesome Rohan! Thinking about this feature more, I think it makes sense if we skip dispatching the event via `dispatchEvent` and write the view to the db directly. The reason being that normally we dispatch events when we want to do something else with the event like sending a push notification or adding it to the feed. In this case we just want to store the view event, so it makes sense to skip it entirely.

Each publicAPI has an instance to `repos` which has a bunch of "repositories", or ways to read and write to the db. There is an event repo in `service/persist/postgres/event.go` that `dispatchEvent` ends up calling, and there is already a method for adding a token-related event. Take a look at the file and lmk if you have questions. You'll also want to add the event repo as a field to the `repo` struct so that you can use it (you'll need to update the NewRepositories function that constructs the Repositories type in `service/persist/postgres/postgres.go`)

I think the logic would look like:
* check if the user is authed
* if the user is authed, use the event repo to add the view event to the db

Constructing the event can be a little tricky because of its many fields, but really briefly:
```
		ID:             this gets auto-generated by event repo, so no need to fill this in
		ActorID:        this should be the authed user that viewed the token
		Action:         should be the viewed token action type
		ResourceTypeID: should be the token resource type
		CollectionID:   this should be the token id being viewed
		Data:           this is additional metadata we want to add to the event, I think it would be great if we could add the contractID of the token in here if possible so we can easily aggregate by contract in the future if we want
		Caption:        can leave empty
		GroupID:        can leave empty
		GalleryID:      can leave empty
```

Feel free to find some time with me if you have any questions!

graphql/testdata/operations.graphql
```diff
@@ -372,6 +372,20 @@ mutation viewGalleryMutation($galleryId: DBID!) {
   }
 }
 
+mutation viewTokenMutation($tokenID: DBID!) {
```
jarrel-b, line 375: Thanks for adding this, we use this file in our integration test.  I can show you how to add a tests later

graphql/graphql_test.go
```diff
@@ -1316,6 +1316,14 @@ func viewGallery(t *testing.T, ctx context.Context, c genql.Client, galleryID pe
 	_ = (*resp.ViewGallery).(*viewGalleryMutationViewGalleryViewGalleryPayload)
 }
 
+// viewToken makes a GraphQL request to view a token
+func viewToken(t *testing.T, ctx context.Context, c genql.Client, tokenID persist.DBID) {
```
jarrel-b, line 1320: Awesome! Have you tried adding a test to `testGraphQL` in this file? Think it'd be awesome to try adding a test for this. It could be as simple as calling `viewToken`. An example command to run a test is:
```
go test -run TestMain/test_GraphQL/should_create_a_user ./graphql/...
```
In go, the spaces in a test name get replaced with underscores, so "should create a user" becomes "should_create_a_user"

graphql/schema/schema.graphql
```diff
@@ -1890,6 +1890,12 @@ type ViewGalleryPayload {
 
 union ViewGalleryPayloadOrError = ViewGalleryPayload | ErrAuthenticationFailed
 
+type ViewTokenPayload {
+  token: Token
+}
+
+union ViewTokenPayloadOrError = ViewTokenPayload | ErrAuthenticationFailed
```
jarrel-b, line 1897: Can we add `ErrTokenNotFound` as a possible returned type too? For context, the error that can return on the line:
```
token, err := api.loaders.TokenByTokenID.Load(tokenID)
```
is the type `persist.ErrTokenNotFoundByID` and that error eventually gets remapped in this file https://github.com/gallery-so/go-gallery/blob/main/graphql/resolver/schema.resolvers.helpers.go#L157 to the `ErrTokenNotFound` graphQL type, so adding `ErrTokenNotFound` to the `ViewTokenPayloadOrError` and run `go generate` again should be all we need to do

jarrel-b (approved):
This looks awesome @Rohan-cp 🚀 🚀 Nice job figuring out the patterns and getting familiarized with the codebase!

publicapi/token.go
```diff
@@ -516,3 +517,40 @@ func (api TokenAPI) GetTokenOwnershipByTokenID(ctx context.Context, tokenID pers
...
+		}
+		eventPtr, err := api.repos.EventRepository.Add(ctx, db.Event{
+		ActorID:        persist.DBIDToNullStr(userID),
+		Action:         persist.ActionViewedToken,
+		ResourceTypeID: persist.ResourceTypeToken,
+		CollectionID:   tokenID,
+		Data: persist.EventData{
+				TokenContractID: token.Contract,
+			},
+		})
```
Rohan-cp, line 549: @jarrel-b do you think the value passed into the EventData here is correct? If not, not sure how to find the actual value.

graphql/schema/schema.graphql
```diff
@@ -1890,6 +1890,16 @@ type ViewGalleryPayload {
...
 
+type ViewTokenPayload {
+  token: Token
+}
+
+union ViewTokenPayloadOrError =
+    ViewTokenPayload
+  | ErrAuthenticationFailed
+  | ErrTokenNotFound
+  | ErrGalleryNotFound
```
jarrel-b, line 1901: I think it can be unintuitive to get a `ErrGalleryNotFound` error instead of `ErrCollectionNotFound` when the client only provided a collection ID. I left a comment later below to address this!

publicapi/token.go
```diff
@@ -516,3 +517,49 @@ func (api TokenAPI) GetTokenOwnershipByTokenID(ctx context.Context, tokenID pers
...
+	}); err != nil {
+		return db.Event{}, err
+	}
+
+	token, err := api.loaders.TokenByTokenID.Load(tokenID)
+	if err != nil {
+		return db.Event{}, err
+	}
+
+	galleryID, err := api.queries.GetGalleryIDByCollectionID(ctx, collectionID)
```
jarrel-b, line 535: Instead of querying for the gallery by the collection ID, can we retrieve the collection by its ID and get the gallery ID from the returned type? One way to do that is with a dataloader:
```
collection, err := api.loaders.CollectionByCollectionID.Load(collectionID)
```

graphql/graphql_test.go
```diff
@@ -782,6 +783,48 @@ func testGetCommunity(t *testing.T) {
 	})
 }
 
+func testViewToken(t *testing.T) {
```
jarrel-b, line 786: 🚀

jarrel-b (approved):
Awesome job!! Just had a few minor suggestions but lgtm

PR #1085, opened 2023-07-20
Refactor gallery previews

jarrel-b:
Before:
Screen Shot 2023-07-20 at 1 02 01 PM

After:
Screen Shot 2023-07-20 at 1 02 32 PM

**Changes**
* Made a few query optimizations, the idea is that the join to `token_ownership` table is expensive so I tried to subset the results as much as possible prior to the join
  * Subset the collections and tokens to the first 8 collections in a gallery and the first 8 tokens in a collection instead of joining on everything
  * Moved join to `token_medias` before the join to `token_ownership`
  * Ordering in the CTE seemed to improve performance a lot
* Adds a dataloader for fetching gallery previews for batching, noticed in the trace that a lot of time seemed spent waiting for the query to start


Robinnnnn (approved):
🧙 🧙 🧙

PR #1094, opened 2023-07-25
Benny/base provider

benny-conn:
Changes:

- **Added** a new `reservoir` provider that has the option to use `persist.ChainBase`
- **Added** new chain to GQL schema
- **Added** new provider to multichain provider dependency injection

Right now the provider only works for `SyncTokens`


jarrel-b (approved):
Awesome!

Robinnnnn (approved):
😍

PR #1096, opened 2023-07-27
Store token ownership inline

radazen:
## Token Ownership
A token in our database can be a creator token, a holder token, both, or neither. Up until now, we've determined ownership status on-the-fly, typically by joining tokens with the `token_ownership` view. This was necessary because, for example, a user might remove the wallet that gave them holder status on a particular token. The token itself wouldn't be updated until the user's next sync, so we'd have to update queries to say "get me all tokens owned by this user AND owned by a wallet that the user still has."

This worked fine, but adding creator status made it more complicated. The `token_ownership` view was designed to abstract away some of this complexity, but it can be slow and difficult to optimize. Given that tokens are updated infrequently and read often, a less complicated and faster solution would be to store the holder/creator status in the token itself, and that's what this PR does.

The `tokens` table has two new columns: `is_holder_token` and `is_creator_token`. `is_holder_token` is actually a generated column equivalent to `len(owned_by_wallets) > 0`, and `is_creator_token` is a regular column that we update manually.

**To specify that a token must currently be displayable by a user, you can now write:**
```
and (tokens.is_holder_token or tokens.is_creator_token)
```

I considered making _another_ generated column as an alias for this; something like `tokens.is_owned = (tokens.is_holder_token or tokens.is_creator_token)`, but then we'd have to index that column too, and it feels a little redundant overall. If specifying both conditions is annoying or error-prone in practice, we can consider this option.

**[EDIT]** - Actually, after finding some queries I had originally missed, I think we probably _do_ want a `tokens.is_owned` column to make it easier to specify that a token must be a holder or creator token and available to be displayed by the user. The index is small (20MB for 2.1 million tokens in the database) and imho it's worthwhile for convenience. We can get rid of it in the future if we need to.

I'm open to suggestions on the name, though. Maybe `tokens.is_displayable` is better than `tokens.is_owned`, since the token is technically "owned" by the user in the database, and we're using holder/creator status to determine whether the token is allowed to be displayed?

## When do we update tokens?
A token's holder/creator status will now be updated:
- During a regular sync (only updates holder status)
- During a creator sync (only updates creator status)
- After removing a wallet (updates both holder and creator status, only _removes_ status)

Wallet removal itself is synchronous, but to handle the token updates, we create an async task for the `tokenprocessing` service to execute. The updates are idempotent and are safe to run without a lock and without worrying about concurrency.

## Other stuff
- I got rid of `token_ownership`, and rewrote queries that used it. They're much less complicated and much faster now.
- I had to make some judgement calls on any query that had previously only specified `tokens.deleted = false`. Since we no longer delete tokens, specifying `(tokens.is_holder_token or tokens.is_creator_token)` is usually correct for these cases, but some queries (e.g. updating media) looked like we'd want them to update _all_ tokens, even if those tokens aren't currently displayable by the user.
- This PR doesn't include tests. I have a few fast follows planned (e.g. normalizing `wallets`, speeding up the `contract_creators` view, optimizing creator syncing) that would likely break some of the tests I'd write now, but we definitely should have tests for this stuff after all that work is merged in.


db/queries/core/query.sql
```diff
@@ -1465,3 +1454,86 @@ JOIN users ON wallets.id = any(users.wallets)
...
+                        or
+                        (c.override_creator_user_id is null and w.address is not null and array[w.id] <@ u.wallets)
+                    )
+    where c.deleted = false
+      and (u.id is not null or coalesce(nullif(c.owner_address, ''), nullif(c.creator_address, '')) is not null)
+)
+select * from unnest(@contract_ids::text[]) as ids
+                  join contract_creators cc on cc.contract_id = ids;
+
+-- name: GetCreatedContractsByUserID :many
```
radazen, line 1482: I rewrote this to _not_ use `contract_creators`, but once wallets are normalized and I make a few tweaks to the `contract_creators` view, it should be fast enough to be used here again.

db/queries/core/token_gallery.sql
```diff
@@ -39,14 +40,15 @@ insert into tokens
...
-    , case when @upserting_creator_tokens::bool then '{}' else ownership_history[ownership_history_start_idx::int:ownership_history_end_idx::int] end
+    , case when @set_holder_fields::bool then ownership_history[ownership_history_start_idx::int:ownership_history_end_idx::int] else '{}' end
     , media
     , fallback_media
     , token_metadata
     , external_url
     , block_number
     , owner_user_id
-    , case when @upserting_creator_tokens then '{}' else owned_by_wallets[owned_by_wallets_start_idx::int:owned_by_wallets_end_idx::int] end
+    , case when @set_holder_fields then owned_by_wallets[owned_by_wallets_start_idx::int:owned_by_wallets_end_idx::int] else '{}' end
```
radazen, line 50: I keep iterating on this. At this point, I think having flags to indicate whether creator/holder fields should be updated is probably the best approach, since we may want to update one, both, or neither.

service/multichain/multichain.go
```diff
@@ -1549,6 +1556,32 @@ func tokensToNewDedupedTokens(tokens []chainTokens, contracts []persist.Contract
...
+			creatorAddress = contract.CreatorAddress
+		}
+
+		if wallet, ok := addressToWallets[contract.Chain.NormalizeAddress(creatorAddress)]; ok {
+			// TODO: Figure out the implication for L2 chains here. Might want a function like
+			// Chain.IsCompatibleWith(Chain) to determine whether a wallet on one chain can claim
+			// ownership of a contract on a different chain.
+			createdContracts[contractAddress] = wallet.Chain == contract.Chain
+		} else {
+			createdContracts[contractAddress] = false
```
jarrel-b, line 1581: Would we want to check if `createdContracts[contractAddress]` is not already true from an override?

jarrel-b (approved):
FYI I glossed over most of the SQL changes besides the new queries, but the new queries look good to me!

service/persist/postgres/user.go
```diff
@@ -448,33 +448,34 @@ func (u *UserRepository) AddWallet(pCtx context.Context, pUserID persist.DBID, p
...
+		return false, nil
 	}
 
 	if _, err := tx.StmtContext(pCtx, u.deleteWalletStmt).ExecContext(pCtx, pWalletID); err != nil {
-		return err
+		return false, err
 	}
 
-	return tx.Commit()
+	return true, tx.Commit()
```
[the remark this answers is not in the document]
radazen: Doh. Yes, we definitely should. Thanks for catching that!

service/multichain/multichain.go
radazen: Hmm, I don't _think_ it's possible. I'd only expect a particular contractAddress to come through this loop once, and if it did happen to come through twice (i.e. we had duplicate contracts passed in), I'd expect it to have all of the same values (including the same override, which should trigger the `continue`).

That being said, looking at this loop made me realize that I don't love using `contractAddress` as a key, since it's possible that we could have a collision between multiple contracts with the same address on different chains.

jarrel-b: Ah I totally missed the `continue`, makes sense!

PR #1097, opened 2023-07-27
Add Curated Feed

jarrel-b:
### Changes

This PR adds the Curated Feed to the GraphQL schema. The Trending Feed uses two factors when scoring an event: the age of the event, and the number of interactions it has. The Curated Feed considers time, the number of interactions, and an additional personalization factor. The new algorithm is the product of these three factors, where all factors are weighted the same.
```
scoreFeedEntity(entity) = timeFactor * engagementFactor * personalizationFactor
```
The personalization factor is computed by calculating three different terms: 
* **Is the user in the same friend circle as the viewer?** The assumption is if Robin is friends with Benny, and I am friends with Robin, then content posted by Benny is more relevant to me.
* **Is the content posted relevant to the viewer?** If the community that is being shared is commonly held/displayed with tokens that I own, that content is more likely to be relevant to me. This means that if other users from a new project frequently display it with tokens that I own, it is more likely to show up on my feed.
* **Is the user similar to the viewer in terms of following and ownership?** If we follow the same people, and belong (own) to the same communities, then we have similar tastes in how we engage with users on Gallery.

The terms are additive and can be calculated as below, where all terms are weighted the same:
```
personalizationFactor(entity) = socialScore + relevanceScore + similarityScore
```

The scoring algorithm can score things quickly by storing various matrices in memory, allowing the Curated Feed to be computed in real-time per request vs. having an offline service generate feeds per user. The matrices are updated periodically, in a similar way to the existing user recommendation algorithm: a read/write mutex limits access to the matrix. Requests are able to read without getting blocked but do get blocked during an update. The Curated Feed should run with just slightly more overhead than the Trending Feed. When Gallery gets to the point that there are too many events in a window where it's unreasonable to fetch and compute in real-time, we can consider fanning out the events to only a subset of Gallery users.

There are a few compromises I made in order to keep the service in the backend and to be able to run it in real-time:
1) The biggest change is that it only considers communities that are displayed. The reason is that the universe of users displaying and communities that are displayed is much smaller than all the users and communities on Gallery. The biggest impact is that users that don't actively update their Gallery won't get the full benefits of the algorithm. Using only displayed communities is a higher signal, however, and it also means that there are fewer communities to check relevance against.
2) The algorithm uses breadth-first search to check if a user is in the same friend circle as another. To avoid going traversing too far, it stops when the depth is greater than four (i.e. a friend of a friend of a friend of a friend).

#### Improvements
* Considering all of a user's tokens (not just displayed) when computing relevance
* Augmenting the social graph with data from Twitter, Lens, Farcaster, etc.
* Using other signals outside of token ownership and displaying such as a user's admires, comments, posts, etc.
* The user recommendation algorithm uses a different data structure, but it can be replaced with the following matrix added in this PR


radazen (commented):
The amount of effort, experimentation, trial and error, etc that went into these data structures and implementations is so evident and impressive. Super excited to see this on dev! 🚀 @jarrel-b and I talked through a few things offline. Documenting them here for visibility:
- Might be worth renaming to something more descriptive than `koala`
- If possible, maybe we could use redis to store the data used to calculate scores, such that each new instance of the backend wouldn't need to build all of the data structures from scratch at startup (the concern being that, if we're scaling up due to traffic, anyone hitting a new instance would see a long response time while waiting for the instance to start up)
- If `RelevanceTo` only gets called in a loop, it may be worthwhile to acquire the read lock once for the whole loop instead of acquiring it and releasing it for every iteration of the loop

And also, just to reiterate: this is awesome.

PR #1100, opened 2023-07-27
types and backfill

benny-conn:
Changes:

- **Added** backfill script for farcaster
- **Added** types for new social account types (farcaster/lens)

### DIFF
--- a/cmd/farcaster_backfill/main.go
+++ b/cmd/farcaster_backfill/main.go
@@ -30,7 +30,7 @@ func main() {
 	ctx := context.Background()
 
 	// get every wallet with their owner user ID
-	rows, err := pg.Query(ctx, `select users.id, wallets.address from users join wallets on wallets.id = any(users.wallets) where users.deleted = false and wallets.chain = 0 order by users.created_at desc;`)
+	rows, err := pg.Query(ctx, `select users.id, wallets.address from users join wallets on wallets.id = any(users.wallets) where users.deleted = false and wallets.chain = 0 and wallets.deleted = false order by users.created_at desc;`)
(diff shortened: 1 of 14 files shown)


cmd/farcaster_backfill/main.go
```diff
@@ -0,0 +1,108 @@
...
+	pg := postgres.NewPgxClient()
+
+	neynar := farcaster.NewNeynarAPI(&http.Client{Timeout: 10 * time.Second})
+
+	queries := coredb.New(pg)
+
+	ctx := context.Background()
+
+	// get every wallet with their owner user ID
+	rows, err := pg.Query(ctx, `select users.id, wallets.address from users join wallets on wallets.id = any(users.wallets) where users.deleted = false and wallets.chain = 0 order by users.created_at desc;`)
```
jarrel-b, line 33: Would this select universal users too, or would we want to process them too?

cmd/farcaster_backfill/main.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/spf13/viper"
+)
+
+func main() {
+
+	setDefaults()
+
+	pg := postgres.NewPgxClient()
+
+	neynar := farcaster.NewNeynarAPI(&http.Client{Timeout: 10 * time.Second})
```
jarrel-b, line 26: https://en.wikipedia.org/wiki/Neymar

jarrel-b (approved):

cmd/farcaster_backfill/main.go
benny-conn: True! We can filter them out!

PR #1102, opened 2023-07-31
Add polygon spam

jarrel-b:
This PR adds polygon spam processing to tokenprocessing's `detect-spam` job. It also normalizes EVM chains to be lower-cased (instead of only ethereum). I also double-checked that the EVM contracts in our database are already lower-cased.


Robinnnnn (approved):

radazen (approved):
Looks good to me! 🎉

PR #1107, opened 2023-07-31
lens gql and backfill

benny-conn:
Changes:

- **Added** `Lens` to the gql social providers
- **Added** a backfill script for setting default lens profiles


Robinnnnn (commented):
let's just mark the contracts as spam, not the tokens!

PR #1113, opened 2023-08-02
Small dataloader changes for feed

jarrel-b:
Modifies the batch settings for the tokens and contracts loader. Also adds options to configure the default dataloader settings.

### DIFF
--- a/graphql/dataloader/dataloaders.go
+++ b/graphql/dataloader/dataloaders.go
@@ -60,8 +60,6 @@ import (
 
 	"github.com/mikeydub/go-gallery/util"
 
-	"github.com/mikeydub/go-gallery/service/tracing"
-
 	"github.com/jackc/pgx/v4"
 	db "github.com/mikeydub/go-gallery/db/gen/coredb"
 	"github.com/mikeydub/go-gallery/service/persist"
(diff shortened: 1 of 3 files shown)


graphql/dataloader/dataloaders.go
```diff
@@ -269,9 +252,11 @@ func NewLoaders(ctx context.Context, q *db.Queries, disableCaching bool) *Loader
 
 	loaders.NewTokensByFeedEventID = NewTokensLoaderByID(defaults, loadNewTokensByFeedEventID(q))
 
-	loaders.ContractByContractID = NewContractLoaderByID(defaults, loadContractByContractID(q), ContractLoaderByIDCacheSubscriptions{
-		AutoCacheWithKey: func(contract db.Contract) persist.DBID { return contract.ID },
-	})
+	loaders.ContractByContractID = NewContractLoaderByID(
+		settingsWithOptions(ctx, disableCaching, withMaxBatchOne(500), withWaitTime(5*time.Millisecond)),
```
jarrel-b: This should have a negligible impact for users with few tokens (since we generally also resolve the contract for each token), but should help a lot for users with many tokens like Kyt who has over 10k. In my testing, this reduced kyt's load time by around half.

graphql/resolver/schema.resolvers.go
```diff
@@ -2577,6 +2577,9 @@ func (r *tokenResolver) OwnerIsCreator(ctx context.Context, obj *model.Token) (*
 
 // Contract is the resolver for the contract field.
 func (r *tokenResolver) Contract(ctx context.Context, obj *model.Token) (*model.Contract, error) {
+	if obj.HelperTokenData.Token.Contract != "" {
```
jarrel-b, line 2580: I'm not sure if this will help much because the token would already be cached at this point. But it does mean that it wouldn't need to call to acquire the lock on the token loader's cache and lookup the key, so it should be very slightly faster and maybe make batching more consistent.

Robinnnnn (approved):

PR #1120, opened 2023-08-04
rasterizing service

benny-conn:
Changes:

- **Added** an express service for rasterizing SVGs
- **Changed** token processing svg rasterizer to call express service instead of calling script directly

I deployed the rasterizer to prod and dev and added the necessary env vars for it to get called. Do we want to include auto deploys in the make file for this? I imagine we don't really want to be re-deploying this service ever unless we specifically make changes to it.


Robinnnnn (approved):
🔥 🔥 🔥 🔥 🔥 🔥

PR #1122, opened 2023-08-06
Push notification error handling

radazen:
An unmarshaling error has stopped us from seeing some push notification errors. This PR fixes that issue (changes a struct field from `map[string]string` to `map[string]any` and adds some additional logging. It also fixes an issue where errors weren't being propagated through push notification batches.

I don't think this will fix the reliability issues we've been seeing with push notifications, but hopefully it'll let us see actual errors explaining what's wrong.


Robinnnnn (approved):
thanks for doing this!

PR #1127, opened 2023-08-07
get rid of the token columns that cause inconsistencies

benny-conn:
Changes:

- **Removed** `media` and `token_metadata` columns from the `tokens` table in the DB and all code associated with those columns. We should only be using the `token_media` columns now.
- **Removed** the `tokenMetadata` field on the `Token` type in GQL, wasn't being used, wasn't being populated correctly


db/migrations/core/000103_unused_token_fields.up.sql
```diff
@@ -0,0 +1,2 @@
+alter table tokens drop column media cascade;
+alter table tokens drop column token_metadata cascade;
```
benny-conn, line 2: I was forced to add cascade, I imagine because of `pii.user_view`?

service/persist/postgres/contract_gallery.go
```diff
@@ -63,7 +63,7 @@ func NewContractGalleryRepository(db *sql.DB, queries *db.Queries) *ContractGall
 	getUserByWalletIDStmt, err := db.PrepareContext(ctx, `SELECT ID,USERNAME FROM users WHERE WALLETS @> ARRAY[$1]:: varchar[] AND DELETED = false`)
 	checkNoErr(err)
 
-	getPreviewNFTsStmt, err := db.PrepareContext(ctx, `SELECT MEDIA->>'thumbnail_url' FROM tokens WHERE CONTRACT = $1 AND DISPLAYABLE AND DELETED = false AND OWNED_BY_WALLETS && $2 AND LENGTH(MEDIA->>'thumbnail_url') > 0 ORDER BY ID LIMIT 3`)
+	getPreviewNFTsStmt, err := db.PrepareContext(ctx, `SELECT token_medias.MEDIA->>'thumbnail_url' FROM tokens LEFT JOIN token_medias ON token_medias.ID = tokens.TOKEN_MEDIA_ID WHERE tokens.CONTRACT = $1 AND tokens.DISPLAYABLE AND tokens.DELETED = false AND tokens.OWNED_BY_WALLETS && $2 AND LENGTH(token_medias.MEDIA->>'thumbnail_url') > 0 ORDER BY tokens.ID LIMIT 3`)
```
benny-conn, line 66: Can't believe this has just been sitting around fetching outdated previews!

graphql/schema/schema.graphql
```diff
@@ -432,7 +432,6 @@ type Token implements Node @goEmbedHelper {
   ownershipHistory: [OwnerAtBlock]
   ownerIsHolder: Boolean
   ownerIsCreator: Boolean
-  tokenMetadata: String # source is map[string]interface{} on backend, not sure what best format is here
```
radazen, line 435: Just want to double-check: are we _sure_ the frontend never queries for this? Even if they don't use the result, I think it'd be a breaking change if they're actually including it in any queries.

But also, here's hoping they aren't, because I'd love to see it go away.

service/multichain/multichain.go
```diff
@@ -707,9 +707,9 @@ func (p *Provider) prepTokensForTokenProcessing(ctx context.Context, tokensFromP
 
 		// Add already existing media to the provider token if it exists so that
 		// we can display media for a token while it gets handled by tokenprocessing
-		if !token.Media.IsServable() && existingToken.Media.IsServable() {
+		if !token.TokenMedia.IsServable() && existingToken.TokenMedia.IsServable() {
 			// TODO remove
```
radazen, line 711: I saw this during a recent PR and was wondering: are we at a point where we can do the `TODO remove` part now? Or do we still need this?

radazen (approved):
Looks good to me! I _love_ the amount of stuff getting removed in this PR. 🔥 🔥 🔥

graphql/schema/schema.graphql
benny-conn: Gosh darnit, they use it for POAP and Mirror tokens. I will add it back :(((

service/multichain/multichain.go
benny-conn: I think you are right, I will explore the purpose of this originally but I am 99% sure we can remove these now.

PR #1129, opened 2023-08-07
backfill connections

benny-conn:
Changes:

- **Added** `external_social_connections` table for keeping track of a user's social connections
- **Added** backfill script for initially populating social connections

Considerations:

- I imagine the only information we find useful is if a user has a "gallery connection" meaning that they follow someone on farcaster that has an account on gallery. We have no need to store any of the other connections? The only use case I can find useful for storing the non-gallery connections would be if we want to have some sort of on-signup trigger that when a user signs up, if they have a farcaster account (or any social account) it will automatically update the connections if they exist in another user's social connections. Do we want to prepare for that now? We could also store on the `pii.for_users.pii_socials->'Farcaster'` JSONB the raw connections for the purpose of regenerating connections easily? (or cache in redis)
- This logic in the backfill I imagine will be offloaded onto some cron job that will continually update social connections for users, and we will also have some sort of on-signup trigger that includes checking for the existence of social profiles and updating social connections.
- I opted for not including any social ids in this table given that we can query them separately from the user and I don't imagine it will be useful in any case. It also would put this table in "should be pii" territory. Do we see any use case where we would want to double up and include the social IDs of both follower/followee on this table?


cmd/farcaster_backfill/connections/main.go
```diff
@@ -0,0 +1,118 @@
...
+	pg := postgres.NewPgxClient()
+
+	neynar := farcaster.NewNeynarAPI(&http.Client{Timeout: 10 * time.Second})
+
+	queries := coredb.New(pg)
+
+	ctx := context.Background()
+
+	// get every farcaster gallery user
+	rows, err := pg.Query(ctx, `select u.id,u.pii_socials->'Farcaster'->>'id' from pii.user_view u where u.pii_socials->>'Farcaster' is not null and not u.pii_socials->>'Farcaster' = '' and u.deleted = false and u.universal = false order by last_updated desc;`)
```
jarrel-b, line 33: Is there a job to update farcaster data, or was this a one-time thing for now?

jarrel-b:
This looks awesome!  Do you think it's a heavy lift if we store all the connections? I can see it being really cool to send a notification like "xyz from Farcaster joined Gallery, want to follow them?".  Although it would suck to store all this data for a user that may never get used.

But I also think it's fine to only store the connections of existing Gallery users since it sounds like the priority atm is to make recommendations for the new user, and the future job will add the connection for existing users eventually. Hopefully, the new user adds everyone they follow from Farcaster on Gallery, and those existing users will end up following them on Gallery anyway.

jarrel-b (approved):

cmd/farcaster_backfill/connections/main.go
benny-conn: One time thing for now, this script can practically just be deployed as a job so once we confirm how we want to go about keeping everything up to date we can create the job. I think I'd want it to be running as a job on a timeline but also be paired with the triggers on signup.

PR #1133, opened 2023-08-08
seen provider

benny-conn:
Changes:

- **Added** `ProviderID` to `BlockchainInfo` so that each provider can get a unique hard coded ID we can dedupe on.
- **Changed** `providersMatchingInterface` to perform the two checks listed below

1. Check if we have already seen a provider with the given provider ID
2. If a provider is a `ProviderSupplier` and matches the interface we are checking, add that provider's sub provider IDs to the duplicate checking map so that we don't add the individual sub providers later

### DIFF
--- a/service/multichain/multichain.go
+++ b/service/multichain/multichain.go
@@ -52,8 +52,9 @@ type Provider struct {
 
 // BlockchainInfo retrieves blockchain info from all chains
 type BlockchainInfo struct {
-	Chain   persist.Chain `json:"chain_name"`
-	ChainID int           `json:"chain_id"`
+	Chain      persist.Chain `json:"chain_name"`
+	ChainID    int           `json:"chain_id"`
+	ProviderID int           `json:"provider_id"`
 }
 
 // ChainAgnosticToken is a token that is agnostic to the chain it is on
@@ -144,7 +145,7 @@ type errWithPriority struct {
 
 // Configurer maintains provider settings
 type Configurer interface {
-	GetBlockchainInfo(context.Context) (BlockchainInfo, error)
+	GetBlockchainInfo() BlockchainInfo
 }
 
 // NameResolver is able to resolve an address to a friendly display name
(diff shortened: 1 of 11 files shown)


service/multichain/multichain.go
```diff
@@ -205,16 +206,33 @@ type WalletOverrideMap = map[persist.Chain][]persist.Chain
...
 func providersMatchingInterface[T any](providers []any) []T {
 	matches := make([]T, 0)
+	seen := map[int]bool{}
 	for _, p := range providers {
-		if it, ok := p.(T); ok {
-			matches = append(matches, it)
+
+		if conf, ok := p.(Configurer); ok && seen[conf.GetBlockchainInfo().ProviderID] {
+			continue
+		} else if ok {
```
radazen, line 214: Is there a legitimate/expected case where `p` doesn't implement `Configurer`? If not, should we add an `else` here and throw an error indicating that we have a provider that doesn't support `Configurer`?

service/multichain/multichain.go
```diff
@@ -52,8 +52,9 @@ type Provider struct {
 
 // BlockchainInfo retrieves blockchain info from all chains
 type BlockchainInfo struct {
-	Chain   persist.Chain `json:"chain_name"`
-	ChainID int           `json:"chain_id"`
+	Chain      persist.Chain `json:"chain_name"`
+	ChainID    int           `json:"chain_id"`
+	ProviderID int           `json:"provider_id"`
```
radazen, line 57: Given that these IDs are spread among various files, I feel like picking a unique `int` for each provider is a little error-prone, since the only way to be sure you're picking a unique ID is to look at all the other ones. Maybe we use strings and give them unique names that specify what provider/chain/etc they're providing? Or literally just paste in generated UUIDs to make it quick and easy to guarantee uniqueness?

radazen (approved):
Thanks for doing this, Benny! I like this approach a lot. 🎉

service/multichain/multichain.go
benny-conn: Pretty much no. I was going to require that the input not be `any` and instead just be `Configurer` since they all implement it but that made `wire` very unhappy. I love the error idea!

service/multichain/multichain.go
benny-conn: Lovely, I think string with the name of basically the file it is in will do just fine!

PR #1135, opened 2023-08-09
Creator Sync Optimization

radazen:
## What's new?
- Replaced the existing `syncCreatedTokens` mutation with `syncCreatedTokensForNewContracts` and `syncCreatedTokensForExistingContract`
  - `syncCreatedTokensForNewContracts` will sync tokens for all contracts the user has created but does not currently have any creator tokens from
  - `syncCreatedTokensForExistingContract` will sync tokens for any contract the user is the creator of, which is useful for fetching new tokens if the contract is still actively creating them
- Refactored the bulk upsert flow
  - The logic for handling creator/holder tokens can be verbose, and ideally we'd only have to write and maintain one version of it

## Upsert refactoring
We had two main token syncing paths:
- `BulkUpsertByOwnerUserID`: insert tokens for a user, delete tokens not involved in the upsert (scoped to chains)
- `BulkUpsertTokensOfContract`: insert tokens for many users (but scoped to a single contract), delete tokens (from that contract) not involved in the upsert

I needed similar functionality, but with different scoping for the delete:
- `BulkUpsertTokensOfContractForUser`: insert tokens for a user (scoped to a single contract), delete tokens (from that contract + user) not involved in the upsert

It seemed to me that the upsert was always the same, and the deletion tended to have many permutations. So I removed `BulkUpsertByOwnerUserID` and `BulkUpsertTokensOfContract`, and replaced them with a single `UpsertTokens` method with parameters to control the insertion and delete steps:

```
type TokenUpsertParams struct {
	SetCreatorFields bool
	SetHolderFields  bool

	// If OptionalDelete is nil, no delete will be performed
	OptionalDelete *TokenUpsertDeletionParams
}

type TokenUpsertDeletionParams struct {
	DeleteCreatorStatus bool
	DeleteHolderStatus  bool
	OnlyFromUserID      *persist.DBID
	OnlyFromContracts   []persist.DBID
	OnlyFromChains      []persist.Chain
}
```

My `UpsertTokens` method was being called by two methods:
- `processTokensForUser`
- `processTokensForOwnersOfContract`

These two methods were also very similar to one another, and I wanted to see if I could collapse them into a single method with some additional parameters to handle their differences. `processTokensForOwnersOfContract` did _almost_ the same thing as calling `processTokensForUser` in a loop for each owner of the contract, except that it aggregated the tokens so it could do a single upsert/delete step for all tokens.

One thing I didn't like about `processTokensForOwnersOfContract` was the fact that it would look up all tokens for every user who owned tokens from the specified contract. I.e. to figure out each user's existing tokens for a given contract, it would look up _all tokens_ for each of those users. That's a lot of tokens and a lot of sequential DB calls.

In the end, I changed the flow to look like this:
- `processTokensForUser` and `processTokensForOwnersOfContract` are trivial wrappers around `processTokensForUsers`
- `processTokensForUsers` takes upsert/deletion parameters and maps of users, provider tokens, and their existing tokens
  - For existing tokens, `processTokensForUser` still looks up all of a user's tokens and passes them to `processTokensForUsers`. `processTokensForOwnersOfContract` looks up all existing tokens in the _contract_, and then builds a map of `userID => tokensFromContract` to pass as existing tokens for each user

Moving forward, I think we should be able to use `processTokensForUsers` as the basis of any token upsert, because it lets us control what tokens we upsert, what tokens we compare them against to see if they already exist in the database, what kind of fields the upsert can set (creator fields, holder fields, both), and what kind of deletion (if any) should happen.


graphql/schema/schema.graphql
```diff
@@ -2445,7 +2461,12 @@ type Mutation {
   setSpamPreference(input: SetSpamPreferenceInput!): SetSpamPreferencePayloadOrError @authRequired
 
   syncTokens(chains: [Chain!]): SyncTokensPayloadOrError @authRequired
-  syncCreatedTokens(input: SyncCreatedTokensInput!): SyncCreatedTokensPayloadOrError @authRequired
```
radazen, line 2448: We weren't calling this method from the frontend yet, so removing it isn't a breaking change.

service/multichain/multichain.go
```diff
@@ -441,9 +441,9 @@ outer:
 
 	var newTokens []persist.TokenGallery
 	if replace {
-		_, newTokens, err = p.ReplaceHolderTokensForUser(ctx, tokensFromProviders, persistedContracts, user, chains)
```
radazen, line 444: Just doing some parameter reordering for consistency among functions

util/helpers.go
```diff
@@ -183,6 +183,15 @@ func Map[T, U any](xs []T, f func(T) (U, error)) ([]U, error) {
 	return result, nil
 }
 
+// MapWithoutError applies a function to each element of a slice, returning a new slice of the same length.
+func MapWithoutError[T, U any](xs []T, f func(T) U) []U {
```
radazen, line 187: I'd like to rename this to `Map` in a follow-up PR, and rename `Map` to `MapWithError`. It seems _much_ more common to map stuff that doesn't return an error, and having a version with a single return value means you can use it inline instead of declaring an empty variable to hold the nil error value.

service/multichain/multichain.go
```diff
@@ -731,116 +749,166 @@ func (p *Provider) prepTokensForTokenProcessing(ctx context.Context, tokensFromP
...
-		tokens, newTokens, err := p.prepTokensForTokenProcessing(ctx, chainTokensForUsers[userID], contracts, user)
+		tokens, newTokens, err := p.prepTokensForTokenProcessing(ctx, chainTokensForUsers[userID], existingTokensForUsers[userID], contracts, user)
 		if err != nil {
-			return err
+			return nil, nil, err
 		}
 
-		start := len(tokensToUpsert)
 		tokensToUpsert = append(tokensToUpsert, tokens...)
-		userTokenOffsets[userID] = [2]int{start, start + len(tokens)}
```
radazen, line 747: I think there was a subtle bug in the token offset approach here. The offsets are determined here, but the tokens returned by `BulkUpsertTokensOfContract` aren't guaranteed to line up with those offsets, because `BulkUpsertTokensOfContract` deduplicates tokens and removes zero quantity tokens.

I ended up looping through the persisted tokens and building a map that way instead.

service/multichain/multichain.go
```diff
@@ -1258,15 +1326,16 @@ outer:
 		return err
 	}
 
-	contract, err := p.Queries.GetContractByChainAddress(ctx, db.GetContractByChainAddressParams{
```
radazen, line 1261: I could be wrong about this, but I'm _pretty sure_ the above call to `p.processContracts(ctx, contractsFromProviders, false)` should return a single persisted `db.Contract` struct, such that we don't need to look up the contract again here.

service/multichain/multichain.go
```diff
@@ -467,60 +467,83 @@ func (p *Provider) SyncCreatedTokens(ctx context.Context, userID persist.DBID, c
...
-	contracts, _ := util.Map(rows, func(row db.GetCreatedContractsByUserIDRow) (db.Contract, error) { return row.Contract, nil })
+	contracts := util.MapWithoutError(rows, func(r db.GetCreatedContractsByUserIDRow) db.Contract { return r.Contract })
+
+	// Sync the individual contracts in parallel, so contracts with a lot of tokens won't block
+	// contracts with fewer tokens, and progress isn't all or nothing (because a subsequent sync will omit
+	// any contract that finished syncing in this attempt)
+	if len(contracts) > 0 {
+		wp := pool.New().WithErrors().WithContext(ctx)
+		for _, contract := range contracts {
+			contract := contract
```
jarrel-b, line 489: Sounds like newer versions of go will address this soon! https://github.com/golang/go/wiki/LoopvarExperiment

jarrel-b (approved):

service/persist/postgres/token_gallery.go
```diff
@@ -257,76 +246,53 @@ func (t *TokenGalleryRepository) GetByTokenID(pCtx context.Context, pTokenID per
...
-		if err != nil {
-			return nil, fmt.Errorf("failed to delete tokens: %w", err)
-		}
+	chains := util.MapWithoutError(d.OnlyFromChains, func(c persist.Chain) int32 { return int32(c) })
+	return db.DeleteTokensBeforeTimestampParams{
+		RemoveCreatorStatus: d.DeleteCreatorStatus,
+		RemoveHolderStatus:  d.DeleteHolderStatus,
+		OnlyFromUserID:      sql.NullString{String: userID, Valid: d.OnlyFromUserID != nil},
+		OnlyFromChains:      chains,
+		OnlyFromContractIds: util.StringersToStrings(d.OnlyFromContracts),
```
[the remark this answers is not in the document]
radazen: Neither did I! I actually wrote my own helper to do it, and then ended up seeing this one later.

service/multichain/multichain.go
radazen: OMG FINALLY

service/multichain/multichain.go
```diff
@@ -731,116 +749,166 @@ func (p *Provider) prepTokensForTokenProcessing(ctx context.Context, tokensFromP
...
+
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to delete tokens: %w", err)
+		}
+
+		logger.For(ctx).Infof("deleted %d tokens", numAffectedRows)
 	}
 
 	// Invariant to make sure that its safe to index persistedTokens
 	if len(tokensToUpsert) != len(persistedTokens) {
```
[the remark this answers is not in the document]
radazen: Ah, yeah. I guess what I said before about offsets was inaccurate: it wouldn't have been a subtle bug, cause we'd have panicked if the indexing would have been off. I'll just remove the check! It should be okay if those two values aren't the same now.

service/multichain/multichain.go
```diff
@@ -1258,15 +1326,16 @@ outer:
...
-	if err != nil {
-		return err
+	// We should receive exactly one contract from processContracts
+	if len(persistedContracts) != 1 {
+		return fmt.Errorf("expected one contract to be returned from processContracts, got %d", len(persistedContracts))
 	}
 
-	return p.processTokensForOwnersOfContract(ctx, contract.ID, ci.Chain, users, chainTokensForUsers, persistedContracts)
+	_, _, err = p.processTokensForOwnersOfContract(ctx, persistedContracts[0], users, chainTokensForUsers, postgres.TokenUpsertParams{
+		SetHolderFields: true,
```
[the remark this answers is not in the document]
radazen: Good catch!

PR #1143, opened 2023-08-11
Enable Admiring Tokens

Rohan-cp:
[GAL-3983]

### DIFF
--- a/graphql/dataloader/dataloaders.go
+++ b/graphql/dataloader/dataloaders.go
@@ -35,7 +35,6 @@
 //go:generate go run github.com/gallery-so/dataloaden FeedEventAdmiresLoader github.com/mikeydub/go-gallery/db/gen/coredb.PaginateAdmiresByFeedEventIDBatchParams []github.com/mikeydub/go-gallery/db/gen/coredb.Admire
 //go:generate go run github.com/gallery-so/dataloaden PostCommentsLoader github.com/mikeydub/go-gallery/db/gen/coredb.PaginateCommentsByPostIDBatchParams []github.com/mikeydub/go-gallery/db/gen/coredb.Comment
 //go:generate go run github.com/gallery-so/dataloaden PostAdmiresLoader github.com/mikeydub/go-gallery/db/gen/coredb.PaginateAdmiresByPostIDBatchParams []github.com/mikeydub/go-gallery/db/gen/coredb.Admire
-//go:generate go run github.com/gallery-so/dataloaden TokenAdmiresLoader github.com/mikeydub/go-gallery/db/gen/coredb.PaginateAdmiresByTokenIDBatchParams []github.com/mikeydub/go-gallery/db/gen/coredb.Admire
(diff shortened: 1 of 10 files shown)


publicapi/interaction.go
```diff
@@ -659,6 +681,62 @@ func (api InteractionAPI) AdmireFeedEvent(ctx context.Context, feedEventID persi
...
+	if !errors.As(err, &notFoundErr) {
+		return "", err
+	}
+
+	admireID, err := api.repos.AdmireRepository.CreateAdmire(ctx, "", "", tokenID, ownerID, userID)
+	if err != nil {
+		return "", err
+	}
+
+	_, err = event.DispatchEvent(ctx, db.Event{
```
jarrel-b, line 723: Let's remove this line that dispatches the event for now. I'm realizing now that this feature is bigger than I originally thought, and I want to make sure that you don't have to worry about all the other parts! I'm thinking we could split up the work as:
* saving an admire (this PR)
* retrieving all admires for a token
* sending notifications

service/persist/postgres/admire.go
```diff
@@ -38,10 +38,28 @@ func (a *AdmireRepository) CreateAdmire(ctx context.Context, feedEventID, postID
...
+
+	var ownerString sql.NullString
+	if tokenID != "" {
+		ownerString = sql.NullString{
+			String: ownerID.String(),
+			Valid:  true,
+		}
+	}
+
 	admireID, err := a.queries.CreateAdmire(ctx, db.CreateAdmireParams{
```
jarrel-b, line 57: We'll want to add a column to the `admires` table to store the token id for the admire. You can do this by creating a new migration file under `db/migrations/core` and adding the SQL statement:
```
alter table admires add column if not exists token_id varchar(255) references token(id);
```
Then, you can run `sqlc generate` in the project's home directory which update the `admires` model with the new token id field. We'll also want to update the `CreateAdmire` query to also accept `token_id` as a param. That query lives in `db/queries/core/admire.sql`. Running `sqlc generate` again will update the model with any changes made to the query.

graphql/dataloader/dataloaders.go
jarrel-b, line 1: Related to my comment above, I think it makes sense for the scope of this PR to be limited to only saving a token admire (vs. retrieving admires for a token, and also supporting notifications), so we can go over dataloader stuff when we get there!

jarrel-b:
This is an awesome start, Rohan! As mentioned in the PR, I realized that there are a few big pieces in this feature, and it makes sense if we break this up into smaller PRs to limit the scope of each. For this PR, let's only worry about saving the admires, and for future PRs we can handle retrieving admires and also supporting notifications.

To support saving token admires, one requirement is that the client will need some way of knowing if the user has already liked a particular token ID. I think you started thinking along those lines with the `loadAdmireByActorIDAndTokenID` function you added in `dataloaders.go`. Since the frontend won't use this feature anytime soon, I think we can save this for a later PR, so long as the backend already enforces that a user can only admire a token ID once.

I think we can kill two birds with one stone in a later PR by adding a field to the `Token` type that returns a connection (connection is just graphQL's name for pagination) that allows the client to paginate through all the admires for a given token
```graphql
type Token {
   # Returns an admires connection 
    admires(
        ...pagination params,
        userID, # client can optionally specify a user to determine if that user already liked a token, should return a list of length 1 if so
    ) AdmiresConnection
}
```

jarrel-b (dismissed):

PR #1153, opened 2023-08-15
coalesce thumbnail for image medias

benny-conn:
Changes:

- **Added** coalescing to thumbnail preview URLs so that images with no thumbnails can be used as well

### DIFF
--- a/db/queries/core/query.sql
+++ b/db/queries/core/query.sql
@@ -267,14 +267,14 @@ select u.* from tokens t
     where t.id = $1 and t.displayable and t.deleted = false and u.deleted = false;
 
 -- name: GetPreviewURLsByContractIdAndUserId :many
-select (tm.media->>'thumbnail_url')::varchar as thumbnail_url
+select coalesce(nullif(tm.media->>'thumbnail_url', ''), nullif(tm.media->>'media_url', ''))::varchar as thumbnail_url
     from tokens t
         inner join token_medias tm on t.token_media_id = tm.id
     where t.contract = $1
       and t.owner_user_id = $2
       and t.displayable
       and t.deleted = false
-      and tm.media ->> 'thumbnail_url' != ''
+      and (tm.media ->> 'thumbnail_url' != '' or tm.media->>'media_type' = 'image')
       and tm.deleted = false
     order by t.id limit 3;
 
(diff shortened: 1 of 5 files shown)


db/queries/core/query.sql
```diff
@@ -267,14 +267,14 @@ select u.* from tokens t
     where t.id = $1 and t.displayable and t.deleted = false and u.deleted = false;
 
 -- name: GetPreviewURLsByContractIdAndUserId :many
-select (tm.media->>'thumbnail_url')::varchar as thumbnail_url
+select coalesce(nullif(tm.media->>'thumbnail_url', ''), nullif(tm.media->>'media_url', ''))::varchar as thumbnail_url
```
radazen, line 270: Are images guaranteed to have a `media->>media_url` field? I'm wondering if we should guard against the possibility that `coalesce(nullif(tm.media->>'thumbnail_url', ''), nullif(tm.media->>'media_url', ''))` could actually return null. Worst case scenario, we could account for it in the `where` clause:

```
and (tm.media ->> 'thumbnail_url' != '' or (tm.media->>'media_type' = 'image' and tm.media->>'media_url' != '')
```

radazen (approved):
Looks great, Benny! Thanks! I searched for other queries where we check the `thumbnail_url` field, and found these:
- query.sql -> `GetGalleryTokenMediasByGalleryIDBatch`
- contract_gallery.go -> `getPreviewNFTsStmt`
- gallery.sql -> `GalleryRepoGetPreviewsForUserID`

Should we use the same pattern in all of these queries?

Thinking about it more, since we probably want this behavior anywhere we'd retrieve a preview image, maybe it's a generated column on the `token_medias` table?

```
alter table token_medias add column if not exists preview_url text
    generated always as (
        case
            when media->>'thumbnail_url' != ''
                then media->>'thumbnail_url'
            when media->>'media_type' = 'image' and media->>'media_url' != ''
                then media->>'media_url'
        end
    ) stored;
```

The only problem with that approach: it took 1m 25s to generate that column on my local machine, and the `token_medias` table is locked for the duration. Testing on dev, it took almost 4 minutes. But aside from that, it's kind of ideal, because the logic for choosing a preview url gets baked into the table and queries can just reference that column.

Maybe we just update the queries for now, and add "preview_url column" to the list of things we'd like to do during our next maintenance downtime?

PR #1154, opened 2023-08-15
Refactor pagination

jarrel-b:
Changes:
* This PR adds the pagination helper, `pageFrom`, and other helper functions to determine the set of edges to return for a page and the pageInfo for that page
* The keyset paginator now uses `pageFrom` to get the final set of edges and pageInfo instead
* The feed paginator now uses `pageFrom` instead of the keyset paginator
* Simplifies logic in getting the curated and trending feed
* Adds test related to pagination


radazen (approved):
Looks awesome! Thanks for doing this!

PR #1160, opened 2023-08-17
Enable Admire Token pt1

Rohan-cp:
Testing:
wrote the `testAdmireToken` test


db/migrations/core/00119-admire-nft.up.sql
jarrel-b, line 1: Could we rename this to `000119_admire_nft.up.sql` to stick with the naming convention of the other migrations? 

Edit: I think it should be named `000105_admire_nft.up.sql` which is the next version when writing this.

jarrel-b (commented):
Looking awesome so far 💪

graphql/schema/schema.graphql
```diff
@@ -2417,6 +2417,14 @@ type AdmirePostPayload {
 
 union AdmirePostPayloadOrError = AdmirePostPayload | ErrInvalidInput | ErrNotAuthorized
 
+type AdmireTokenPayload {
+  viewer: Viewer
+  token: Token @goField(forceResolver: true)
+  admire: Admire @goField(forceResolver: true)
+}
+
+union AdmireTokenPayloadOrError = AdmireTokenPayload | ErrInvalidInput | ErrNotAuthorized
```
jarrel-b, line 2426: If we add the check for if the token exists, can we add `ErrTokenNotFound` as a possible returned type here?

service/persist/postgres/admire.go
```diff
@@ -52,6 +52,28 @@ func (a *AdmireRepository) CreateAdmire(ctx context.Context, feedEventID, postID
 
 }
 
+func (a *AdmireRepository) CreateAdmireToken(ctx context.Context, tokenID, actorID persist.DBID) (persist.DBID, error) {
+	var tokenString sql.NullString
```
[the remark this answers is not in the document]
Rohan-cp: hmm, not sure how I would use it here because I would still need to check if tokenID != "" before doing tokenID.String()  (to then pass to `ToNullString`) right?

graphql/graphql_test.go
```diff
@@ -784,6 +785,14 @@ func testGetCommunity(t *testing.T) {
 	})
 }
 
+func testAdmireToken(t *testing.T) {
+	ctx := context.Background()
+	userF := newUserFixture(t)
+	c := authedHandlerClient(t, userF.ID)
+	alice := newUserWithTokensFixture(t)
+	admireToken(t, ctx, c, alice.TokenIDs[0])
```
Rohan-cp, line 793: not sure how to test further if `admireToken` is working because it doesn't return anything

publicapi/interaction.go
```diff
@@ -659,6 +659,58 @@ func (api InteractionAPI) AdmireFeedEvent(ctx context.Context, feedEventID persi
...
+	if err != nil {
+		return "", err
+	}
+
+	userID, err := getAuthenticatedUserID(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	/*
```
jarrel-b, line 680: I think we can remove this comment since the table constraint we added should enforce this now.

service/persist/postgres/admire.go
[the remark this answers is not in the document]
jarrel-b: Ah yea, my mistake!

jarrel-b (approved):
Looks awesome!

PR #1161, opened 2023-08-17
deleted comments and paginated replies

benny-conn:
Changes:

- **Added** `source` fields to `Comment` and `Admire` graphQL types that represent the Post or Feed Event that the admire or comment is on
- **Added** `deleted` field to `Comment` graphQL types and removed any deleted checks in the queries pertaining to comments on posts or feed events or replies (currently all of the queries, but there could be a future where we want to query for comments by a user in which case deleted should probably still be filtered)
- **Added** `replies` field to `Comment` that is the paginated replies on a comment.
- **Changed** queries for fetching/paginating comments for posts/feed events to only return top level comments (comments that are not replying to any other comment). The replies will be paginated separately in a different query and resolved in the `replies` resolver on the `Comment`.


graphql/schema/schema.graphql
```diff
@@ -792,24 +792,33 @@ union CollectionTokenByIdOrError = CollectionToken | ErrCollectionNotFound | Err
...
   dbid: DBID!
   creationTime: Time
   lastUpdated: Time
   replyTo: Comment @goField(forceResolver: true)
   commenter: GalleryUser @goField(forceResolver: true)
   comment: String
-  # should we include the feed event here?
+  replies(before: String, after: String, first: Int, last: Int): CommentsConnection
+    @goField(forceResolver: true)
+  source: CommentSource @goField(forceResolver: true)
```
benny-conn, line 818: It is worth noting that whatever this `CommentSource` is will also be explicitly typed on the `___CommentEdge` for either `Post` or `FeedEvent`. Do we still want to include the `Post` on the `PostCommentEdge` and `FeedEvent` on the `FeedEventCommentEdge` if that can be determined from the `Comment` itself?

db/migrations/core/000106_comments_deletable.up.sql
```diff
@@ -0,0 +1,6 @@
+drop index if exists comments_created_at_id_feed_event_id_idx;
+CREATE UNIQUE INDEX IF NOT EXISTS comments_created_at_id_feed_event_id_idx ON comments (created_at desc, id desc, feed_event_id);
+drop index if exists comments_created_at_id_post_id_idx;
+CREATE UNIQUE INDEX IF NOT EXISTS comments_created_at_id_post_id_idx ON comments (created_at desc, id desc, post_id);
```
benny-conn, line 4: The current index for this wrongly uses `feed_event_id` instead of `post_id` so this is a double whammy change!

radazen (approved):
Looks good to me! The only thing I'm iffy on is the use of deleted comments. I think `deleted` in our database should mean "this has been actually deleted and is no longer available to use, but it's here as an audit trail if we need to see what happened or need to restore something by hand." Imho, we should be able to hard delete everything in the database that has `deleted = true` and the backend should still function the same as it ever did. If hard deleting our soft-deleted rows has a functional impact on the site (in this case, replies), then I think it's worth considering a different approach.

In this case: can we give comments a new column to indicate that the comment was removed? If we still want to reference the comment's ID, the user who posted it, etc -- even if we _don't_ want to reference the actual "comment" string itself -- then I think the comment should still be a regular, not-deleted row in the database. You could even keep the "deleted" field in the GraphQL schema, since it's fine for the frontend to think of the comment as "deleted." I just don't think it should be marked as deleted on the backend if we're still actively using it for stuff.

graphql/schema/schema.graphql
benny-conn: Great! I'll check with kaito if that is going to be a breaking change for the FE.

PR #1165, opened 2023-08-21
Cursor refactor

jarrel-b:
This PR factors out cursor logic from the paginators and adds a cursor interface that cursor types can implement:
```go
type cursorer interface {
	Pack() (string, error) // Encodes the cursor as a string
	Unpack(string) error // Decodes the cursor into the receiver
}
```
This should pave the path for comparing cursors and faster cursor searches. At the moment, cursor equality is handled by comparing the cursor strings instead of the structs themselves. Most cursors do support this besides the more complex cursors that contain slices since go doesn't support slice comparisons out of the box. I figure I can save this for when we implement sorting + binary search.

### DIFF
--- a/publicapi/pagination.go
+++ b/publicapi/pagination.go
@@ -237,8 +237,14 @@ type timeIDPagingParams struct {
 
 func (p *timeIDPaginator) paginate(before *string, after *string, first *int, last *int) ([]any, PageInfo, error) {
 	queryFunc := func(limit int32, pagingForward bool) ([]any, error) {
-		beforeCur := timeIDCursor{Time: defaultCursorBeforeTime, ID: defaultCursorBeforeID}
-		afterCur := timeIDCursor{Time: defaultCursorAfterTime, ID: defaultCursorAfterID}
+		beforeCur := timeIDCursor{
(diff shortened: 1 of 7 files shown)


publicapi/pagination.go
```diff
@@ -1115,3 +874,347 @@ func (d *cursorDecoder) readUInt64() (uint64, error) {
...
+	return func(node any) (c cursorer, err error) {
+		var cur timeIDCursor
+		cur.Time, cur.ID, err = f(node)
+		return &cur, err
+	}
+}
+
+func (curs) NewBoolBoolIntIDCursorer(f func(any) (bool, bool, int64, persist.DBID, error)) cursorable {
+	return func(node any) (c cursorer, err error) {
+		var cur boolBootIntIDCursor
```
radazen, line 908: Small (but amusing) typo: `boolBoot`

radazen (approved):
Looks awesome, 10/10, A+++++++, would review again

publicapi/pagination.go
jarrel-b: haha how did I not see that

PR #1169, opened 2023-08-22
mention data

benny-conn:
Changes:

- **Added** mentions to the `comments` an `posts` tables in the db which will be JSONB for the mention data
- **Added** resolvers for mentions

In the way this is implemented, a `Mention` is either a user or a community and then there is an optional `index` (optional just in case we want to allow mentions in the future that are like tags on a picture instead of direct text mentions) that represents the start of where the mention is and its length. This works fine but fails in a few ways:

1. A user edits the comment (we don't have this built out but just theorizing if we wanted to edit comments) and the indexes need to get shifted and mentions the same.
2. A user changes their username, the mention is still able to reference the new user but there will be a mismatch between the real username and what the mention looks like to the user. A user could hover on a mention and it might be confusing because the mention looks different than what shows up in the hover popup.

An alternative method that has its own different problems is to interpolate the mentions by some key. For example:

1. Frontend gets comment after user searches for user while typing the mention: `Love it, @kaito` 
2. Frontend or backend replaces the mention with the DBID or some generated ID: `Love it, @{2aY8sb...}`
3. Frontend requests comments to be rendered, backend sends to frontend something that can be used to turn use that ID and get the actual entity it points to, community or user
4. When frontend renders the comment, each `@{id}` is replaced with the data (username or community name) from the resolved entity that is being mentioned.

Problems with this method:

1. This method requires that the frontend resolves the mentions to properly render the comment. In the indexing method, the comment look acceptable without resolving any mentions (although the mentions might not be bolded) but in this method the comment will look like a hacker wrote it unless the frontend resolves the mentions and fills in the comment with data from the user or community.
2. I feel like this method is slightly less "typed" because some of the important mention information is stored on the comment itself and the comment is just a string. In the indexing method, the comment doesn't know or care about mentions, it is the mention's responsibility only to understand the comment. The roles are reversed here.
3. What if a user really wants to include `@{2aY8sb...}` raw in their comment because that random string somehow means something to them but we keep replacing it with a gallery user's username :(

But the other problems are solved because now a comment can be edited and the comment can still be interpolated correctly because indexes are not involved and mentions can even be removed from the comment with the mention data being unchanged. Obviously the username changing is not an issue because the username will come from resolving the current user by ID.

### DIFF
--- a/db/gen/coredb/batch.go
+++ b/db/gen/coredb/batch.go
@@ -868,7 +868,7 @@ func (b *GetCollectionsByGalleryIdBatchBatchResults) Close() error {
 }
 
 const getCommentByCommentIDBatch = `-- name: GetCommentByCommentIDBatch :batchone
-SELECT id, version, feed_event_id, actor_id, reply_to, comment, deleted, created_at, last_updated, post_id, removed FROM comments WHERE id = $1 and deleted = false
+SELECT id, version, feed_event_id, actor_id, reply_to, comment, deleted, created_at, last_updated, post_id, removed, mentions FROM comments WHERE id = $1 and deleted = false
 `
 
 type GetCommentByCommentIDBatchBatchResults struct {
@@ -912,6 +912,7 @@ func (b *GetCommentByCommentIDBatchBatchResults) QueryRow(f func(int, Comment, e
 			&i.LastUpdated,
 			&i.PostID,
 			&i.Removed,
+			&i.Mentions,
 		)
 		if f != nil {
 			f(t, i, err)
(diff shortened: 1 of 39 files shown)


kaitoo1:
thanks for the detailed explanation @benny-conn !

my thoughts

I agree with the downsides you mentioned for the alternative method, and I think those are good enough reasons to decide against that path.

So for the way you have it currently implemented, addressing the issues you mentioned:

1. Edited comments
if/when we allow editing comments, I'd imagine it would be pretty straightforward to update the indices when the user saves? similar to how we figure out the index when we first create a comment. since the user could delete or add mentions when they edit we probably need to update everything anyway.


2. if the user changes their username
Whether or not we displayed the new username in the comment definitely has pros / cons from a user perspective so I'll chat with the team to decide. In the meantime though, ideally we set this up in a way where we have the information we need for the frontend to do either, which gives us more flexibility to decide one way or another down the line. It sounds like you've done this which is great!

want to make sure I understand what you have in mind:
The mentions table contains the following
- comment ID
- mentioned userID
- index of mention
( not sure if you're planning to have a new row per mention even if there are multiple mentions per comment, either way this should work)

The index allows the frontend to know exactly where in the comment each mention starts, without having to know what the old username was. 
The comment text itself in the database would have the old username, so the frontend could either 
a. display this original text, and simply display the mention as a link to the user
b. dynamically replace the mentioned username with the user's new username. 

To do b. I think we'd need not just the start position but also the end position of the mention, so that the frontend know exactly what text to replace without needing to know the old username. for usernames it could theoretically just assume that it needs to replace anything following "@" until the next blank space, but I imagine this wouldn't work for Community Names that have spaces.

PR #1170, opened 2023-08-23
Benny/mention data new

benny-conn:
Changes:

- **Added** mentions to the comments an posts tables in the db which will be JSONB for the mention data
- **Added** resolvers for mentions

Better than other PR because the commits are fixed


db/migrations/core/000106_comments_deletable.up.sql
```diff
@@ -5,4 +5,7 @@ CREATE UNIQUE INDEX IF NOT EXISTS comments_created_at_id_post_id_idx ON comments
 
 alter table comments add column removed bool default false not null;
 
+alter table comments add column mentions jsonb default '{}'::jsonb not null;
```
radazen, line 8: Thinking out loud: should we prefer a `mentions` table to a `jsonb` column?

Benefits of jsonb: it's inline, you don't have to make sure to insert mentions before inserting comments, you don't have to do a separate lookup to get a comment or post's mentions

Benefits of a table: data has a defined schema, can create one row per mention instead of an inline array, more SQL-ish way to model data

I think the drawbacks of a table can be mitigated with a CTE; you know the ID of the comment/post, so you can just insert the comment in a CTE and then reference it in the mentions (I think!). There's still overhead when looking up mentions for a comment, though we could mitigate those with some planned refactoring to dataloaders.

The biggest question for me: might we someday want a feed of mentions for a user? Like, maybe a thing I click to see a chronological list of things I've been mentioned in? And if so, does our storage format in `jsonb` allow for indexing that by user? I know jsonb allows indexing in the first place, but does it allow us to index a mention by user? Seems like it's not a given, since the data could be something like an array of mentions, and I'm not sure if it's feasible to index a field of each element of that array.

cc @kaitoo1 @Robinnnnn, since some of this is more of a product question! If we don't see ourselves wanting to look up mentions-by-user in the future, then there's probably no reason to change anything here.

radazen (approved):
🔥 Added some questions/comments, but everything looks good overall! I'm super excited to see this in action. 📣 🗣️

PR #1171, opened 2023-08-23
Enable Admire Token pt.2

Rohan-cp:


publicapi/interaction.go
```diff
@@ -677,6 +762,15 @@ func (api InteractionAPI) AdmireToken(ctx context.Context, tokenID persist.DBID)
 		return "", err
 	}
 
+	admire, err := api.GetAdmireByActorIDAndTokenID(ctx, userID, tokenID)
```
jarrel-b, line 765: I think it'd be awesome if the admire token operation was idempotent, such that if an admire already exists, we just return the existing admire. That way, we save a call to the database to check if an admire exists which means we don't need the dataloader and the query to fetch a specific admire anymore. It also means we don't need to return an error when an admire already exists.

We can update the existing query to something like this (haven't actually tested if this is valid SQL, but should be similar):
```
insert into admires (id, feed_event_id, post_id, token_id, actor_id) values ($1, sqlc.narg('feed_event'), sqlc.narg('post'), sqlc.narg('token'), $2)
on conflict (token_id, actor_id) do nothing
returning id;
```

The change is the `on conflict...` for the specific constraint (token_id, actor_id) that is being violated. I'd double-check if `returning id` returns the existing id, but I'm pretty sure it does!

It'd actually be awesome to do this for all the other things we can admire (post_id, feed_event_id), but feel free to skip unless its easy to also do!

db/queries/core/query.sql
```diff
@@ -544,6 +544,16 @@ SELECT * FROM admires WHERE post_id = sqlc.arg('post_id') AND deleted = false
 -- name: CountAdmiresByPostIDBatch :batchone
 SELECT count(*) FROM admires WHERE post_id = $1 AND deleted = false;
 
+-- name: PaginateAdmiresByTokenIDBatch :batchmany
```
jarrel-b, line 547: We can actually combine `PaginateAdmiresByTokenIDBatch` and `GetAdmireByActorIDAndTokenID` into a single query! We use this pattern in a few places where an optional filter is included based on a flag, like here for example: https://github.com/gallery-so/go-gallery/blob/9a38560fc24ad644a960bed30ebb1ac73579103d/db/queries/core/query.sql#L624

We can add a line to the where clause that's something like:
```
where (not @only_for_actor::bool or actor_id = @actor_id)
```
so when `only_for_actor` is `false` we get admires from every user, but when `true` the second part of the check is executed resulting in admires for a specific `actor_id`.  This should help simplify the resolver logic as well!

graphql/dataloader/dataloaders.go
jarrel-b, line 1: Nice job figuring out how to add a dataloader!

jarrel-b (commented):
Looking awesome so far @Rohan-cp ! Just had a few comments and suggestions.

publicapi/interaction.go
```diff
@@ -430,6 +430,73 @@ func (api InteractionAPI) PaginateCommentsByFeedEventID(ctx context.Context, fee
...
+		"tokenID": validate.WithTag(tokenID, "required"),
+	}); err != nil {
+		return nil, PageInfo{}, err
+	}
+
+	if err := validatePaginationParams(api.validator, first, last); err != nil {
+		return nil, PageInfo{}, err
+	}
+
+	onlyForActor := *userID != ""
```
jarrel-b, line 446: I think this will lead to a runtime error if `userID` is `nil`, example: https://play.golang.com/p/1EdlZ59E-7X. The way you can get around this is something like:
```
var onlyForActor bool

if userID != nil {
    onlyForActor = true
}
```

publicapi/interaction.go
```diff
@@ -430,6 +430,73 @@ func (api InteractionAPI) PaginateCommentsByFeedEventID(ctx context.Context, fee
...
+	queryFunc := func(params timeIDPagingParams) ([]interface{}, error) {
+		admires, err := api.loaders.AdmiresByTokenID.Load(db.PaginateAdmiresByTokenIDBatchParams{
+			TokenID:       tokenID,
+			Limit:         params.Limit,
+			OnlyForActor:  onlyForActor,
+			ActorID: 	   *userID,
```
jarrel-b, line 452: This will also cause a runtime error if you dereference `userID` when its `nil`!

graphql/resolver/schema.resolvers.helpers.go
```diff
@@ -171,6 +171,12 @@ func errorToGraphqlType(ctx context.Context, err error, gqlTypeName string) (gql
 		mappedErr = model.ErrAddressOwnedByUser{Message: message}
 	case persist.ErrAdmireNotFound:
 		mappedErr = model.ErrAdmireNotFound{Message: message}
+	case persist.ErrAdmireFeedEventNotFound:
+		mappedErr = model.ErrAdmireFeedEventNotFound{Message: message}
```
jarrel-b, line 175: Could we map all of these errors to `ErrAdmireNotFound` instead? I think it's helpful to distinguish between the different errors in the backend, but from the POV of the client, it's not that useful to distinguish between the three. A neat go trick is that you can map all three to the same error in one line:
```
case persist.ErrAdmireNotFound,persist.ErrAdmirePostNotFound,persist.ErrAdmireTokenNotFound:
	mappedErr = model.ErrAdmireNotFound{Message: message}
```

graphql/schema/schema.graphql
```diff
@@ -1634,6 +1654,18 @@ type ErrAdmireAlreadyExists implements Error {
   message: String!
 }
 
+type ErrAdmireFeedEventNotFound implements Error {
```
jarrel-b, line 1657: I don't think we'll need these errors anymore if we map everything to `ErrAdmireNotFound`!

jarrel-b (approved):
Looks awesome @Rohan-cp, left a few more comments. Feel free to merge when you're ready 👌

graphql/resolver/schema.resolvers.helpers.go
Rohan-cp: that makes a lot of sense! thanks

PR #1175, opened 2023-08-25
Handle feed deletes

jarrel-b:
Fixes the issue where deleted posts prevent retrieving the trending and curated feed. Deleted posts are possible because a post can be deleted in between refreshes of the materialized feed view. This PR refactors `loadFeedEntities` to handle deleted feed entities.

### DIFF
--- a/publicapi/feed.go
+++ b/publicapi/feed.go
@@ -6,6 +6,7 @@ import (
 	"database/sql"
 	"fmt"
 	"math"
+	"sort"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


publicapi/feed.go
```diff
@@ -647,73 +649,84 @@ func loadFeedEntities(ctx context.Context, d *dataloader.Loaders, typs []persist
...
 
-			for j, id := range ids {
-				if it, ok := idsToFeedPosts[id]; ok {
-					entities[j] = it
-				}
-			}
-		case err := <-incomingErrors:
-			return nil, err
+	// Sort in descending order
+	sort.Slice(errored, func(i, j int) bool { return errored[i] > errored[j] })
```
jarrel-b, line 717: Worse case `len(errored) = len(ids)` but that should rarely be the case

Robinnnnn (approved):

PR #1176, opened 2023-08-28
Add email signup

jarrel-b:
### Auth flow
image

### Schema changes
The login and create payloads can now return the errors `ErrEmailUnverified`, and `ErrEmailAlreadyUsed` respectively. The former is to notify the client to resend a verification email, and the latter is to notify the client to prompt the user to provide a different email or to sign-in with the original email.
```graphql
union LoginPayloadOrError =
    LoginPayload
  | # ... existing errors
  | ErrEmailUnverified # new error type

union CreateUserPayloadOrError =
    CreateUserPayload
  | # ... existing errors
  | ErrEmailAlreadyUsed # new error type
```

### Primary Wallets
Since a user can now sign up with an email, it's now possible for a user to have no wallets and no primary wallet. However, wallet removal remains the same where a user can't delete their primary wallet nor can they delete all of their wallets. ~~When adding a wallet, if they don't have a primary wallet already set, do we want to set the primary wallet for them automatically @kaitoo1?~~ (Update: when a user adds a wallet and no primary wallet is set, we'll set that wallet as their primary for them.) For integration, the `primaryWallet` field is already optional in the schema:
```graphql
type GalleryUser {
  primaryWallet: Wallet @goField(forceResolver: true)
}
```
... but the frontend will need to handle a null wallet correctly since it throws an error if it is missing (cc @kaitoo1)

### Email notifications and Newsletter subscriptions
When a user signs up by email, they are also subscribed to email notifications and the newsletter automatically. Subscription happens asynchronously via a queue so that we can reprocess failed subscriptions. ~~We also discussed handling all email verification through MagicLink. This would require the frontend to send the MagicLink email instead of the backend sending the Sendgrid email. Should I add this now, or scope it for later (@Robinnnnn, @kaitoo1)? I can add this pretty quickly, just wasn't sure if this was a priority since the lift is mostly on the frontend.~~ (Update: slated for later, added a ticket https://linear.app/galleryso/issue/GAL-4163/consolidate-email-verification-to-one-provider)


publicapi/user.go
```diff
@@ -1655,3 +1664,65 @@ func standardizeURI(u string) string {
...
 }
+
+func createNewUserParamsWithAuth(ctx context.Context, authenticator auth.Authenticator, username string, bio string, email *persist.Email) (persist.CreateUserInput, error) {
+	authResult, err := authenticator.Authenticate(ctx)
+	if err != nil && !util.ErrorAs[persist.ErrUserNotFound](err) {
+		return persist.CreateUserInput{}, auth.ErrAuthenticationFailed{WrappedErr: err}
+	}
+
+	if authResult.User != nil && !authResult.User.Universal.Bool() {
+		if _, ok := authenticator.(auth.MagicLinkAuthenticator); ok {
```
jarrel-b, line 1675: I wasn't super happy with having the callers check what authenticator was passed and also catching the specific errors depending on the context since `ErrUserNotFound` is an error for login, but not for signup. I was originally thinking of extending the authenticator interface with something like:
```go
type Authenticator interface {
    Authenticate()
    AuthorizedForLogin()
    AuthorizedForSignup()
}
```
so that it's up to the authenticators to figure that out, but this seems to me for now!

service/persist/postgres/user.go
```diff
@@ -254,30 +255,30 @@ func (u *UserRepository) Create(pCtx context.Context, pUser persist.CreateUserIn
...
-
-	userID := persist.GenerateID()
-	err = queries.InsertUser(pCtx, db.InsertUserParams{
-		ID:                   userID,
+	userID, err := queries.InsertUser(pCtx, db.InsertUserParams{
+		ID:                   persist.GenerateID(),
 		Username:             util.ToNullString(pUser.Username, true),
 		UsernameIdempotent:   util.ToNullString(strings.ToLower(pUser.Username), true),
 		Bio:                  util.ToNullString(pUser.Bio, true),
-		Wallets:              persist.WalletList{persist.Wallet{ID: walletID}},
```
jarrel-b, line 268: Setting the `wallets` and `primary_wallet_id` columns is now handled by `createWalletWithTx`.

service/auth/auth.go
```diff
@@ -82,6 +82,11 @@ var ErrSignatureInvalid = errors.New("signature invalid")
 
 var ErrInvalidMagicLink = errors.New("invalid magic link")
 
+// TODO: Figure out a better scheme for handling user-facing errors
+var ErrEmailUnverified = errors.New("The email address you provided is unverified. Login with QR code instead, or verify your email at gallery.so/settings.")
```
radazen, line 86: I think this error is part of why we _do_ have authenticators throwing errors when they probably shouldn't be. Easier to do Magic Link specific stuff when you know you're in the Magic Link authenticator, but in a perfect world, we'd separate the two and have any "email login" method:
- be aware that the user was attempting to log in via email, and
- throw an appropriate error if the authenticator returns a user whose current email address in the database is unverified

emails/handler.go
```diff
@@ -17,6 +17,7 @@ func handlersInitServer(router *gin.Engine, loaders *dataloader.Loaders, queries
 	sendGroup := router.Group("/send")
 
 	sendGroup.POST("/notifications", middleware.AdminRequired(), adminSendNotificationEmail(queries, s))
+	sendGroup.POST("/process/add-to-mailing-list", processAddToMailingList(queries))
```
radazen, line 20: Might be cool to start including the `BasicHeaderAuthRequired` middleware on internal tasks like this one, since nothing would currently stop external callers from signing users up for our mailing list.

radazen (approved):
Looks great! I left some comments about authenticators, but I don't think an authenticator refactor needs to block this PR!

publicapi/user.go
jarrel-b: Ah this is a tidy way of thinking about it! So an authenticator solely handles authentication, and it's up to the caller to decide if a user is authorized to access a resource / do something based on that info.

emails/handler.go
jarrel-b: good call!

PR #1177, opened 2023-08-30
Benny/incremental sync

benny-conn:
Changes:

- **Added** new `GetTokensIncrementallyByWalletAddress` interface func for alchemy and opensea (for now) that will be used by the new `SyncTokensIncrementallyByUserID` function.
- **Changed** merging logic to take in the existing tokens as an input so that we can keep track of the existing tokens without requerying the DB every time

TODO:

- [x] more providers
- [x] tests


service/multichain/multichain.go
```diff
@@ -166,10 +172,16 @@ type Verifier interface {
 
 // TokensOwnerFetcher supports fetching tokens for syncing
 type TokensOwnerFetcher interface {
-	GetTokensByWalletAddress(ctx context.Context, address persist.Address, limit int, offset int) ([]ChainAgnosticToken, []ChainAgnosticContract, error)
+	GetTokensByWalletAddress(ctx context.Context, address persist.Address) ([]ChainAgnosticToken, []ChainAgnosticContract, error)
```
benny-conn, line 175: We were not ever using limit or offset for this and we talked about removing it and felt now was a good time since I was already adding to every provider.

service/multichain/fallback.go
```diff
@@ -136,16 +143,70 @@ func (f SyncFailureFallbackProvider) GetBlockchainInfo() BlockchainInfo {
...
+func getTokensIncrementallyByWalletAddressWithFallback(ctx context.Context, address persist.Address, primary TokensIncrementalOwnerFetcher, fallback any, rec chan<- ChainAgnosticTokensAndContracts, errChan chan<- error, processTokens func(context.Context, []ChainAgnosticToken) []ChainAgnosticToken, processContracts func(context.Context, []ChainAgnosticContract) []ChainAgnosticContract) {
+	defer close(rec)
+	// create sub channels so that we can separately keep track of the results and errors and handle them here as opposed to the original receiving channels
+	subErrChan := make(chan error)
+	subRec := make(chan ChainAgnosticTokensAndContracts)
+	go primary.GetTokensIncrementallyByWalletAddress(ctx, address, subRec, subErrChan)
+	for {
+		select {
+		case err := <-subErrChan:
+			if tiof, ok := fallback.(TokensIncrementalOwnerFetcher); ok {
```
jarrel-b, line 169: Rather than flipping to incrementally syncing from `fallback`  on the first from primary error, would it make sense to continue to process incrementally from the primary, but if we fail to get a token, try to fetch it from the fallback (not sure the exact interface we have for fetching a single token) and push the result to the `rec` channel?  I'm thinking in the worst case, we incrementally sync all of a user's tokens except the last one, at which point we revert to the fallback and start the sync process all over.

service/multichain/fallback.go
```diff
@@ -136,16 +143,70 @@ func (f SyncFailureFallbackProvider) GetBlockchainInfo() BlockchainInfo {
...
+						}
+						if processContracts != nil {
+							tokens.Contracts = processContracts(ctx, tokens.Contracts)
+						}
+						rec <- tokens
+					}
+				}
+			}
+			errChan <- err
+			return
```
jarrel-b, line 194: Not sure if we want to return here and skip receiving other tokens from the primary.

service/multichain/multichain.go
```diff
@@ -334,6 +346,80 @@ func (p *Provider) SyncTokensByUserID(ctx context.Context, userID persist.DBID,
...
+								}
+								result <- chainTokensAndContracts{
+									tokens:    chainTokens{chain: chain, tokens: ts.Tokens, priority: priority},
+									contracts: chainContracts{chain: chain, contracts: ts.Contracts, priority: priority},
+								}
+
+							case err := <-errs:
+								logger.For(ctx).Errorf("error while syncing tokens for user %s: %s (provider: %d (%T))", user.Username, err, priority, fetcher)
+								errChan <- err
+								return
```
jarrel-b, line 405: Also not sure if we want to return here and skip receiving other tokens from the fetcher, or continue to get the other pages from it.

jarrel-b (commented):
Looks awesome Benny, I had a few design questions but looks great so far!

graphql/schema/schema.graphql
```diff
@@ -2518,7 +2518,7 @@ type Mutation {
   updateTokenInfo(input: UpdateTokenInfoInput!): UpdateTokenInfoPayloadOrError @authRequired
   setSpamPreference(input: SetSpamPreferenceInput!): SetSpamPreferencePayloadOrError @authRequired
 
-  syncTokens(chains: [Chain!]): SyncTokensPayloadOrError @authRequired
+  syncTokens(chains: [Chain!], incrementally: Boolean): SyncTokensPayloadOrError @authRequired
   syncCreatedTokensForNewContracts(
```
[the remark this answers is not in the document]
benny-conn: Yep, to start we wanted to have this be an optional parameter and if all goes well we make it the default.

service/persist/token_gallery.go
```diff
@@ -51,6 +51,8 @@ type TokenGallery struct {
 	BlockNumber          BlockNumber `json:"block_number"`
 	IsUserMarkedSpam     *bool       `json:"is_user_marked_spam"`
 	IsProviderMarkedSpam *bool       `json:"is_provider_marked_spam"`
+
+	Priority *int `json:"-"`
```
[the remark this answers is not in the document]
benny-conn: Yes, this field is only used in memory to keep track of the tokens priorities over the course of an incremental sync. If we store these tokens in memory with their priorities, when we come across the same tokens later, we can compare priorities to determine what to do with the token dedupe.

service/persist/postgres/contract_gallery.go
```diff
@@ -40,6 +41,9 @@ func NewContractGalleryRepository(db *sql.DB, queries *db.Queries) *ContractGall
 	getByAddressesStmt, err := db.PrepareContext(ctx, `SELECT ID,VERSION,CREATED_AT,LAST_UPDATED,ADDRESS,SYMBOL,NAME,OWNER_ADDRESS,CHAIN,IS_PROVIDER_MARKED_SPAM FROM contracts WHERE ADDRESS = ANY($1) AND CHAIN = $2 AND DELETED = false;`)
 	checkNoErr(err)
 
+	getByTokenIDsStmt, err := db.PrepareContext(ctx, `select contracts.ID,contracts.VERSION,contracts.CREATED_AT,contracts.LAST_UPDATED,contracts.ADDRESS,contracts.SYMBOL,contracts.NAME,contracts.OWNER_ADDRESS,contracts.CHAIN,contracts.IS_PROVIDER_MARKED_SPAM from contracts join tokens on contracts.id = tokens.contract where tokens.id = any($1) and contracts.deleted = false;`)
```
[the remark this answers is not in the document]
benny-conn: Good idea!

service/persist/contract_gallery.go
```diff
@@ -27,6 +27,8 @@ type ContractGallery struct {
 	BadgeURL              NullString `json:"badge_url"`
 	IsProviderMarkedSpam  bool       `json:"is_provider_marked_spam"`
 	OverrideCreatorUserID DBID       `json:"override_creator_user_id"`
+
+	Priority *int `json:"-"`
```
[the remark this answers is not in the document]
benny-conn: Same as above

service/multichain/multichain.go
```diff
@@ -1803,38 +1988,47 @@ type contractMetadata struct {
 	ProfileImageURL string
 	Description     string
 	IsSpam          bool
+	priority        *int
```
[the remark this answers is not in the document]
benny-conn: Same as above

service/multichain/multichain.go
```diff
@@ -484,6 +575,65 @@ outer:
...
+		select {
+		case inc, ok := <-result:
+			if !ok {
+				break outer
+			}
+			newContracts, err := p.processContracts(ctx, []chainContracts{inc.contracts}, currentContracts, false)
+			if err != nil {
+				return err
+			}
+			currentContracts = append(currentContracts, newContracts...)
```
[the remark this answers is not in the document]
benny-conn: Modified both `processContracts` and `processTokensForUsers` to return the whole current state of tokens for the user not just the tokens that were persisted in the batch.

service/multichain/fallback.go
benny-conn: Anytime an err is sent to the errChan we expect the flow to stop there, this channel represents an an error incrementally syncing, not necessarily just an error for a single token. It will stop execution further up in the multichain provider anyway so we might as well return.

service/multichain/multichain.go
benny-conn: Same as above.

service/multichain/fallback.go
benny-conn: This would be ideal but we would require the primary provider to include some detail on which token failed (if it is in fact just a single token) or the group of tokens that failed given that each send on the receiving channel is actually a batch of tokens and contracts. I feel like because of the generic nature of the primary and fallback, we can't expect them to return such a detailed error and have the fallback be capable of taking that error and processing just the tokens involved.

PR #1182, opened 2023-09-12
Add retries to syncs and improve processing state

jarrel-b:
### Changes
* Tokens that failed during syncs are now individually retried
* More accurate syncing and invalid states of token media
* Fixes small bug on refreshes where token descriptors failed to update
* Removes some code duplication in [cloudtask.go](https://github.com/gallery-so/go-gallery/pull/1182/files#diff-0f94b92bc53e282559ef1a27907f395daf7ccd3e46cdf815d8a18cd42c75f273)

### How does it work?
This PR introduces `tokenmanage`, which manages the processing lifecycle of a token. Rather than submitting tokens directly to Cloud Tasks, tokens are now sent to `tokenmanage`, which maintains token state in Redis and is responsible for submitting tokens for processing. Before a worker starts working on a token, it must first ask `tokenmanage` for a lock. If a lock is acquired, `tokenmanage` continually updates the TTL for the token's key to indicate the token is processing. When the worker is done, it calls a callback so that `tokenmanage` can mark the token as finished and release resources on the token.

If an error occurs while processing, `tokenmanage` will try to re-enqueue the token up to a configured `maxRetries`. The rate at which a token is retried is configured using the `limiters.KeyRateLimiter` which buffers the rate a token is processed in Cloud Tasks. Currently, this is only enabled for syncing (regular syncs + optimistic syncs) and not for refreshes. For retries, the pipeline also now always refreshes metadata, since metadata commonly changes shortly after minting.

The media resolver now consults with `tokenmanage` to figure out if a token is still in flight. If `tokenmanage` says that the token is processing, the resolver sets the media type to `Syncing` (instead of `Invalid` or `Unknown`). This lookup only occurs for tokens with no media or invalid media so that a call to Redis isn't made for every media query.

### Follow-ups
We made `chains` an optional field for tokenprocessing, after this PR is merged I'll remove it to simplify logic in a few places.

### DIFF
--- a/db/queries/core/query.sql
+++ b/db/queries/core/query.sql
@@ -1366,7 +1366,8 @@ update tokens
         description = @description,
         last_updated = now()
     where token_id = @token_id
-      and contract = (select id from contracts where address = @contract_address)
+      and contract = @contract_id
+      and chain = @chain
       and deleted = false;
 
 -- name: GetTopCollectionsForCommunity :many
--- a/graphql/resolver/schema.resolvers.go
+++ b/graphql/resolver/schema.resolvers.go
@@ -8,8 +8,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"strings"
-	"time"
 
 	"github.com/99designs/gqlgen/graphql"
 	"github.com/mikeydub/go-gallery/db/gen/coredb"
@@ -21,7 +19,6 @@ import (
 	"github.com/mikeydub/go-gallery/service/logger"
 	"github.com/mikeydub/go-gallery/service/mediamapper"
 	"github.com/mikeydub/go-gallery/service/persist"
-	"github.com/mikeydub/go-gallery/service/rpc/ipfs"
 	sentryutil "github.com/mikeydub/go-gallery/service/sentry"
 	"github.com/mikeydub/go-gallery/util"
 	"github.com/mikeydub/go-gallery/validate"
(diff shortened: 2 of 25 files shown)


db/queries/core/query.sql
```diff
@@ -1366,7 +1366,8 @@ update tokens
         description = @description,
         last_updated = now()
     where token_id = @token_id
-      and contract = (select id from contracts where address = @contract_address)
```
jarrel-b, line 1369: I noticed during testing that refreshes were failing for some contracts because the subquery returned more than one row. I updated the query to require a `contract_id` instead.

graphql/resolver/schema.resolvers.go
```diff
@@ -2580,33 +2577,12 @@ func (r *tokenResolver) Media(ctx context.Context, obj *model.Token) (model.Medi
 	}
 
 	tokenMedia, err := publicapi.For(ctx).Token.MediaByTokenID(ctx, obj.Dbid)
-	if err != nil || !tokenMedia.Active {
```
jarrel-b, line 2583: Most of this logic was factored out and moved to [schema.resolvers.helpers.go](https://github.com/gallery-so/go-gallery/pull/1182/files#diff-9832c54fe698f6fc3c6bba74974108fddaf27f4858708108a22d626b8edccf24).

service/multichain/multichain.go
```diff
@@ -1253,27 +1246,26 @@ func (p *Provider) RefreshToken(ctx context.Context, ti persist.TokenIdentifiers
...
-		Name:            util.ToNullString(finalTokenDescriptors.Name, true),
-		Description:     util.ToNullString(finalTokenDescriptors.Description, true),
-		TokenID:         ti.TokenID,
-		ContractAddress: ti.ContractAddress,
-	}); err != nil {
-		return err
-	}
-
-	if err := p.Repos.ContractRepository.UpsertByAddress(ctx, ti.ContractAddress, ti.Chain, persist.ContractGallery{
+	contractID, err := p.Repos.ContractRepository.UpsertByAddress(ctx, ti.ContractAddress, ti.Chain, persist.ContractGallery{
```
jarrel-b, line 1249: Swapped the order of this around to first update the contract, and then update the token.

service/task/cloudtask.go
jarrel-b, line 1: Most of the changes to this file are removing some code duplication.

Robinnnnn (approved):

benny-conn (commented):
Looks beautiful and is mega thorough!

PR #1184, opened 2023-09-14
Share to Gallery

jarrel-b:
[Share To Gallery Flow](https://app.excalidraw.com/s/12gEPcX20Oi/A0nqrlpPASv)
### GraphQL changes
Added mutations and queries:
```graphql
type query {
  # Provides details to populate the post composer, should be polled after calling `referralPostPreflight`
  postComposerDraftDetails(
    input: PostComposerDraftDetailsInput!
  ): PostComposerDraftDetailsPayloadOrError
}
type mutation {
  # Should be called as soon as possible so the backend can start gathering information and processing media for the token
  referralPostPreflight(input: ReferralPostPreflightInput!): ReferralPostPreflightPayloadOrError
  # Post a token from a referral site
  referralPostToken(input: ReferralPostTokenInput!): ReferralPostTokenPayloadOrError @authRequired
}
```
When the backend receives a `referralPostPreflight` query, it places the requested token in the queue. There are two levels of retries: the first is a shorter retry period that occurs during the request. If we still can't find a token, the token is placed back on the queue for retry with longer delays in between each attempt. When we do find the token, we start processing media for it (which has its own retry logic).

Adds a new GraphQL scalar type: `TokenId` which represents a token's ID in the contract. When read by the backend, the value is converted to a hex string which is how we usually represent token IDs in the backend. When it is returned to the client, it is also converted to a hex string prefixed with `0x` (regardless of the original input). When the input string doesn't begin with `0x`, it's assumed to be in decimal.

I also updated how we map backend errors to GraphQL types. Instead of using a type `switch` statement, it now uses `errors.Is` and `errors.As`. This is because `errors.Is` and `errors.As` both follow the error chain in order to check if an error is of the given type. When we search for a token, we wrap the check function in a retry. If we use all the retries, the last errors get wrapped in a retry error. It's useful though to map the original error to the GraphQL type hence the refactor.

### Tokenprocessing changes
The pipeline now supports running from token identifiers instead of from a token DBID. This allows media to be processed even if there's no user with the token. If a token DBID is provided it's used as a fallback for when fetching metadata fails. When we migrate the tokens table, I think it makes sense to update the pipeline to accept IDs from the new token entities instead. This makes sense for the `tokenmanage` system too, since it's more useful to know when a token is processing and not when an instance of a token is processing, which is how it works now.

### Chain support
Share to Gallery is currently supported for chains (might already support few more, but I only tested these chains so far):
* Ethereum
* Base
* Arbitrum 
* Zora is NOT supported yet but can be, it's missing a single interface needed to support it


tokenprocessing/process.go
```diff
@@ -151,44 +134,29 @@ func processMediaForTokenIdentifiers(tp *tokenProcessor, tokenRepo *postgres.Tok
 	}
 }
 
-// processMediaForTokenInstance processes a single token instance. It's only called for tokens that failed during a sync.
-func processMediaForTokenInstance(tp *tokenProcessor, tokenRepo *postgres.TokenGalleryRepository, contractRepo *postgres.ContractGalleryRepository, tm *tokenmanage.Manager) gin.HandlerFunc {
+// processMediaForTokenManaged processes a single token instance. It's only called for tokens that failed during a sync.
+func processMediaForTokenManaged(tp *tokenProcessor, tokenRepo *postgres.TokenGalleryRepository, contractRepo *postgres.ContractGalleryRepository, tm *tokenmanage.Manager) gin.HandlerFunc {
```
benny-conn, line 138: Is this called anywhere?

service/multichain/multichain.go
```diff
@@ -433,6 +417,82 @@ func (p *Provider) SyncTokensByUserIDAndTokenIdentifiers(ctx context.Context, us
...
+
+	retryCondition := func(err error) bool {
+		logger.For(ctx).Errorf("polling for token: %s: retrying on error: %s", token.String(), err.Error())
+		return true
+	}
+
+	return contract, retry.RetryFunc(ctx, searchF, retryCondition, r)
+}
+
+func (p *Provider) PollForTokenByUserTokenIdentifiers(ctx context.Context, userID persist.DBID, w []persist.Wallet, t persist.TokenIdentifiers, r retry.Retry) (token db.Token, err error) {
```
benny-conn, line 442: The names of these two functions I think led me to a bit of confusion where they were being used in token processing. The `PollForTokenByTokenIdentifiers` returns a contract and an error despite saying it is polling for a token,  but in reality is refreshing a token's descriptors with a retry to ensure the token can actually have metadata to find those fields. On the other hand, `PollForTokenByUserTokenIdentifiers` is named similarly, but returns a token and an error, and actually fetches the token from the various providers. This one seems to add the token to the user where as the first one just wants to ensure metadata fields for any of the matching tokens in the DB are accurate.

benny-conn (approved):
Looks wonderful! Just a few questions, love how fault tolerant this all feels

PR #1186, opened 2023-09-18
Benny/targeted optimistic sync

benny-conn:
Changes:

- **Added** new handler for receiving tokens from alchemy webhooks
- **Fixed** optimistic syncing push notifications (forgot a return statement)

Alchemy webhooks will send a request with as much grouping as possible for all activity for a given contract. Right now we will set it up to just receive prohibition data.

### DIFF
--- a/service/notifications/notifications.go
+++ b/service/notifications/notifications.go
@@ -609,6 +609,7 @@ func createPushMessage(ctx context.Context, notif db.Notification, queries *db.Q
 		} else {
 			message.Body = fmt.Sprintf("You received a new %s token", name)
 		}
+		return message, nil
 	}
 
 	return task.PushNotificationMessage{}, fmt.Errorf("unsupported notification action: %s", notif.Action)
--- a/tokenprocessing/process.go
+++ b/tokenprocessing/process.go
@@ -377,7 +377,6 @@ func processOwnersForAlchemyTokens(mc *multichain.Provider, queries *coredb.Quer
 
 			userID, ok := addressToUsers[persist.NewChainAddress(activity.ToAddress, chain)]
 			if !ok {
-
 				user, err := queries.GetUserByAddress(c, coredb.GetUserByAddressParams{
 					Address: activity.ToAddress,
 					Chain:   int32(chain),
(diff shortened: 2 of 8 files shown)


service/notifications/notifications.go
```diff
@@ -609,6 +609,7 @@ func createPushMessage(ctx context.Context, notif db.Notification, queries *db.Q
 		} else {
 			message.Body = fmt.Sprintf("You received a new %s token", name)
 		}
+		return message, nil
```
benny-conn, line 612: This is all that is needed to have push notifs work for optimistic syncing given that I already set them up earlier.

tokenprocessing/process.go
```diff
@@ -243,6 +246,272 @@ func processOwnersForUserTokens(mc *multichain.Provider, queries *coredb.Queries
...
+				chain = persist.ChainETH
+			case "POLYGON_MAINNET":
+				chain = persist.ChainPolygon
+			case "ARBITRUM_MAINNET":
+				chain = persist.ChainArbitrum
+			case "OPTIMISM_MAINNET":
+				chain = persist.ChainOptimism
+			default:
+				util.ErrResponse(c, http.StatusInternalServerError, fmt.Errorf("invalid alchemy network: %s", input.Event.Network))
+				return
```
jarrel-b, line 375: Would we extend the webhook to allow for other chains in the future? Wonder if we should `continue` here? But probably fine if its just for alchemy!

tokenprocessing/process.go
```diff
@@ -243,6 +246,272 @@ func processOwnersForUserTokens(mc *multichain.Provider, queries *coredb.Queries
...
+			newTokens, err := mc.SyncTokensByUserIDAndTokenIdentifiers(c, userID, util.MapWithoutError(tokens, func(t alchemyTokenIdentifiers) persist.TokenUniqueIdentifiers {
+				return persist.TokenUniqueIdentifiers{
+					Chain:           t.Chain,
+					ContractAddress: t.ContractAddress,
+					TokenID:         t.TokenID,
+					OwnerAddress:    t.OwnerAddress,
+				}
+			}))
+			if err != nil {
+				util.ErrResponse(c, http.StatusInternalServerError, err)
```
jarrel-b, line 452: Would we want to `continue` here so we could process other users?

jarrel-b (approved):
Looks awesome!

PR #1191, opened 2023-09-20
Remove displayable filter on posts

jarrel-b:
Removing the displayable criteria for posts so that tokens can still be viewed.

### DIFF
--- a/graphql/dataloader/dataloaders.go
+++ b/graphql/dataloader/dataloaders.go
@@ -80,66 +80,67 @@ type IDAndLimit struct {
 // a single request, nor should they be shared between requests (since the data returned is
 // relative to the current request context, including the user and their auth status).
 type Loaders struct {
-	UserByUserID                  *UserLoaderByID
-	UserByUsername                *UserLoaderByString
-	UserByAddress                 *UserLoaderByAddress
-	UsersWithTrait                *UsersLoaderByString
-	GalleryByGalleryID            *GalleryLoaderByID
-	GalleryByCollectionID         *GalleryLoaderByID
-	GalleriesByUserID             *GalleriesLoaderByID
-	CollectionByCollectionID      *CollectionLoaderByID
-	CollectionsByGalleryID        *CollectionsLoaderByID
-	MembershipByMembershipID      *MembershipLoaderById
-	WalletByWalletID              *WalletLoaderById
-	WalletsByUserID               *WalletsLoaderByID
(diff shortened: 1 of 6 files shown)


graphql/dataloader/dataloaders.helpers.go
```diff
@@ -90,99 +90,33 @@ func withMaxBatchOne(batchSize int) func(*settings) {
...
 }
 
-func withWaitTime(t time.Duration) func(*settings) {
+func withMaxWait(t time.Duration) func(*settings) {
 	return func(s *settings) {
 		s.waitTime = t
 	}
 }
 
-func withContext(ctx context.Context) func(interface {
```
jarrel-b, line 99: Didn't realize these helpers were already here! I updated them to update settings instead.

Robinnnnn (approved):

graphql/dataloader/dataloaders.go
```diff
@@ -238,6 +239,12 @@ func NewLoaders(ctx context.Context, q *db.Queries, disableCaching bool) *Loader
 		AutoCacheWithKey: func(token db.Token) persist.DBID { return token.ID },
 	})
 
+	loaders.TokenByTokenIDIgnoreDisplayable = NewTokenLoaderByID(
+		defaultSettingsPlusOpts(ctx, disableCaching, &subscriptionRegistry, &mutexRegistry, withPublishResults(false)),
```
radazen, line 243: Good call on not publishing the results! I was initially thinking that if we made `displayable` a parameter in the regular `TokenByTokenID` fetcher, we wouldn't need to do this...but every other query that fetches tokens would need to qualify it with a `displayable` parameter, too. This seems cleaner.

graphql/dataloader/dataloaders.helpers.go
```diff
@@ -90,99 +90,33 @@ func withMaxBatchOne(batchSize int) func(*settings) {
...
-}) {
-	return func(s interface {
-		setContext(ctx context.Context)
-		setWait(time.Duration)
-		setMaxBatch(int)
-		setDisableCaching(bool)
-		setPublishResults(bool)
-	}) {
-		s.setWait(wait)
+func withContext(ctx context.Context) func(*settings) {
```
radazen, line 99: Man, I spent a _really long_ time trying to make these as concise as possible. Cannot believe I missed this obvious syntax.

radazen (approved):
🔥 🔥 🔥

PR #1195, opened 2023-09-21
manually add lens and farcaster accounts

benny-conn:
Changes:

- **Added** connect social functionality for lens and farcaster by specifying an address

Considerations:

- We could do this without having the user specify an address but then we would have to iterate through their wallets and they could have different lens/farcaster accounts for their many addresses. This forces them to specify one. (Our current backfill does iterate through all of their wallets, meaning users could have incorrect farcaster/lens accounts if they have multiple and prefer one over the other) @Robinnnnn


service/farcaster/neynar.go
```diff
@@ -51,10 +51,35 @@ UserByVerification
...
+		var i int
+		if err := json.Unmarshal(b, &i); err != nil {
+			return err
+		}
+		*n = NeynarID(fmt.Sprintf("%d", i))
+	} else {
+		*n = NeynarID(s)
+	}
+	return nil
+}
```
benny-conn, line 73: Why this is necessary, I have no idea, ask Neynar lol

radazen (approved):
👍 👍 👍

PR #1207, opened 2023-09-28
Add index for contract lookup by chain+address

radazen:


db/migrations/core/000108_update_contract_indexes.up.sql
```diff
@@ -0,0 +1,2 @@
+alter index if exists contracts_chain_address_idx rename to contracts_chain_address_no_parent_idx;
+create unique index if not exists contracts_chain_address_idx on contracts(chain, address) where not deleted;
```
radazen, line 2: I made this unique regardless of parent status because:
- there are no non-null parent fields in the database right now
- I'm planning to remove the parent column when I finish up community support

Robinnnnn (approved):

PR #1213, opened 2023-10-03
Include viewer in feed

jarrel-b:
**Changes**
* Viewer is now included in the For You Feed
* Viewer posts are only ranked using the post's engagement instead of the post's engagement and relevance to the user
* Normalizes personalization scores to 1.0
* Reduces the shuffle window size from 8 to 4

### DIFF
--- a/publicapi/feed.go
+++ b/publicapi/feed.go
@@ -427,16 +427,8 @@ func (api FeedAPI) GlobalFeed(ctx context.Context, before *string, after *string
 	return paginator.paginate(before, after, first, last)
 }
 
-func fetchFeedEntityScores(ctx context.Context, queries *db.Queries, excludeUserID persist.DBID) (map[persist.DBID]db.GetFeedEntityScoresRow, error) {
-	params := db.GetFeedEntityScoresParams{
-		IncludeViewer: true,
-		WindowEnd:     time.Now().Add(-feedLookback),
-	}
-	if excludeUserID != "" {
-		params.IncludeViewer = false
-		params.ViewerID = excludeUserID
-	}
-	scores, err := queries.GetFeedEntityScores(ctx, params)
+func fetchFeedEntityScores(ctx context.Context, queries *db.Queries) (map[persist.DBID]db.GetFeedEntityScoresRow, error) {
+	scores, err := queries.GetFeedEntityScores(ctx, time.Now().Add(-feedLookback))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


service/recommend/userpref/personalize.go
```diff
@@ -196,6 +196,11 @@ func (p *Personalization) RelevanceTo(userID persist.DBID, e db.FeedEntityScore)
 		return 0
 	}
 
+	// If scoring themselves, return 1 which is the max score
+	if userID == e.ActorID {
+		return 1.0
```
jarrel-b, line 201: This score doesn't actually get used in practice but it seems more correct to say that a user's relevance to themselves is 1

service/recommend/userpref/personalize.go
```diff
@@ -288,6 +293,9 @@ func (p *Personalization) readCache(ctx context.Context) {
 
 // calcSocialScore determines if vIdx is in the same friend circle as qIdx by running a bfs on userM
 func calcSocialScore(userM *sparse.CSR, vIdx, qIdx int) float64 {
+	if vIdx == qIdx {
```
jarrel-b, line 296: Users' distance to themselves is 1 - fixes float division which led to scores of `+Inf`. This doesn't actually get executed in practice because of the earlier guard clause that returns early if we're scoring a user's relevance to themselves.

publicapi/feed.go
```diff
@@ -603,22 +595,19 @@ func (api FeedAPI) CuratedFeed(ctx context.Context, before, after *string, first
...
-		topNByEngagement = api.scoreFeedEntities(ctx, 128, topNByEngagement, func(e db.FeedEntityScore) float64 { return personalizationScores[e.ID] })
+		topNByPersonalization := api.scoreFeedEntities(ctx, 128, scores, func(e db.FeedEntityScore) float64 { return engagementScores[e.ID] })
+		topNByPersonalization = api.scoreFeedEntities(ctx, 128, topNByPersonalization, func(e db.FeedEntityScore) float64 { return personalizationScores[e.ID] })
+
 		// Rank by personalization, then by engagement
-		topNByPersonalization := api.scoreFeedEntities(ctx, 128, scores, func(e db.FeedEntityScore) float64 { return personalizationScores[e.ID] })
-		topNByPersonalization = api.scoreFeedEntities(ctx, 128, topNByPersonalization, func(e db.FeedEntityScore) float64 { return engagementScores[e.ID] })
+		topNByEngagement := api.scoreFeedEntities(ctx, 128, scores, func(e db.FeedEntityScore) float64 { return personalizationScores[e.ID] })
+		topNByEngagement = api.scoreFeedEntities(ctx, 128, topNByEngagement, func(e db.FeedEntityScore) float64 { return engagementScores[e.ID] })
```
jarrel-b, line 610: Just renaming confusing variable names

service/recommend/userpref/personalize.go
```diff
@@ -314,7 +322,7 @@ func calcSimilarityScore(simM *sparse.CSR, vIdx, qIdx int) float64 {
 }
 
 // idLookup uses a slice to lookup a value by its index
-// Its purpose is to avoid using a map because indexing a slice is suprisingly much quicker than a map lookup.
+// Its purpose is to avoid using a map because indexing a slice is surprisingly much quicker than a map lookup.
```
Robinnnnn, line 325: thank you

Robinnnnn (approved):

PR #1217, opened 2023-10-04
Add reservoir to polygon

jarrel-b:
**Changes**
* Replaces Opensea as a provider with Reservoir for Polygon
* Refactors `inject.go`:
    * Providers now share the same client instance of the token metadata cache instead of keeping separate clients
    * Removes throwaway inject functions
    * Cleans up each chain's provider sets so it's easier to see which providers are used per chain
    * Moves provider de-duplication logic to multichain initialization
* Minor refactor in `pipeline.go` to remove the need of a channel when processing media

### DIFF
--- a/server/inject.go
+++ b/server/inject.go
@@ -371,24 +371,20 @@ func baseRequirements(
 // polygonProviderSet is a wire injector that creates the set of polygon providers
 func polygonProviderSet(*http.Client) polygonProviderList {
 	wire.Build(
-		rpc.NewEthClient,
 		polygonProvidersConfig,
-		wire.Value(persist.ChainPolygon),
 		newTokenProcessingCache,
 		// Add providers for Polygon here
 		newPolygonProvider,
-		opensea.NewProvider,
 	)
 	return polygonProviderList{}
 }
 
 // polygonProvidersConfig is a wire injector that binds multichain interfaces to their concrete Polygon implementations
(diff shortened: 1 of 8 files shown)


server/inject.go
```diff
@@ -105,28 +101,26 @@ func newQueries(p *pgxpool.Pool) *db.Queries {
...
 
 // ethProvidersConfig is a wire injector that binds multichain interfaces to their concrete Ethereum implementations
-func ethProvidersConfig(indexerProvider *eth.Provider, openseaProvider *opensea.Provider, fallbackProvider multichain.SyncFailureFallbackProvider, alchemyProvider *alchemy.Provider) ethProviderList {
+func ethProvidersConfig(indexerProvider *eth.Provider, openseaProvider *opensea.Provider, fallbackProvider multichain.SyncFailureFallbackProvider) ethProviderList {
 	wire.Build(
 		wire.Bind(new(multichain.NameResolver), util.ToPointer(indexerProvider)),
 		wire.Bind(new(multichain.Verifier), util.ToPointer(indexerProvider)),
 		wire.Bind(new(multichain.TokensOwnerFetcher), util.ToPointer(fallbackProvider)),
-		wire.Bind(new(multichain.TokensContractFetcher), util.ToPointer(alchemyProvider)),
+		wire.Bind(new(multichain.TokensContractFetcher), util.ToPointer(fallbackProvider)),
```
jarrel-b, line 123: The fallback provider should be more robust than the alchemy provider (and uses alchemy anyways as the primary)

PR #1219, opened 2023-10-05
l1 chain

benny-conn:
Changes:

- **Added** a new `l1_chain` column to `wallets` and `nonces` table.
- **Changed** many queries to search by `l1_chain`
- **Changed** `WalletOverrides` to live in persist instead of wire injection


db/migrations/core/000111_l1_chain.up.sql
```diff
@@ -0,0 +1,12 @@
...
+alter table nonces alter column l1_chain set not null;
+alter table wallets add column l1_chain int;
+update wallets set l1_chain = 0;
+alter table wallets alter column l1_chain set not null;
+update nonces set l1_chain = 4 where chain = 4;
+update wallets set l1_chain = 4 where chain = 4;
+
+create index nonces_l1_chain_idx on nonces (address,chain,l1_chain);
+create index wallets_l1_chain_idx on wallets (address,chain,l1_chain);
+create unique index wallets_l1_chain_unique_idx on wallets (address,l1_chain) where deleted = false;
```
benny-conn, line 12: Do we need to remove any of the current indexes, specifically the unique ones on `(address,chain)`? I imagined we still would want them because we still need those two to be unique (despite them always being unique given we correctly use the l1_chain, e.g. there will be no case where two rows with `(addr: 0x9, ch: BASE)` and `(addr:0x9, ch: BASE)` getting through because they will always fail on `(addr: 0x9, l1_ch: ETH)` anyway). I imagine leaving them would not hurt us.

service/persist/postgres/wallet.go
```diff
@@ -101,20 +97,3 @@ func (w *WalletRepository) GetByUserID(ctx context.Context, userID persist.DBID)
...
-	}
-
-	// rather than using the ID generated above, we must retrieve it because in the case of conflict the ID above would be inaccurate.
-	wa, err := w.GetByChainAddress(ctx, chainAddress)
-	if err != nil {
-		return "", err
-	}
-
-	return wa.ID, nil
-}
```
benny-conn, line 120: Unused

sqlc.yaml
```diff
@@ -210,14 +210,14 @@ sql:
...
+          - column: '*.*.l1_chain'
+            go_type: 'github.com/mikeydub/go-gallery/service/persist.Chain'
           - column: '*.*.action'
             go_type: 'github.com/mikeydub/go-gallery/service/persist.Action'
           - column: '*.*.thumbnail_url'
             go_type: 'github.com/mikeydub/go-gallery/service/persist.NullString'
           - column: '*.*.media_url'
             go_type: 'github.com/mikeydub/go-gallery/service/persist.NullString'
-          - column: '*.*.chain'
-            go_type: 'github.com/mikeydub/go-gallery/service/persist.Chain'
```
benny-conn, line 220: Duplicated above

db/queries/core/query.sql
```diff
@@ -1584,40 +1587,15 @@ join wallets on wallets.id = any(tokens.owned_by_wallets)
...
 select sqlc.embed(c),
        w.id as wallet_id,
        false as is_override_creator
 from users u, contracts c, wallets w
 where u.id = @user_id
   and c.chain = any(@chains::int[])
-  and w.id = any(u.wallets) and coalesce(nullif(c.owner_address, ''), nullif(c.creator_address, '')) = w.address
-  and c.chain = w.chain
+  and w.id = any(u.wallets) and coalesce(nullif(c.owner_address, ''), nullif(c.creator_address, '')) = w.address 
+  and w.chain = any(@l1_chains::int[])
```
radazen, line 1598: Hmm, this makes me think we should add the `l1_chain` column to the `contracts` table, too. That way, we can say "give me contracts on any(@chains) where the user owns a wallet on the contract's L1 chain," which is (I think) the most correct way to phrase this query.

...but also, that would add some pretty significant scope to this PR, and I don't think it's necessary to add it for the sake of this one query. I think your current workaround is good here, but we should definitely add `l1_chain` to contracts in the near future!

service/persist/auth.go
```diff
@@ -22,6 +22,8 @@ type UserNonce struct {
 	LastUpdated  time.Time  `json:"last_updated"`
 	Value        NullString `json:"value"`
 	Address      Address    `json:"address"`
+	Chain        Chain      `json:"chain"`
+	L1Chain      Chain      `json:"l1_chain"`
```
radazen, line 26: Could we add a new type for `L1Chain` instead of repurposing `Chain`? IMHO we wouldn't want to accept an `L1Chain` anywhere we'd accept a regular `Chain`, or vice versa.

service/persist/postgres/wallet.go
```diff
@@ -25,16 +25,13 @@ func NewWalletRepository(db *sql.DB, queries *db.Queries) *WalletRepository {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
 	defer cancel()
 
-	insertStmt, err := db.PrepareContext(ctx, `INSERT INTO wallets (ID,VERSION,ADDRESS,CHAIN,WALLET_TYPE,DELETED) VALUES ($1,$2,$3,$4,$5,false) ON CONFLICT (ADDRESS,CHAIN) WHERE DELETED = false DO NOTHING;`)
+	getByIDStmt, err := db.PrepareContext(ctx, `SELECT ID,VERSION,CREATED_AT,LAST_UPDATED,ADDRESS,WALLET_TYPE,CHAIN,L1_CHAIN FROM wallets WHERE ID = $1 AND DELETED = FALSE;`)
 	checkNoErr(err)
 
-	getByIDStmt, err := db.PrepareContext(ctx, `SELECT ID,VERSION,CREATED_AT,LAST_UPDATED,ADDRESS,WALLET_TYPE,CHAIN FROM wallets WHERE ID = $1 AND DELETED = FALSE;`)
+	getByChainAddressStmt, err := db.PrepareContext(ctx, `SELECT ID,VERSION,CREATED_AT,LAST_UPDATED,ADDRESS,WALLET_TYPE,CHAIN,L1_CHAIN FROM wallets WHERE ADDRESS = $1 AND L1_CHAIN = $2 AND DELETED = FALSE;`)
```
radazen, line 31: Do we need the concept of an `L1ChainAddress` that uses `L1Chain` instead of a regular chain? It's starting to feel that way to me! A query named `getByChainAddress` makes me think we'll be passing in a `Chain`, but we're actually using an `L1Chain` here, and it'd probably make the codebase clearer if we separated those concepts.

radazen (approved):
Looks awesome, Benny! I really like how this turned out, and I think it's going to simplify a lot of the L1 chain issues we've been running into!

db/queries/core/query.sql
benny-conn: Agreed, was thinking this as well when I had to make this little work around. If we do contracts, might as well do tokens too at some point...

service/persist/auth.go
benny-conn: Sure!

service/persist/postgres/wallet.go
benny-conn: For sure!

PR #1224, opened 2023-10-10
Benny/zora rest

benny-conn:
Changes:

- **Changed** zora provider to use rest instead of GQL


service/multichain/zora/zora.go
```diff
@@ -165,229 +152,67 @@ func (d *Provider) GetTokensIncrementallyByWalletAddress(ctx context.Context, ad
...
+		if strings.EqualFold(token.ContractAddress.String(), contractAddress.String()) {
+			filteredTokens = append(filteredTokens, token)
+		}
+	}
+	foundContract := multichain.ChainAgnosticContract{}
+	for _, contract := range contracts {
+		if strings.EqualFold(contract.Address.String(), contractAddress.String()) {
+			foundContract = contract
+		}
+	}
```
benny-conn, line 203: I will see if I can suggest to them to allow us to query by multiple fields like we used to be able to in the GQL API. For now, we have to get all of a user's tokens then filter them by the ones we want, for example, by contract address.

PR #1225, opened 2023-10-10
someone posted your work notification

benny-conn:
Changes:

- **Added** a new notification for when someone posts tokens from a contract that a user owns
- **Changed** post handlers to attempt to find a contract creator user and dispatch events accordingly


jarrel-b (approved):
🚢 🚢 🚢

PR #1227, opened 2023-10-10
Contract L1 Chain

benny-conn:
Changes:

- **Added** `L1Chain` to the contracts table
- **Changed** `contract_creators` view to use the `L1Chain` of the `wallets` and `contracts` table

TODO:

- [x] Update contract queries to use L1 chain

### DIFF
--- a/db/queries/core/contract_gallery.sql
+++ b/db/queries/core/contract_gallery.sql
@@ -14,7 +14,7 @@ insert into contracts(id, deleted, version, created_at, address, symbol, name, o
     , unnest(@profile_image_url::varchar[])
     , unnest(@provider_marked_spam::bool[])
 )
-on conflict (l1_chain, address) where parent_id is null
+on conflict (l1_chain, chain, address) where parent_id is null
 do update set symbol = coalesce(nullif(excluded.symbol, ''), nullif(contracts.symbol, ''))
   , version = excluded.version
   , name = coalesce(nullif(excluded.name, ''), nullif(contracts.name, ''))
@@ -46,7 +46,7 @@ insert into contracts(id, deleted, version, created_at, name, address, creator_a
     , unnest(@description::varchar[])
     , unnest(@parent_ids::varchar[])
 )
-on conflict (l1_chain, parent_id, address) where parent_id is not null
+on conflict (l1_chain, chain, parent_id, address) where parent_id is not null
 do update set deleted = excluded.deleted
   , name = excluded.name
   , creator_address = excluded.creator_address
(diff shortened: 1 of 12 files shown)


db/queries/core/contract_gallery.sql
```diff
@@ -9,11 +9,12 @@ insert into contracts(id, deleted, version, created_at, address, symbol, name, o
     , unnest(@name::varchar[])
     , unnest(@owner_address::varchar[])
     , unnest(@chain::int[])
+    , unnest(@l1_chain::int[])
     , unnest(@description::varchar[])
     , unnest(@profile_image_url::varchar[])
     , unnest(@provider_marked_spam::bool[])
 )
-on conflict (chain, address) where parent_id is null
+on conflict (l1_chain, address) where parent_id is null
```
radazen, line 17: Can there be multiple contracts at the same address across L2 chains? This looks like it'd essentially merge all contracts at the same address across L2 chains if they're on the same L1, but that sounds okay to me as long as it's a valid thing to actually do! I defer to your blockchain expertise here.

service/multichain/multichain.go
```diff
@@ -1463,8 +1461,9 @@ func (p *Provider) RefreshTokenDescriptorsByTokenIdentifiers(ctx context.Context
 		return persist.ErrTokenNotFoundByTokenIdentifiers{Token: ti}
 	}
 
-	contractID, err := p.Repos.ContractRepository.UpsertByAddress(ctx, ti.ContractAddress, ti.Chain, persist.ContractGallery{
+	contractID, err := p.Repos.ContractRepository.UpsertByAddress(ctx, ti.ContractAddress, persist.ContractGallery{
 		Chain:           ti.Chain,
+		L1CHain:         ti.Chain.L1Chain(),
```
radazen, line 1466: typo: `L1CHain` -> `L1Chain`

service/recommend/userpref/personalize.go
```diff
@@ -266,7 +266,7 @@ func (p *Personalization) update(ctx context.Context) {
 
 	if p.pM == nil {
 		logger.For(ctx).Infof("no data loaded, reading from cache")
-		p.readCache(ctx)
+		// p.readCache(ctx)
```
radazen, line 269: Accidental commit?

radazen (approved):
Looks awesome, @benny-conn! I love the direction we're going with this. 🚀 🚀 🚀

db/queries/core/contract_gallery.sql
benny-conn: You are in fact correct, although infinitesimally improbable, it can happen. Good thing we are keeping the original unique index around!

PR #1232, opened 2023-10-16
Benny/email lock and templates

benny-conn:
Changes:

- **Added** a redis lock to the notification email subscription receiver so that only one service handles the pubsub and sends notification emails
- **Added** new email notification types to the notification template data creator and shared logic with push notifications

### DIFF
--- a/emails/send.go
+++ b/emails/send.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/mikeydub/go-gallery/service/auth"
 	"net/http"
 	"sync/atomic"
 	"time"
+
+	"github.com/bsm/redislock"
+	"github.com/mikeydub/go-gallery/service/auth"
 
 	"cloud.google.com/go/pubsub"
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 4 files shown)


Robinnnnn (approved):

emails/send.go
```diff
@@ -141,12 +137,17 @@ func adminSendNotificationEmail(queries *coredb.Queries, s *sendgrid.Client) gin
...
-func autoSendNotificationEmails(queries *coredb.Queries, s *sendgrid.Client, psub *pubsub.Client) error {
+func autoSendNotificationEmails(queries *coredb.Queries, s *sendgrid.Client, psub *pubsub.Client, r *redislock.Client) error {
 	ctx := context.Background()
 	sub := psub.Subscription(env.GetString("PUBSUB_NOTIFICATIONS_EMAILS_SUBSCRIPTION"))
 
 	return sub.Receive(ctx, func(ctx context.Context, msg *pubsub.Message) {
-		err := sendNotificationEmailsToAllUsers(ctx, queries, s, env.GetString("ENV") == "production")
+		l, err := r.Obtain(ctx, "send-notification-emails", time.Minute*5, nil)
+		if err != nil {
+			return // don't ack message
```
jarrel-b, line 147: Does leaving a message un-acked cause the message to get re-delivered? If so, I think ack'ing the message if `ErrNotObtained` is returned (since something else is handling it and will call `Nack` if it fails) so that it won't get consumed again.

jarrel-b (approved):
fyi I glossed over the template stuff, but lgtm!

emails/send.go
benny-conn: You are correct! Thank you!

PR #1237, opened 2023-10-24
write auth for farcaster and lens

benny-conn:
Changes:

- **Changed** farcaster and lens auth to optionally include fields for signing and authenticating with OAuth
- **Added** polling task queue for polling farcaster signer approval

Farcaster is by far the more complex flow, here is how this PR works to accomplish authenticated farcaster connections:

1. When connecting social auth, user can optionally request `withSigner` in the graphQL request that will have Neynar generate a "signer". A signer is useless until approved by the user later in this process but we don't want to expose the `signer_uuid` which is sort of like the "access token" in farcaster.
2. We use our official gallery farcaster private key to sign the public key associated with this signer and send it to neynar, basically saying "gallery wants this guy to approve us to act on his behalf". 
3. We get back an "approval URL" that will be sent to the user and can be used to approve gallery in an on-chain transaction. This will be returned in the `FarcasterSocialAccount` type as well as the status of their approval. The frontend has to give this URL to the user.
4. Simultaneously, we kick off a retryable task to the `autosocial` service that will continuously check if the user approved the token. Once they do, it will update their status to approved.

The last step is to match the "deadline" of the signature to whatever the cloud task total retry time is so that we don't run into a case where a user successfully approves of a token and we didn't poll for it. If this does happen, this isn't the end of the world, but might lead to us telling the user "you haven't approved of gallery to act on your behalf" when they in fact have. We could also remove the polling altogether but then the frontend would have no way of showing the user that they successfully connected.


service/task/cloudtask.go
```diff
@@ -218,6 +223,14 @@ func CreateTaskForAutosocialProcessUsers(ctx context.Context, message Autosocial
 	return submitTask(ctx, client, queue, url, withJSON(message), withTrace(span))
 }
 
+func CreateTaskForAutosocialPollFarcaster(ctx context.Context, message AutosocialPollFarcasterMessage, client *gcptasks.Client) error {
+	span, ctx := tracing.StartSpan(ctx, "cloudtask.create", "createTaskForAutosocialPollFarcaster")
+	defer tracing.FinishSpan(span)
+	queue := env.GetString("AUTOSOCIAL_POLL_QUEUE")
+	url := fmt.Sprintf("%s/checkFarcasterApproval?signer_uuid=%s&user_id=%s", env.GetString("AUTOSOCIAL_URL"), message.SignerUUID, message.UserID)
+	return submitTask(ctx, client, queue, url, withTrace(span))
```
radazen, line 231: What do you think the retry on this will look like? Seems like we'll want to try a bunch early on so we can be quick to update their account if they do the on-chain thing, but we probably want to back off significantly if they don't do the transaction right away.

radazen (approved):
Good job figuring this stuff out, @benny-conn! Looks great to me. 🥇

autosocial/handler.go
```diff
@@ -65,6 +67,55 @@ func processUsers(q *coredb.Queries, n *farcaster.NeynarAPI, l *lens.LensAPI) gi
...
 
+func checkFarcasterApproval(q *coredb.Queries, n *farcaster.NeynarAPI) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		var in task.AutosocialPollFarcasterMessage
+		if err := c.ShouldBindQuery(&in); err != nil {
+			util.ErrResponse(c, http.StatusBadRequest, err)
+			return
+		}
+
+		logger.For(c).Infof("checking farcaster approval for %s", in.SignerUUID)
```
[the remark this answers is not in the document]
benny-conn: Yes!

service/task/cloudtask.go
benny-conn: Sounds good to me. I imagine this will all be in the span of about 10 minutes as well.

PR #1238, opened 2023-10-24
Token Normalization

jarrel-b:
#### DB Changes
This PR adds the `token_definitions` table, which contains data representing the canonical definition of a token. The schema changes are outlined below (metadata columns i.e. `created_at`, `last_updated`, `deleted` omitted for brevity):
```
token_definitions
-----------------
name
description
token_type
token_id
external_url
chain
metadata
fallback_media
contract_address
is_provider_marked_spam
contract_id FK(contracts)
token_media_id FK(token_medias)

tokens
-----------------
/* Added columns */
+ token_definition_id FK(token_definitions)

/* Columns planned for deprecation */
- name
- description
- token_type
- token_id
- ownership_history
- external_url
- chain
- is_provider_marked_spam
- token_uri
- fallback_media
- token_media_id

token_medias
-----------------
/* Columns planned for deprecation */
- name
- description
- metadata
- contract_id
- token_id
- chain
```
I removed the `ownership_history` and `token_uri` entirely from the database schema since neither field is used in the app currently. The `ownership_history` field wasn't comprehensive, and the `token_uri` was only used internally by tokenprocessing to retrieve metadata, but it doesn't need to be committed to the database.

Note that the consequence of the new schema is that updates to a `token_definition`'s media are used by all `tokens` instances of that definition since a `tokens` entity no longer has a reference to its own `token_media_id`.

Note that I also kept `contract_id` (renamed from `contract`) on the `tokens` table since there were quite a few views and queries that depended on it and it seemed useful to keep around.

#### GraphQL Changes
The `TokenDefinition` type is now available in the schema. The existing `Token` type has a reference to its `TokenDefinition` and also has fields marked for future deprecation. I didn't add queries for getting a `TokenDefinition` yet, but when the app needs only non user-scoped token data, we'll want to add ways to get a `TokenDefinition` since it's quicker than resolving a `Token` now that resolving once trivial fields (`name`, `description`, `tokenId`) requires a waterfall to retrieve its token definition internally.

```graphql
type TokenDefinition implements Node @goEmbedHelper {
  id: ID!
  dbid: DBID!
  creationTime: Time
  lastUpdated: Time
  media: MediaSubtype @goField(forceResolver: true)
  tokenType: TokenType
  chain: Chain
  name: String
  description: String
  tokenId: String
  tokenMetadata: String @goField(forceResolver: true)
  community: Community @goField(forceResolver: true)
  externalUrl: String
}

type Token {
  definition: TokenDefinition! @goField(forceResolver: true)
}
```

#### Multichain changes
Multichain is updated to remove references to `persist.TokenGallery` and `persist.ContractGallery` in favor of sqlc's `db.Contract` and `persist.TokenFullDetails` which embeds `db.Token`, `db.Contract`, and `db.TokenDefinition` in its struct. The addition of new token definitions is also handled in `multichain`. Before upserting, token data from providers are de-duplicated to produce a set of token definitions and a set of user tokens. Both sets are used as input to the upsert token data atomically: the definitions are first added, then a reference to the definition is added to the token instance.

#### Tokenprocessing changes
Tokenprocessing now takes token definition DBIDs as inputs rather than the original token DBIDs. Tokenmanage now keeps track of the state using the definition ID rather than the token DBID. The core of the pipeline itself now also runs on `persist.TokenIdentifiers` and `persist.ContractIdentifiers` and has no dependencies with the db types.


graphql/resolver/schema.resolvers.helpers.go
```diff
@@ -1958,47 +1945,54 @@ func tokenHolderToModel(ctx context.Context, tokenHolder persist.TokenHolder) *m
...
+		Chain:                 nil, // handled by dedicated resolver
+		Name:                  nil, // handled by dedicated resolver
+		Description:           nil, // handled by dedicated resolver
+		TokenID:               nil, // handled by dedicated resolver
+		TokenMetadata:         nil, // handled by dedicated resolver
+		Contract:              nil, // handled by dedicated resolver
+		ExternalURL:           nil, // handled by dedicated resolver
+		IsSpamByProvider:      nil, // handled by dedicated resolver
+		OwnedByWallets:        nil, // handled by dedicated resolver
+		BlockNumber:           nil,
```
jarrel-b, line 1994: Not used in the app currently, but we still record this in the database.

graphql/stub_test.go
jarrel-b, line 1: Refactoring a few test methods, feel free to skip.

publicapi/token.go
```diff
@@ -274,36 +274,7 @@ func (api TokenAPI) GetTokensByUserID(ctx context.Context, userID persist.DBID,
 		params.IncludeCreator = true
 	}
 
-	results, err := api.loaders.GetTokensByUserIdBatch.Load(params)
```
jarrel-b, line 277: This query embedded the `contract` before, but the contract never actually gets used.

service/persist/postgres/token_gallery.go
jarrel-b, line 1: I removed most of the hand-written SQL from this file, there are a few queries I left in here because they were sticky to factor out.

service/persist/postgres/token_gallery.go
```diff
@@ -307,174 +128,81 @@ func (t *TokenGalleryRepository) UpsertTokens(ctx context.Context, pTokens []per
...
 
-type uniqueConstraintKey struct {
-	tokenID     persist.TokenID
-	contract    persist.DBID
-	chain       persist.Chain
-	ownerUserID persist.DBID
+	return upserted[0].Token.LastSynced, upsertedTokens, nil
 }
 
-func (t *TokenGalleryRepository) dedupeTokens(pTokens []persist.TokenGallery) []persist.TokenGallery {
```
jarrel-b, line 463: I removed the deduping logic and made it the caller's responsibility on how tokens should be deduped since they have more context on what data to keep and merge.

db/migrations/core/000115_token_norm_update_feed_entity_score_view.up.sql
jarrel-b, line 1: The following migrations are needed to update existing views and foreign keys to reference the rebuilt `tokens` table.

service/recommend/recommend.go
```diff
@@ -58,6 +58,7 @@ func NewRecommender(queries *db.Queries) *Recommender {
 
 	r.SaveResultFunc = func(ctx context.Context, userID persist.DBID, recommendedIDs []persist.DBID) error {
 		params := db.UpdatedRecommendationResultsParams{}
+		recommendedIDs = util.Dedupe(recommendedIDs, true)
```
jarrel-b, line 61: Unrelated, but noticed during testing that user recommendations contained duplicated user IDs somehow 🤔. The algo should prevent this, but I added this as a work around for the time being.

graphql/graphql_test.go
jarrel-b, line 1: Refactoring a few fixtures, feel free to skip

service/persist/postgres/token_gallery.go
```diff
@@ -3,262 +3,80 @@ package postgres
...
+		return TokenFullDetails{}, err
 	}
-
-	if err := rows.Err(); err != nil {
-		return nil, err
-	}
-
-	return tokens, nil
-
+	return TokenFullDetails{
```
radazen, line 43: The more I think about it, the more I think that `tokens` should be something like `token_instances`, and we should work exclusively with `Token` and `UserToken` types that encapsulate all the rest of this stuff. Seems like a good discussion topic for future work!

radazen: Also, should this struct be in the `persist` package? I don't have a strong opinion, just curious because this feels like an outlier among DB types!

radazen (approved):
Man, the sheer amount of thought and effort that went into this PR is inspiring. I left some design comments, but they're all pretty minor! Looking forward to testing it out on dev! Awesome stuff, @jarrel-b! Thanks for refactoring and cleaning stuff up along the way, too!

publicapi/token.go
jarrel-b: Makes sense! I'll add this back.

cmd/token_normalization_migrate/main.go
```diff
@@ -0,0 +1,499 @@
...
+	token_type
+) (
+	select
+		id,
+		nullif(name, ''),
+		nullif(description, ''),
+		nullif(external_url, ''),
+		metadata,
+		fallback_media,
+		nullif(contract_id, ''),
```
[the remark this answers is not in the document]
jarrel-b: Good call, will fix

db/queries/core/query.sql
```diff
@@ -91,8 +91,87 @@ SELECT c.* FROM galleries g, unnest(g.collections)
...
+-- name: GetTokenFullDetailsByContractId :many
+select sqlc.embed(tokens), sqlc.embed(token_definitions), sqlc.embed(contracts)
+from tokens
+join token_definitions on tokens.token_definition_id = token_definitions.id
+join contracts on token_definitions.contract_id = contracts.id
+where contracts.id = $1 and tokens.displayable and not tokens.deleted and not token_definitions.deleted and not contracts.deleted
+order by tokens.block_number desc;
+
+-- name: UpdateTokenCollectorsNoteByTokenDbidUserId :exec
+update tokens set collectors_note = $1, last_updated = now() where id = $2 and owner_user_id = $3;
```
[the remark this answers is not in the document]
jarrel-b: Good to know, will keep this in mind!

graphql/resolver/schema.resolvers.go
```diff
@@ -2932,6 +2895,83 @@ func (r *tokenResolver) ViewerAdmire(ctx context.Context, obj *model.Token) (*mo
...
+	td, err := resolveTokenDefinitionByID(ctx, obj.HelperTokenData.Token.TokenDefinitionID)
+	if err != nil {
+		return nil, err
+	}
+	return td.ExternalURL, nil
+}
+
+// Media is the resolver for the media field.
+func (r *tokenDefinitionResolver) Media(ctx context.Context, obj *model.TokenDefinition) (model.MediaSubtype, error) {
+	media, err := publicapi.For(ctx).Token.GetMediaByTokenDefinitionID(ctx, obj.Dbid)
```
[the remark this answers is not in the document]
jarrel-b: Ah good to know, will add a method for `MediaByMediaID` anyways to remove some indirection.

graphql/schema/schema.graphql
```diff
@@ -100,7 +100,6 @@ type GalleryUser implements Node @goEmbedHelper {
   # as opposed to retrieving user -> wallets -> tokens, which would contain duplicates for any token
   # that appears in more than one of the user's wallets.
   tokens(ownershipFilter: [TokenOwnershipType!]): [Token] @goField(forceResolver: true)
-  tokensByChain(chain: Chain!): ChainTokens @goField(forceResolver: true)
```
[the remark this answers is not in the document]
jarrel-b: Yup! Confirmed this isn't used anywhere

service/multichain/multichain.go
```diff
@@ -944,117 +936,80 @@ func (p *Provider) SyncTokensCreatedOnSharedContracts(ctx context.Context, userI
...
-
-		for _, token := range currentTokensForUser {
-			if newTokensForUser[token.TokenIdentifiers()] {
-				newPersistedTokens = append(newPersistedTokens, token)
-				newPersistedTokenIDs = append(newPersistedTokenIDs, token.ID)
-				newPersistedTokenIdentifiers = append(newPersistedTokenIdentifiers, token.TokenIdentifiers())
-			}
-		}
-
+		newPersistedTokens := util.Filter(currentTokensForUser, func(t postgres.TokenFullDetails) bool { return t.Instance.CreatedAt.Equal(t.Instance.LastUpdated) }, false)
```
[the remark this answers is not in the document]
jarrel-b: very good point! I'll  update this to use the old way which compared tokens before the sync to all tokens after the sync

publicapi/merch.go
```diff
@@ -83,60 +73,32 @@ func (api MerchAPI) GetMerchTokens(ctx context.Context, address persist.Address)
...
-				return nil, fmt.Errorf("failed to marshal token metadata: %w", err)
-			}
-			var metadata merchMetadata
-			if err := json.Unmarshal(asBytes, &metadata); err != nil {
-				return nil, fmt.Errorf("failed to unmarshal token metadata: %w", err)
-			}
-			for _, attr := range metadata.Attributes {
-				if attr.TraitType == "Object" {
-					switch attr.Value {
-					case "001":
```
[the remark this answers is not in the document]
jarrel-b: Oh yea I changed the heuristic to use `name` instead of `trait_type` because I saw a few tokens had missing `trait_type` in our db, but `name` was always present. re: using the tokenURI instead that column was cleared a few months ago so it wasn't hitting that code path anyway.

service/multichain/multichain.go
```diff
@@ -2236,15 +2167,19 @@ func (e errWithPriority) Error() string {
...
+	return util.DedupeWithTranslate(wallets, false, func(w persist.Wallet) string {
+		return fmt.Sprintf("%d:%s", w.Chain, w.Address)
+	})
+}
 
-	ret := make([]persist.Wallet, 0, len(wallets))
-	for _, wallet := range deduped {
-		ret = append(ret, wallet)
-	}
+func dedupeTokenDefinitions(tDefs []db.TokenDefinition) (uniqueDefs []db.TokenDefinition) {
```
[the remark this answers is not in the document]
jarrel-b: Good idea! Will add

service/persist/token.go
```diff
@@ -1071,5 +1070,29 @@ type ErrContractCreatorNotFound struct {
 }
 
 func (e ErrContractCreatorNotFound) Error() string {
-	return fmt.Sprintf("ContractCreator not found for contractID %s", e.ContractID)
+	return fmt.Sprintf("contractCreator not found by contractID=%s", e.ContractID)
+}
+
+type ErrTokenDefinitionNotFoundByID struct {
```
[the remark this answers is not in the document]
jarrel-b: Ah yea, the one thing I don't like about the more general error like:
```go
type ErrTokenDefinitionNotFound {
  ID DBID
  TokenDBID DBID
  TokenIdentifiers persist.TokenIdentifiers
}
```
is that it's easy to forget to update the message when there's a different lookup e.g.:
```go
func (e ErrTokenDefinitionNotFound) Error() string {
   if e.ID != "" {
      return "not found by id:" + e.ID
   }
   // Forgot to add a message for token identifiers!!
   return "not found by token dbid:" + e.TokenDBID
}
```

but that could just be me, and alternatively, you could print every field which is probably fine too. I also like searching for the individual error and finding exactly where that error is returned. That in mind, maybe we should go with option b) to get the best of both worlds and callers can just catch `ErrNotFound` or `ErrTokenDefinitionNotFound` depending on the context

```go
type ErrNotFound struct {}

func (e ErrNotFound) Error() string { return "entity not found" }

type ErrTokenDefinitionNotFound struct {}

func (e ErrTokenDefinitionNotFound) Error() string { return "token definition not found" }
func (e ErrTokenDefinitionNotFound) Unwrap() error { return ErrNotFound{} }

type ErrTokenDefinitionNotFoundByID struct { ID persist.DBID }
func (e ErrTokenDefinitionNotFoundByID) Error() string { return "token definition not found by id:" + e.ID }
func (e ErrTokenDefinitionNotFoundByID) Unwrap string { return  ErrTokenDefinitionNotFound{} }
```

service/persist/postgres/token_gallery.go
jarrel-b: Moving it to `persist` caused an import cycle which is why it's in `postgres` 😬 . I was thinking of moving this to `multichain` since it's the only thing that uses this type

db/queries/core/query.sql
```diff
@@ -91,8 +91,87 @@ SELECT c.* FROM galleries g, unnest(g.collections)
...
+    and tokens.id = $1
+    and not tokens.deleted
+    and not token_definitions.deleted;
+
+-- name: GetTokenDefinitionByTokenIdentifiers :one
+select *
+from token_definitions
+where (chain, contract_address, token_id) = (@chain, @contract_address, @token_id) and not deleted;
+
+-- name: GetTokenDefinitionAndMediaByTokenIdentifiersIgnoringStatus :one
```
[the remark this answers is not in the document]
jarrel-b: Ahh good catch. That will be an issue...the default starting state sounds like a good idea and having `token_media_id` non-nullable would be nice too, but I'll need to think through how that might impact the frontend and how it displays that token. 

I intended for this to be an optimization so that the definition and the media can be fetched in one query rather than two separate ones. I think for now I'll use the `sqlc.embed` cache trick to also include the definition in the root query, so at least the definition will be cached at that point, so it's only one additional query to fetch the media.

service/persist/postgres/token_gallery.go
jarrel-b: Going to leave this in `postgres` for now since I see this type going away when we have non-nullable media ids / configure sqlc null pointers!

service/multichain/multichain.go
```diff
@@ -1962,53 +1959,21 @@ func tokensToNewDedupedTokens(tokens []chainTokens, existingTokens []persist.Tok
...
 			}
 
-			// if the token exists already in the DB, then we need to merge the token with the existing token
-			// in the case that it washigher priority above, the "curSeenToken" will be the existing token so it will be merging against itself, essentially a no-op.
-			// if the token was not a higher priority token, then we will still use it to merge to ensure that even lower priority tokens have their fields considered
-			// when a higher priority token may have bad or empty data
-			if existsAlready {
-				curSeenToken := seenTokens[ti]
-				seenTokens[ti] = mergeToken(curSeenToken, existingToken)
+			// Last write wins
```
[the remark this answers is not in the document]
jarrel-b: This is an interesting one! `mergeToken` used the metadata and the fallback media to decide which token to keep, but the values will be the same for every instance of that token now (they'll both point to the same token definition). The remaining fields (owner, block number, contract) will always be the same values across instances.

PR #1240, opened 2023-10-31
Add missing line to contract_creators view

radazen:
This PR fixes an issue that triggers errors like this one:
```
contract creator for community ID=2SvX1eBNRAHpLexv7r9DuM5QwOl is invalid: must have either an address or a user ID
```

The `contract_creators` view was updated in a recent PR, but the update was missing the final line from the original view, which filtered out empty results (contracts that have no creator address or owner ID). This PR adds that line back in.


jarrel-b (approved):
nice catch!

PR #1241, opened 2023-11-01
Make go vet happy

radazen:
Generated dataloader code was doing this:
```
dataloader := generator.NewDataloader(ctx, maxBatchSize, batchTimeout, cacheResults, publishResults, fetchWithHooks)
d.Dataloader = *dataloader
```

That counts as copying a mutex (because there's a mutex in the underlying sync.Map), though AFAIK it's okay to copy a zero-value mutex prior to locking it. All the same, `go vet` flags this as a problem, so now the generated code does this instead:

```
d.Dataloader = *generator.NewDataloader(ctx, maxBatchSize, batchTimeout, cacheResults, publishResults, fetchWithHooks)
```

and that makes `go vet` happy.

### DIFF
--- a/cmd/dataloaders/generator/template.go
+++ b/cmd/dataloaders/generator/template.go
@@ -72,12 +72,10 @@ func new{{.Name}}(
 	}
 
 	{{ if .KeyIsComparable }}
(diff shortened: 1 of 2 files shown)


jarrel-b (approved):
woo thanks for looking into it!

PR #1242, opened 2023-11-01
zora prefixed token ID, notif type, creator token l1 lookup

benny-conn:
Changes:

- **Added** new notification type for the posted your work notification
- **Changed** zora token ID parsing to always split on `-` and use the number following the `-`
- **Changed** `is_creator_token` determination to use L1 chain

### DIFF
--- a/service/multichain/zora/zora.go
+++ b/service/multichain/zora/zora.go
@@ -29,16 +29,8 @@ type Provider struct {
 type tokenID string
 
 func (t tokenID) toBase16String() string {
-	copy := string(t)
-	if strings.Contains("-", copy) {
-		// take what is after the dash
-		parts := strings.Split(copy, "-")
-		if len(parts) != 2 {
-			panic(fmt.Sprintf("invalid token id %s", t))
-		}
-		copy = parts[1]
-	}
-	big, ok := new(big.Int).SetString(copy, 10)
+	spl := strings.Split(string(t), "-")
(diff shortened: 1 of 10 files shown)


service/multichain/zora/zora.go
```diff
@@ -29,16 +29,8 @@ type Provider struct {
...
-		// take what is after the dash
-		parts := strings.Split(copy, "-")
-		if len(parts) != 2 {
-			panic(fmt.Sprintf("invalid token id %s", t))
-		}
-		copy = parts[1]
-	}
-	big, ok := new(big.Int).SetString(copy, 10)
+	spl := strings.Split(string(t), "-")
+	big, ok := new(big.Int).SetString(spl[len(spl)-1], 10)
```
jarrel-b, line 33: Do you know why Zora has `premint-` for some of their tokens? Wondering if tokenID `premint-1` is a separate token to tokenID `1` and if might be overwriting the owner of tokenID `1` with the owner of `premint-1`?

I know we standardize on hex, but if this is the actual token ID I wonder if we should to keep the ID as is 🤔

jarrel-b (approved):
Had a question on what `premint` means, and looks like tests are failing with:
```
panic: NodeFetcher handler validation failed: no handler set for NodeFetcher.OnSomeonePostedYourWorkNotification
```
but otherwise lgtm!

service/multichain/zora/zora.go
benny-conn: I will look into this... very good question.

PR #1264, opened 2023-11-07
notification handling for posting first post [GAL-4716]

benny-conn:
Changes:

- **Added** a new action `UserPostedFirstPost`
- **Added** a new type of event notification handler for "follower" events that don't have a single owner and are going to be associated with the followers of the event actor
- **Added** a new type of notification handler for "follower" notifications that will fan out notifications to the followers of the determined actor based on the notification type (e.g. a user posted first notification will fan out to the followers of the poster user)
- **Added** new GQL type `SomeoneYouFollowPostedTheirFirstPostNotification` to the schema

How it works:

1. A user posts, we check if this is their first post (posts.count == 1). If it is, we send a single event saying this user posted their first work
2. A new event handler picks up that event and creates a notification that says, "here is all the data needed for this notification, but we don't have an owner just yet because this notification is meant to be sent out to many owners, the followers of the event actor"
3. A new notification handler picks up that notification and uses it to insert batch many notifications after determining who the "actor" of the notification is. In the case of the first post, this will be the user who posted.
4. The new notifications are sent as they are usually to the push notification task queue

### DIFF
--- a/db/queries/core/query.sql
+++ b/db/queries/core/query.sql
@@ -820,6 +820,10 @@ INSERT INTO notifications (id, owner_id, action, data, event_ids, feed_event_id,
 -- name: CreateUserPostedYourWorkNotification :one
 INSERT INTO notifications (id, owner_id, action, data, event_ids, post_id, contract_id) VALUES ($1, $2, $3, $4, $5, sqlc.narg('post'), $6) RETURNING *;
 
+-- name: CreateUserPostedFirstPostNotifications :many
+INSERT INTO notifications (id, owner_id, action, data, event_ids, post_id) select $1, follows.follower, $2, $3, $4, sqlc.narg('post') from follows where follows.followee = @actor_id RETURNING *;
+
+
 -- name: CreateSimpleNotification :one
 INSERT INTO notifications (id, owner_id, action, data, event_ids) VALUES ($1, $2, $3, $4, $5) RETURNING *;
 
--- a/service/notifications/notifications.go
+++ b/service/notifications/notifications.go
@@ -12,6 +12,7 @@ import (
 	"github.com/mikeydub/go-gallery/service/redis"
 	sentryutil "github.com/mikeydub/go-gallery/service/sentry"
 	"github.com/mikeydub/go-gallery/service/task"
+	"github.com/sourcegraph/conc/pool"
 
 	cloudtasks "cloud.google.com/go/cloudtasks/apiv2"
 	"cloud.google.com/go/pubsub"
@@ -163,6 +164,7 @@ func New(queries *db.Queries, pub *pubsub.Client, taskClient *cloudtasks.Client,
 
 	def := defaultNotificationHandler{queries: queries, pubSub: pub, taskClient: taskClient, limiter: limiter}
 	group := groupedNotificationHandler{queries: queries, pubSub: pub, taskClient: taskClient, limiter: limiter}
+	follower := followerNotificationHandler{queries: queries, pubSub: pub, taskClient: taskClient, limiter: limiter}
 	tokenIDGrouped := tokenIDGroupedNotificationHandler{queries: queries, pubSub: pub, taskClient: taskClient, limiter: limiter}
 	view := viewedNotificationHandler{queries: queries, pubSub: pub, taskClient: taskClient, limiter: limiter}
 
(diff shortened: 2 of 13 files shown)


db/queries/core/query.sql
```diff
@@ -820,6 +820,10 @@ INSERT INTO notifications (id, owner_id, action, data, event_ids, feed_event_id,
 -- name: CreateUserPostedYourWorkNotification :one
 INSERT INTO notifications (id, owner_id, action, data, event_ids, post_id, contract_id) VALUES ($1, $2, $3, $4, $5, sqlc.narg('post'), $6) RETURNING *;
 
+-- name: CreateUserPostedFirstPostNotifications :many
+INSERT INTO notifications (id, owner_id, action, data, event_ids, post_id) select $1, follows.follower, $2, $3, $4, sqlc.narg('post') from follows where follows.followee = @actor_id RETURNING *;
```
benny-conn, line 824: I feel like I might've tripped up on followee vs follower but this seems correct to me, the followee is the person being followed (the poster) and the followers are the users following the person being posted.

service/notifications/notifications.go
```diff
@@ -631,7 +647,7 @@ func createPushMessage(ctx context.Context, notif db.Notification, queries *db.Q
 			return task.PushNotificationMessage{}, err
 		}
 
-		if err = limiter.tryMention(ctx, commenter.ID, notif.OwnerID, notif.FeedEventID); err != nil {
+		if err = limiter.tryMention(ctx, commenter.ID, notif.OwnerID, notif.PostID); err != nil {
```
radazen, line 650: Similar to another comment I left elsewhere: are we still firing notifications for things like comments on feed events? If so, I don't think we can just change this (and the ones for mentions) to PostID; I think it'd need to use the FeedEventID for comments on feed events, and the PostID for comments on posts.

service/notifications/notifications.go
```diff
@@ -691,19 +707,10 @@ func createPushMessage(ctx context.Context, notif db.Notification, queries *db.Q
...
 
-		if !actor.Username.Valid {
-			return task.PushNotificationMessage{}, fmt.Errorf("user with ID=%s has no username", actor.ID)
-		}
-		contract, err := queries.GetContractByID(ctx, notif.ContractID)
-		if err != nil {
-			return task.PushNotificationMessage{}, err
-		}
-		message.Body = fmt.Sprintf("%s posted your work: %s", actor.Username.String, contract.Name.String)
-		return message, nil
```
radazen, line 706: Was this `return` removed intentionally? It looks like we're no longer returning anything here, and will fall through to the `unsupported notification action` block below.

service/notifications/notifications.go
```diff
@@ -1025,9 +1075,39 @@ func insertAndPublishNotif(ctx context.Context, notif db.Notification, queries *
...
+
+	p := pool.New().WithErrors().WithContext(ctx).WithMaxGoroutines(10)
+	for _, notif := range notifs {
+		notif := notif
+		p.Go(func(ctx context.Context) error {
+			return sendAndPublishPushNotifications(ctx, notif, queries, taskClient, limiter, ps)
+		})
+	}
+
+	logger.For(ctx).Infof("pushed new notification to pubsub: %s", notif.OwnerID)
```
radazen, line 1102: I'm not sure if I understand this log message. We actually pushed a bunch of notifications to pubsub, and none of them were sent to `notif.OwnerID`, right? And since `sendAndPublishPushNotifications` sends push notifications too, it's more than just pubsub!

Maybe "pushed new notifications to followers of userID %s"?

service/notifications/notifications.go
```diff
@@ -1025,9 +1075,39 @@ func insertAndPublishNotif(ctx context.Context, notif db.Notification, queries *
...
+
+	return nil
+}
+
+func sendAndPublishPushNotifications(ctx context.Context, newNotif db.Notification, queries *db.Queries, taskClient *cloudtasks.Client, limiter *pushLimiter, ps *pubsub.Client) error {
```
radazen, line 1107: For clarity, maybe this would be `sendNotifications`, and it'd call `sendPushNotifications` and `sendPubsubNotifications`? I was originally trying to retain the `publish` terminology to mean "sent pubsub notifications," but at this point maybe we just call them pubsub notifications.

radazen (approved):
🎉🏆🎉  Looks good to me! I left some minor comments!

PR #1267, opened 2023-11-08
Add return to ActionUserPostedYourWork

jarrel-b:
Noticed when I mentioned Figure31 that this sentry error popped up: https://usegallery.sentry.io/issues/4612178636/?referrer=slack&notification_uuid=68366714-4e34-4a68-80df-c9d14a7e1041&alert_rule_id=13000544&alert_type=issue

The notification was sent, but it fell through and hit the error because it was missing a `return`.


radazen (approved):
🥇

PR #1268, opened 2023-11-09
Tokenprocessing pipeline updates

jarrel-b:
### What's new?
* Added a new job option `requireImage` which requires an image to be processed if an image keyword is in the metadata. `requireImage` is enabled for jobs started by the `processPostPreflight` endpoint to help ensure that preview images are available for posts.
* Refactored how the Opensea fallback is used. Before, when processing an animation, if the pipeline failed to download content from the animation URL, OS was used to download any asset type (even a non-animation asset). The same was true for images. Now, we try to fetch specific assets from OS. The fallback is now only used at the top-level of the pipeline and only if the pipeline requirements aren't met yet (e.g. if the job has `requireImage` enabled and an image wasn't cached yet)
* Non `2xx` HTTP codes are now categorized as `errInvalidMedia` instead of `errNoDataFromReader`. `errNoDataFromReader` errors are now wrapped as `ErrBadToken` so those errors are excluded from error reporting and altering.
* Fixes bug where higher priority object types weren't used in the final media result

### DIFF
--- a/service/persist/token_processing_jobs.go
+++ b/service/persist/token_processing_jobs.go
@@ -98,52 +98,58 @@ func (p PipelineStepStatus) MarshalJSON() ([]byte, error) {
 }
 
 type PipelineMetadata struct {
-	MetadataRetrieval                        PipelineStepStatus `json:"metadata_retrieval,omitempty"`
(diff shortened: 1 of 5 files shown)


benny-conn (approved):
Looks great! Very thorough and love all of the naming adjustments, made it very easy to understand!

PR #1276, opened 2023-11-14
show nested comments as a flat list

benny-conn:
Changes:

- **Added** a new column to the `comments` table called `reply_ancestors` that is an array of all of the reply ancestors (including the most recently replied to comment in `reply_to`) so that when we ask for the replies to a comment we can return any comment where the ancestors contains this comment we are querying for. 

CleanShot 2023-11-13 at 17 47 30@2x (1)

I experimented with a few paths before arriving here:

- My first instinct was to just have the get replies query traverse the reply tree until it reaches the comment we are querying for. SQLC hated this approach and no matter how much experimentation I did with the query and chatGPT and copilot I could not get a recursive reply pagination query to work so I went looking for other answers.
- My second instinct was to just have a new column that stores the "top level comment" ID. This mostly works for the use case we are looking for given that we only really care about showing a flat list of comments under the top level comment (no matter who is replying to who and how deeply nested, it will show flat ordered by comment time). The problem with this is it kind of separates top level comments into their own class of comments and "replies" would have to function differently for them than any nested reply. If I ask for `top_level_comment.replies` we simply get all comments where the `top_level_comment_id = top_level_comment.id`. What about if I want the replies to a reply in the future? Well asking for any comments where `top_level_comment_id = nested_reply.id` will return nothing and if we ask for  `top_level_comment_id = nested_reply.top_level_comment_id` it will return comments that aren't replying to this comment. 
- So then I'm thinking, well do we even need nested replies if we are going to always just show a flat list? Why don't we just drop the idea of a `reply_to` and just have there be a top level comment ID stored or just not allow a reply to a reply? Well despite the comments being ordered by date, there has to be some way for a user to see visually that a comment further in the chain is replying to them. Some way or another we need to store what specific comment is being replied to, or at least who is being replied to for the user to then infer which comment.

That landed me with two final directions:

1. The frontend stops replying to nested replies and instead does what instagram does and whenever a nested comment is being replied to it just mentions the user being replied to. We could then match this on the backend by adding a constraint to ensure no replies are replied to and they all are replying to the top level comment only. e.g. TL comment: I love this post, reply to TL: agreed, reply to reply: @replier I disagree
2. We store an ancestry of replies on the comment so that we can lookup the comments that contain the queried comment in their ancestry. This works for any level of nesting and will always return only the comments that are below in the reply tree.

Although I feel that solution 1 is better, solution 2 leaves us with the option to show nested comments reddit style later on and also means the frontend can have the "reply" button always function the same for top level comments and nested comments, they don't need to worry about "do we mention here or do we reply here?". What do we think?

### DIFF
--- a/db/queries/core/query.sql
+++ b/db/queries/core/query.sql
@@ -702,16 +702,16 @@ SELECT * FROM admires WHERE token_id = sqlc.arg('token_id') AND (not @only_for_a
 SELECT count(*) FROM admires WHERE token_id = $1 AND deleted = false;
 
 -- name: GetCommentByCommentID :one
-SELECT * FROM comments WHERE id = $1 and deleted = false;
+SELECT * FROM comments WHERE id = $1;
 
 -- name: GetCommentsByCommentIDs :many
-SELECT * from comments WHERE id = ANY(@comment_ids) and deleted = false;
+SELECT * from comments WHERE id = ANY(@comment_ids);
 
 -- name: GetCommentByCommentIDBatch :batchone
-SELECT * FROM comments WHERE id = $1 and deleted = false;
+SELECT * FROM comments WHERE id = $1;
 
 -- name: PaginateCommentsByFeedEventIDBatch :batchmany
-SELECT * FROM comments WHERE feed_event_id = sqlc.arg('feed_event_id') AND reply_to is null AND deleted = false
+SELECT * FROM comments WHERE feed_event_id = sqlc.arg('feed_event_id') AND reply_to is null
     AND (created_at, id) < (sqlc.arg('cur_before_time'), sqlc.arg('cur_before_id'))
     AND (created_at, id) > (sqlc.arg('cur_after_time'), sqlc.arg('cur_after_id'))
     ORDER BY CASE WHEN sqlc.arg('paging_forward')::bool THEN (created_at, id) END ASC,
@@ -719,10 +719,10 @@ SELECT * FROM comments WHERE feed_event_id = sqlc.arg('feed_event_id') AND reply
     LIMIT sqlc.arg('limit');
 
 -- name: CountCommentsByFeedEventIDBatch :batchone
-SELECT count(*) FROM comments WHERE feed_event_id = sqlc.arg('feed_event_id') AND reply_to is null AND deleted = false;
+SELECT count(*) FROM comments WHERE feed_event_id = sqlc.arg('feed_event_id') AND reply_to is null;
 
 -- name: PaginateCommentsByPostIDBatch :batchmany
-SELECT * FROM comments WHERE post_id = sqlc.arg('post_id') AND reply_to is null AND deleted = false
+SELECT * FROM comments WHERE post_id = sqlc.arg('post_id') AND reply_to is null
     AND (created_at, id) < (sqlc.arg('cur_before_time'), sqlc.arg('cur_before_id'))
     AND (created_at, id) > (sqlc.arg('cur_after_time'), sqlc.arg('cur_after_id'))
     ORDER BY CASE WHEN sqlc.arg('paging_forward')::bool THEN (created_at, id) END ASC,
@@ -730,7 +730,7 @@ SELECT * FROM comments WHERE post_id = sqlc.arg('post_id') AND reply_to is null 
     LIMIT sqlc.arg('limit');
 
 -- name: PaginateRepliesByCommentIDBatch :batchmany
-SELECT * FROM comments WHERE reply_to = sqlc.arg('comment_id') AND deleted = false
+SELECT * FROM comments WHERE sqlc.arg('comment_id') = any(reply_ancestors)
     AND (created_at, id) < (sqlc.arg('cur_before_time'), sqlc.arg('cur_before_id'))
     AND (created_at, id) > (sqlc.arg('cur_after_time'), sqlc.arg('cur_after_id'))
     ORDER BY CASE WHEN sqlc.arg('paging_forward')::bool THEN (created_at, id) END ASC,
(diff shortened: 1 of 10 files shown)


db/queries/core/query.sql
```diff
@@ -702,46 +702,64 @@ SELECT * FROM admires WHERE token_id = sqlc.arg('token_id') AND (not @only_for_a
...
     ORDER BY CASE WHEN sqlc.arg('paging_forward')::bool THEN (created_at, id) END ASC,
              CASE WHEN NOT sqlc.arg('paging_forward')::bool THEN (created_at, id) END DESC
     LIMIT sqlc.arg('limit');
 
 -- name: PaginateRepliesByCommentIDBatch :batchmany
-SELECT * FROM comments WHERE reply_to = sqlc.arg('comment_id') AND deleted = false
-    AND (created_at, id) < (sqlc.arg('cur_before_time'), sqlc.arg('cur_before_id'))
-    AND (created_at, id) > (sqlc.arg('cur_after_time'), sqlc.arg('cur_after_id'))
+SELECT * FROM comments WHERE 
+    (
```
radazen, line 734: Are subqueries necessary here? It looks to me like this would give the same result without requiring additional querying:

```
select * from comments where
    (
        (reply_to is null and top_level_comment_id = sqlc.arg('comment_id')) 
        or 
        (reply_to is not null and reply_to = sqlc.arg('comment_id'))
    )
    and (comments.created_at, comments.id) < (sqlc.arg('cur_before_time'), sqlc.arg('cur_before_id'))
    and (comments.created_at, comments.id) > (sqlc.arg('cur_after_time'), sqlc.arg('cur_after_id'))
```
 ---

Same for `CountRepliesByCommentIDBatch` below

service/persist/postgres/comment.go
```diff
@@ -33,12 +35,25 @@ func NewCommentRepository(db *sql.DB, queries *db.Queries) *CommentRepository {
 	deleteStmt, err := db.PrepareContext(ctx, `UPDATE comments SET REMOVED = TRUE, COMMENT = 'comment removed' WHERE ID = $1;`)
 	checkNoErr(err)
 
+	topAncestorStmt, err := db.PrepareContext(ctx, `WITH RECURSIVE comment_thread AS (
```
radazen, line 38: Do we need this to be recursive? Once the migration is run and all comments have their `top_level_comment_id` filled in, wouldn't the `top_level_comment_id` for any reply be either:

1.  the `top_level_comment_id` of the comment being replied to, or
2. the `id` of the comment being replied to, if `top_level_comment_id` is null?

If the above is true, inserting could probably be a single query with a CTE, like:

```
with top_id as (
    select coalesce(top_level_comment_id, id) from comments where id = sqlc.arg('reply_to') and not deleted
)
```

Also, aside from legacy reasons (e.g. continuing to use a transaction started by raw SQL statements), if the `RECURSIVE` is gone, is there a reason this needs to be raw SQL instead of using sqlc?

db/queries/core/query.sql
```diff
@@ -702,46 +702,64 @@ SELECT * FROM admires WHERE token_id = sqlc.arg('token_id') AND (not @only_for_a
 SELECT count(*) FROM admires WHERE token_id = $1 AND deleted = false;
 
 -- name: GetCommentByCommentID :one
-SELECT * FROM comments WHERE id = $1 and deleted = false;
```
radazen, line 705: IMHO, we should think of soft deletes as "hard deletes with an audit trail." I.e., I think we should really avoid using any row that has `deleted = true`. If we need to "delete" comments, could we add a separate column for that kind of deletion? I'd call it `user_deleted` or something, except I'm sure we might also delete the comments via admin portal sometimes, so it probably needs a better name. `removed`? `content_deleted`?

I like to think of it like this: if we needed to hard delete rows with `deleted = true` -- either to reclaim space, or for performance, or regulatory compliance, or something else -- would that have a functional impact on the backend? Ideally, the answer should be no!

radazen (commented):
🚀 Looks good, Benny! I left some comments!

db/queries/core/query.sql
benny-conn: Agreed, blame chatGPT

service/persist/postgres/comment.go
benny-conn: True! Yes I left it in raw SQL for the sake of the transaction but I will get rid of this

db/queries/core/query.sql
```diff
@@ -1714,3 +1702,24 @@ select * from mentions where id = @id and not deleted;
...
 select u.id, w.address, u.pii_socials->>'Lens' is null, u.pii_socials->>'Farcaster' is null from pii.user_view u join wallets w on w.id = any(u.wallets) where u.deleted = false and w.chain = 0 and w.deleted = false and u.universal = false and (u.pii_socials->>'Lens' is null or u.pii_socials->>'Farcaster' is null) order by u.created_at desc;
+
+-- name: InsertMention :one
+INSERT INTO mentions (ID, COMMENT_ID, USER_ID, CONTRACT_ID, START, LENGTH) VALUES ($1, $2, sqlc.narg('user'), sqlc.narg('contract'), $3, $4) RETURNING ID;
+
+-- name: InsertComment :one
+INSERT INTO comments 
+(ID, FEED_EVENT_ID, POST_ID, ACTOR_ID, REPLY_TO, TOP_LEVEL_COMMENT_ID, COMMENT) 
+VALUES 
+(sqlc.arg('id'), sqlc.narg('feed_event')::varchar, sqlc.narg('post')::varchar, sqlc.arg('actor_id'), sqlc.narg('reply')::varchar,
```
radazen, line 1713: For consistency, could these be `feed_event_id`, `post_id`, and `reply_to`?

radazen (approved):
Updates look great! Thanks for the diligence! 🏆

db/queries/core/query.sql
benny-conn: I always run into a problem when I do that where it forces the type to be a DBID which cannot be nullable. This forces the type to be sql.NullString. Would be great to maybe use pointers at some point or have a nullable DBID type?

PR #1278, opened 2023-11-14
Add notification tests

radazen:
This PR adds tests for notifications and push notifications for comments/admires on posts/feed events. We should probably add some of the newer notification types (mentions, first post, etc) in a future follow-up.

The biggest change is refactoring our `task` package to pass around our own `task.Client` implementation, instead of passing around a GCP task client. By encapsulating task creation in our own struct, we can change how tasks are dispatched (in this case, tests skip the queue emulator and invoke task targets directly).


service/task/cloudtask.go
```diff
@@ -122,125 +129,163 @@ type PushNotificationMessage struct {
...
 
 // NewClient returns a new task client with tracing enabled.
-func NewClient(ctx context.Context) *gcptasks.Client {
+func NewClient(ctx context.Context) *Client {
+	skipQueues := make(map[string]bool)
+	for _, q := range env.GetStringSlice("CLOUD_TASKS_SKIP_QUEUES") {
+		skipQueues[q] = true
+	}
+
+	if env.GetBool("CLOUD_TASKS_DIRECT_DISPATCH_ENABLED") {
```
radazen, line 256: I'd rather use a constructor option than an environment variable here, but that'd necessitate more refactoring, such that everywhere we instantiate a task client, we could also pass in some parameters for testing...and that felt pretty clunky.

service/task/cloudtask.go
```diff
@@ -122,125 +129,163 @@ type PushNotificationMessage struct {
...
+		return &Client{skipQueues: skipQueues, sendFunc: useCloudTasks(ctx, newGCPClient(ctx))}
+	}
+}
+
+func useDirectDispatch(ctx context.Context) func(ctx context.Context, queue string, task *taskspb.Task) error {
+	logger.For(ctx).Info("Initializing task client with direct dispatch")
+	httpClient := &http.Client{}
+	return func(ctx context.Context, queue string, task *taskspb.Task) error {
+		go func() {
+			// Ignore the passed-in context. We're likely creating a task in response to a gin request,
```
radazen, line 268: This is annoying, but I'm not sure what the alternative is. If we don't use `context.Background()`, the request might not be dispatched before the whole gin context is cancelled, and then tests start failing.

service/task/cloudtask.go
jarrel-b, line 1: feel free to ignore: maybe `cloudtask.go` isn't the best name anymore since it doesn't strictly use cloudtasks now

service/task/cloudtask.go
jarrel-b: This sounds good to me! If this is only intended for testing, we could also expose a way to construct a client with a provided `sendFunc`, so that `useDirectDispatch` can be moved to tests and changed to a synchronous function.

jarrel-b (approved):
Looks awesome!

service/task/cloudtask.go
radazen: It's definitely only intended for testing! Honestly, the send function could even be synchronous _now_, but I made it async so I could emulate the "with delay" option that some tasks use. But if we didn't support that, we could just make this synchronous.

graphql/graphql_test.go
```diff
@@ -803,6 +804,39 @@ func testViewToken(t *testing.T) {
...
+	c2 := authedHandlerClient(t, alice.ID)
+	registerPushToken(t, ctx, c)
+
+	admirePost(t, ctx, c2, userF.PostIDs[0])
+	commentOnPost(t, ctx, c2, userF.PostIDs[0], "post comment 1")
+	commentOnPost(t, ctx, c2, userF.PostIDs[0], "post comment 2")
+	admireFeedEvent(t, ctx, c2, userF.FeedEventIDs[0])
+	commentOnFeedEvent(t, ctx, c2, userF.FeedEventIDs[0], "feed event comment")
+
+	require.Eventuallyf(t, func() bool {
```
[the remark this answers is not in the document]
radazen: > TIL of `Eventually`

You're only a few days behind me there!

> Does this mean this takes generally 30 seconds to run?

Nope! It checks the condition every second, and will fail if the condition isn't satisfied in 30 seconds. In the common/passing case, it should only take as long as push notifications take to propagate.

PR #1281, opened 2023-11-16
Use retries by default for all Postgres clients

radazen:
After adding retries for Postgres clients created by the main backend server, I saw tests continuing to fail (without retries!) for clients created by other services (e.g. `pushnotifications`). In general, it seems to me that we want to retry connections before giving up, so this PR makes it the default behavior whenever we create a Postgres client, and adds a `WithNoRetries` option to disable the behavior.

### DIFF
--- a/util/retry/retry.go
+++ b/util/retry/retry.go
@@ -101,7 +101,9 @@ func RetryFunc(ctx context.Context, f func(ctx context.Context) error, shouldRet
 			return err
 		}
 
-		r.Sleep(i)
+		if i != r.Tries-1 {
+			r.Sleep(i)
+		}
 	}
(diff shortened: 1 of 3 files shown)


util/retry/retry.go
```diff
@@ -101,7 +101,9 @@ func RetryFunc(ctx context.Context, f func(ctx context.Context) error, shouldRet
 			return err
 		}
 
-		r.Sleep(i)
+		if i != r.Tries-1 {
```
radazen, line 104: Minor tweak: we probably don't need to sleep between the final retry and returning an error.

jarrel-b (approved):

PR #1283, opened 2023-11-16
Benny/activity stats

benny-conn:
Changes:

- **Added** a new badge for top X active user
- **Added** a new service for periodically checking the top X users

Also added a new retool section for updating the formula params: (Feature Flag Retool)

Screenshot 2023-11-16 at 2 10 06 PM


db/queries/core/query.sql
```diff
@@ -1707,6 +1707,68 @@ select * from mentions where id = @id and not deleted;
...
+    WHERE u.deleted = false AND u.universal = false
+)
+SELECT actor_id, username, admires_given, admires_received, comments_made, comments_received, score
+FROM scores
+WHERE actor_id IS NOT NULL AND score > 0
+ORDER BY scores.score DESC
+LIMIT $1;
+
+-- name: UpdateTopActiveUsers :exec
+UPDATE users
```
jarrel-b, line 1764: We may want to add an index for users with the top_activity trait at some point since the cost scales with the number of users. However, this runs infrequently and quickly enough currently, so it's probably for now! Although it might not be a bad idea to add one anyway if we want to know who are the most active users currently.

db/queries/core/query.sql
```diff
@@ -1707,6 +1707,68 @@ select * from mentions where id = @id and not deleted;
...
+        (COALESCE(ag.admire_given, 0) * @admire_given_weight::int) + 
+        (COALESCE(cm.comments_made, 0) * @comments_made_weight::int) + 
+        (COALESCE(cr.comments_received, 0) * @comments_received_weight::int)) AS score,
+        COALESCE(nullif(ag.actor_id,''), nullif(ar.actor_id,''), nullif(cm.actor_id,''), nullif(cr.actor_id,'')) AS actor_id,
+        COALESCE(ag.admire_given, 0) AS admires_given,
+        COALESCE(ar.admire_received, 0) AS admires_received,
+        COALESCE(cm.comments_made, 0) AS comments_made,
+        COALESCE(cr.comments_received, 0) AS comments_received
+        
+    FROM users u
```
jarrel-b, line 1750: Since we're only interested in users who interacted in the last week, it'd be awesome to avoid joining on `users` since most users won't have interactions. We could do full outer joins on the users that do match the criteria:
```sql
with scores as (
    select coalesce(...), coalesce(...)
    from ag
    full outer join ar using(actor_id)
    full outer join cm using(actor_id)
    full outer join cr using(actor_id)
)
```
and then join on `users` at the end to filter users in that set:
```sql
with scores as (...)
select *
from scores
join users on scores.actor_id = users.id
where u.deleted = false and u.universal = false
```
This way should prevent the DB needing to work with the entire users table.

jarrel-b (approved):
Looks awesome, just had a few sql suggestions!

PR #1293, opened 2023-11-21
Add contract mint link

jarrel-b:
#### Changes
Adds the `mintURL` field to the `Contract` type. `mintURL` initially supports Ethereum, Optimism, Base, Zora.
| Chain | Referral |
|-|-|
|Ethereum |mint.fun |
|Optimism |mint.fun |
|Base |mint.fun |
|Zora |zora.co |

### DIFF
--- a/db/gen/coredb/batch.go
+++ b/db/gen/coredb/batch.go
@@ -752,7 +752,7 @@ func (b *GetAdmiresByActorIDBatchBatchResults) Close() error {
 }
 
 const getChildContractsByParentIDBatchPaginate = `-- name: GetChildContractsByParentIDBatchPaginate :batchmany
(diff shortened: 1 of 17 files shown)


benny-conn (approved):
Looks great! 

I don't think it would make that much of a difference but it might be worth it to have the mint link get generated on each multichain provider just like you added for tezos. It only saves us an update to the database on the first call to retrieve a contract after this goes out which isn't really saving much, maybe it just makes more sense to me for a query field resolver not to be responsible with updating the database.

jarrel-b:
> Looks great!
> 
> I don't think it would make that much of a difference but it might be worth it to have the mint link get generated on each multichain provider just like you added for tezos. It only saves us an update to the database on the first call to retrieve a contract after this goes out which isn't really saving much, maybe it just makes more sense to me for a query field resolver not to be responsible with updating the database.

Good call! This is me being too lazy to write a backfill script to populate the links for existing contracts, I'll address this tmrw!

PR #1301, opened 2023-11-28
tez contracts by owner

benny-conn:
Changes:

- **Added** contracts by owner query support for the default tezos provider
- **Added** demo code that shows how to retrieve both fxhash and objkt (hic et nunc) collections by the owner address

### DIFF
--- a/service/multichain/tezos/objkt.go
+++ b/service/multichain/tezos/objkt.go
@@ -37,6 +37,7 @@ type contract struct {
 	Creator_Address persist.Address
 	Level           int
 	Type            tokenStandard
+	Collection_ID   string
 	Logo            string
 }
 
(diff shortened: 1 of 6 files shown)


service/multichain/tezos/objkt.go
```diff
@@ -87,6 +88,60 @@ type tokensByIdentifiersQuery struct {
...
+        name
+      }
+    }
+  }
+
+  collections by creator
+  fa(where: {creator: {address: {_eq: "tz1hghbrDGx33HendcHRAdKUDtdFoFP24uaE"}}}) {
+	... fa fields
+  }
+*/
```
benny-conn, line 131: Here is how we would get objkt/HEN created subcommunities

service/multichain/tezos/tezos.go
```diff
@@ -494,6 +513,187 @@ func (d *Provider) GetContractByAddress(ctx context.Context, addr persist.Addres
...
+            address
+          }
+        }
+      }
+    }
+  }
+}
+
+
+*/
```
benny-conn, line 632: Here is how we would get fxhash created subcommunities

jarrel-b (approved):
lgtm!

PR #1303, opened 2023-11-29
Add first post flag

jarrel-b:
#### Changes
* Adds `isFirstPost` field to `Post` type
* Adds `is_first_post` to `posts` table
* `is_first_post` is populated when a post is added to the DB


db/migrations/core/000129_add_is_first_post.up.sql
jarrel-b, line 1: This migration should only take a few seconds

Robinnnnn (approved):

PR #1307, opened 2023-11-30
digest email

benny-conn:
Changes:

- **Added** a new email send handler that fetches the top communities and posts (and future will also include top first posts and galleries) as well as retool handlers for getting the current values as well as setting overrides
- **Changed** auto send to handle as a request that will be kicked off by cloud scheduler instead of a pubsub, this will basically make it impossible for multiple emails to get sent out (and as an added layer of security we still have a redis lock). This is because the cron job will only send 1 request as opposed to every active email service receiving the published message and whichever one is first to grab the lock can send the email
- **Added** cron jobs to make

Also added a new retool dashboard for setting overrides and examining what is going to go out.


emails/send.go
```diff
@@ -273,6 +268,110 @@ outer:
...
+	UnsubscribeToken string       `json:"unsubscribeToken"`
+}
+
+func sendDigestEmails(queries *coredb.Queries, loaders *dataloader.Loaders, s *sendgrid.Client, r *redis.Cache, stg *storage.Client, fapi *publicapi.FeedAPI) gin.HandlerFunc {
+	return func(ctx *gin.Context) {
+		// mimic backend auth
+		ctx.Set("auth.user_id", persist.DBID(""))
+		ctx.Set("auth.auth_error", nil)
+
+		l, _ := r.Get(ctx, "send-digest-emails")
```
jarrel-b, line 283: What's the purpose of `send-digest-emails`? Is it to guarantee only one handler sends the email? I think a lock would be safer on the off-chance that two handlers both see that `send-digest-emails` isn't set and then set it.

emails/send.go
```diff
@@ -273,6 +268,110 @@ outer:
...
+		if l != nil && len(l) > 0 {
+			logger.For(ctx).Infof("digest emails already being sent")
+			return
+		}
+		r.Set(ctx, "send-digest-emails", []byte("sending"), 1*time.Hour)
```
jarrel-b, line 288: Is this to limit the frequency of sending to at most one per hour, or to prevent other processes from also sending emails while this process is sending emails? Should the key be removed at the end of sending?

emails/digest.go
```diff
@@ -0,0 +1,363 @@
...
+	if collection.ProfileImageUrl.String == "" {
+		tokens, err := q.GetTokensByContractIdPaginate(ctx, coredb.GetTokensByContractIdPaginateParams{
+			ID:               collection.ID,
+			Limit:            1,
+			GalleryUsersOnly: true,
+		})
+		if err == nil && len(tokens) > 0 {
+			media, err := l.GetMediaByMediaIdIgnoringStatusBatch.Load(tokens[0].TokenDefinition.TokenMediaID)
+			if err == nil {
+				collection.ProfileImageUrl.String = util.FirstNonEmptyString(media.Media.ThumbnailURL.String(), media.Media.MediaURL.String())
```
jarrel-b, line 244: Wonder if we want to use imgix URLs vs our bucket so we take advantage of their CDN, etc?

emails/digest.go
```diff
@@ -0,0 +1,363 @@
...
+			}
+		}
+
+	}
+	return UserFacingContract{
+		ContractID:      collection.ID,
+		Name:            collection.Name.String,
```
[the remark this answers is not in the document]
jarrel-b: Relatedly, I'm wondering if this makes sense to fetch data via graphQL vs from the DB if we do use communities because the resolvers will handle the business logic of getting the right preview URLs and handling overrides.

PR #1311, opened 2023-12-04
Add fxhash specific gateway

jarrel-b:
Fxhash has a gateway that caches their content locally, so we can use that to retrieve fxhash-specific assets faster.


benny-conn (approved):
Looks great!

PR #1321, opened 2023-12-12
Add onboarding user recs

jarrel-b:
### Schema changes
Adds a query field `onboardingUserRecommendations` meant to get users to recommend during onboarding. It pulls users who are either hand-picked by us (from the new `user_internal_recommendations` table), who have the active badge, or who are frequently recommended. The order is determined by the number of times a user appears across lists, so users who are in all three lists are highly recommended. `onboardingUserRecommendations` doesn't check if the user already follows them since it's assumed to be used solely for onboarding. The `suggestedUsers` field on the `Viewer` type should be used instead since it checks if you already follow a user and gives more personalized recommendations (see follow-ups below).

Adds a mutation `followAllOnboardingRecommendations` that follows all users that are recommended. An optional `cursor` can be provided as input to follow all users from when the cursor was made.

```graphql
type query {
  onboardingUserRecommendations(
    before: String
    after: String
    first: Int
    last: Int
  ): UsersConnection
}

type mutation {
  followAllOnboardingRecommendations(
    """
    A cursor returned from `onboardingUserRecommendations` can be provided to follow all users from when the
    cursor was generated. Otherwise, all users from the current recommendation set will be followed instead.
    """
    cursor: String
  ): FollowAllOnboardingRecommendationsPayloadOrError @authRequired
}
```

### Paginator changes
Paginators have been converted to use generics! To initialize a paginator, you'll need to declare the type of the node the paginator returns e.g:
```go
// paginator that returns pages of db.Post
var paginator intTimeIDPaginator[db.Post]
```
If you're defining a paginator that uses an underlying `keysetPaginator`, you'll also need to declare the type of the cursor that is returned e.g:
```go
// converts nodes of type Node to intTimeIDCursors for comparison purposes
var paginator keysetPaginator[Node, *intTimeIDCursor]
```
The bulk of this PR is related to updating where paginators are used to declare the type of node the paginator returns and removing unneeded type assertions and places where we previously needed to convert slices of `any`s to concrete types and vice versa.

### Misc changes
* Removed waterfalls of a few resolvers by updating old queries written before `sqlc.embed` to now use `sqlc.embed`
* Updates some queries that take an array of IDs as input to also take `@cur_before_pos` and `@cur_after_pos` params so that the DB can only return rows that match instead of having to return all rows
* Adds `sqlc.embed(medias)` to some queries that return `db.Token`s so that media is already cached by the dataloader when it comes time to resolve media

### Follow-ups
* There aren't retool admin mutations to manage the `user_internal_recommendations`. It will be manually updated for the time being.
* Update `suggestedUsers` on `Viewer` type to include the new recommendation sources


db/queries/core/query.sql
```diff
@@ -179,6 +180,21 @@ where t.token_definition_id = td.id
     and not td.deleted
     and not c.deleted;
 
+-- name: GetTokenByUserTokenIdentifiersIgnoreDisplayableBatch :batchone
```
jarrel-b, line 183: This is used to fetch an ENS token by its token identifiers. `displayable` is ignored so that ENS profile images match the behavior of token profile images, letting you use the PFP even if you transferred the token.

db/queries/core/query.sql
```diff
@@ -369,16 +385,6 @@ select count(distinct users.id) from users, tokens, contracts
     and (not @gallery_users_only::bool or users.universal = false)
     and tokens.deleted = false and users.deleted = false and contracts.deleted = false;
 
--- name: GetTokenOwnerByID :one
```
jarrel-b, line 372: Was able to remove this query and a waterfall by embedding the user in `GetTokensByContractIdPaginate`.

publicapi/feed.go
jarrel-b, line 1: Feel free to skip this file - the majority of these changes are refactoring.

jarrel-b, line 1: Feel free to skip this file and (collection.go, community.go, contract.go, token.go, etc) - the majority of these changes are refactoring related to pagination.

db/queries/core/query.sql
```diff
@@ -159,10 +159,11 @@ join token_definitions td on t.token_definition_id = td.id
 where t.id = $1 and t.displayable and t.deleted = false and td.deleted = false;
 
 -- name: GetTokenByIdIgnoreDisplayableBatch :batchone
-select sqlc.embed(t), sqlc.embed(td)
+select sqlc.embed(t), sqlc.embed(td), sqlc.embed(tm)
```
jarrel-b, line 162: I added `sqlc.embed(tm)` so that the media is already cached for when media gets resolved for a PFP or for display.

publicapi/pagination.go
```diff
@@ -29,9 +29,9 @@ var (
 	defaultCursorAfterKey = ""
 
 	// Some position that comes after any other position
-	defaultCursorBeforePosition = -1
+	defaultCursorBeforePosition = math.MaxInt32 - 1
```
jarrel-b, line 32: I had this reversed but it wasn't used in queries until now!

PR #1324, opened 2023-12-14
webhook for optimistic syncing zora

benny-conn:
Changes;

- **Added** a webhook handler on the token processing service (same as alchemy webhook handler) that takes in an update to the `token_1155_holder` entity representing an address that holds a zora token on the zora chain and syncs that specific token, also kicking off an event/notification for the new token


tokenprocessing/process.go
```diff
@@ -518,6 +520,129 @@ func isValidAlchemySignatureForStringBody(
...
+			return
+		}
+
+		if len(newTokens) == 0 {
+			l.Info("no new tokens found")
+		}
+
+		for _, token := range newTokens {
+
+			dbToken, err := queries.GetUniqueTokenIdentifiersByTokenID(c, token.Instance.ID)
```
jarrel-b, line 610: It looks like we're interested in only the quantity, `newTokens` is a list of a type with a reference to the user token if that is useful!

tokenprocessing/process.go
```diff
@@ -518,6 +520,129 @@ func isValidAlchemySignatureForStringBody(
...
+				UserID:         user.ID,
+				TokenID:        token.Instance.ID,
+				Action:         persist.ActionNewTokensReceived,
+				Data: persist.EventData{
+					NewTokenID:       token.Instance.ID,
+					NewTokenQuantity: persist.HexString(newBalance.Text(16)),
+				},
+			})
+			if err != nil {
+				l.Errorf("error dispatching event: %s", err)
```
jarrel-b, line 638: maybe report to sentry here too?

jarrel-b (approved):

service/multichain/zora/zora.go
```diff
@@ -180,8 +180,31 @@ func (d *Provider) GetTokensIncrementallyByWalletAddress(ctx context.Context, ad
 }
 
 func (d *Provider) GetTokenByTokenIdentifiersAndOwner(ctx context.Context, ti multichain.ChainAgnosticIdentifiers, owner persist.Address) (multichain.ChainAgnosticToken, multichain.ChainAgnosticContract, error) {
-	url := fmt.Sprintf("%s/contract/ZORA-MAINNET/%s?token_id=%s", zoraRESTURL, ti.ContractAddress.String(), ti.TokenID.Base10String())
-	return d.getToken(ctx, owner, url)
+	tokens, contracts, err := d.GetTokensByWalletAddress(ctx, owner)
```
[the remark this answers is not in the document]
jarrel-b: hmm now that I think about it my suggestion could be bad for massive ERC-1155 mints

PR #1327, opened 2023-12-15
Use actual communities in the sharedCommunities resolver

radazen:
# What's new?

This PR creates the `owned_communities` materialized view, which is almost identical to the existing `owned_contracts` materialized view, except that also includes communities.

The "get shared communities" resolver is also updated to use this view instead of the contracts view, so users can see which subcommunities they have in common with others.


jarrel-b (approved):
looks great!

Robinnnnn (approved):
🔥

PR #1329, opened 2023-12-20
Change holder syncs to append only

jarrel-b:
**Changes**
* Change holder syncs to append only instead of replacing holder tokens for now
* Fix race condition in `receiveSyncedTokensForUser` where the error returned from a provider may not get checked
* Minor changes to the Alchemy provider: fix bug where the error from marshalling wasn't returned and also remove extra print statement
* Remove `logFallbackFailure` - this could create an uncapped number of goroutines, so I removed it for now
* Add a test that checks that tokens aren't deleted from a failing provider
* Remove multiErr from syncs and instead returns the latest error so that it can get wrapped properly to the correct graphQL error type

### DIFF
--- a/graphql/graphql_test.go
+++ b/graphql/graphql_test.go
@@ -95,7 +95,7 @@ func testTokenSyncs(t *testing.T) {
 		{title: "should sync new tokens multichain", run: testSyncNewTokensMultichain},
 		{title: "should submit new tokens to tokenprocessing", run: testSyncOnlySubmitsNewTokens},
 		{title: "should not submit old tokens to tokenprocessing", run: testSyncSkipsSubmittingOldTokens},
-		{title: "should delete old tokens", run: testSyncDeletesOldTokens},
+		{title: "should not delete old tokens", run: testSyncNotDeletesOldTokens},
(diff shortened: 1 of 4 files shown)


graphql/graphql_test.go
```diff
@@ -874,6 +874,23 @@ func testSyncNewTokens(t *testing.T) {
...
+		require.Greater(t, len(userF.TokenIDs), 0)
+		h := handlerWithProviders(t, submitUserTokensNoop, newStubProvider(withReturnError(fmt.Errorf("can't get tokens"))))
+		c := customHandlerClient(t, h, withJWTOpt(t, userF.ID))
+
+		// Sync tokens with a failing provider
+		response, err := syncTokensMutation(ctx, c, []Chain{ChainEthereum}, nil)
+
+		// Expect the sync to fail because of bad provider
+		require.NoError(t, err)
+		_ = (*response.SyncTokens).(*syncTokensMutationSyncTokensErrSyncFailed)
```
Robinnnnn, line 888: 🤯

radazen (approved):
👍👍 Very happy to have this resolved! 🚀

PR #1343, opened 2024-01-03
Benny/opensea streaming

benny-conn:
Changes:

- **Added** a new singleton service called `opensea-streamer` that listens with a websocket to all token transfers and sends them to the token-processing service
- **Added** a handler on the token processing service for the webhook

Considerations:

- This new service makes A LOT of requests to token processing. 1 for every transfer, most are returned pretty quickly by the token processing service if the user does not exist but it will make a round trip for every incoming transfer to the DB to check. Maybe in the future we batch these? Right now with just zora it should be fine, but opensea supports all the chains we support (minus tezos) so we could maybe use them for those chains as well. I noticed specifically polygon has LOADS of transfers every second so that one might be unwieldy and batchworthy.


opensea-streamer/main.go
```diff
@@ -0,0 +1,362 @@
...
+	ctx = logger.NewContextWithFields(ctx, logrus.Fields{
+		"approximate_bloom_zize": bf.ApproximatedSize(),
+	})
+
+	// Health endpoint
+	router.GET("/health", func(c *gin.Context) {
+		c.String(http.StatusOK, "OK")
+	})
+
+	router.GET("/updateBloomFilter", func(c *gin.Context) {
```
radazen, line 72: Is this used by anything, or is it just for debugging?

opensea-streamer/main.go
```diff
@@ -0,0 +1,362 @@
...
+			if err != nil {
+				logger.For(ctx).Error(err)
+				continue
+			}
+
+			if dbWalletCount == curWalletCount {
+				continue
+			}
+
+			bf, err = resetBloomFilter(ctx, queries, bloomCache)
```
radazen, line 107: Updating `bf` here won't change the reference being used by `streamOpenseaTransfers`. It'll change what the outer `bf` variable in `main()` points to, but it's a pointer -- not a pointer to a pointer -- so it isn't going to change what `streamOpenseaTransfers` sees.

Assuming it's quick to update the bloom filter -- and I think that's true, right? -- we probably don't need to use Redis at all. I was thinking we'd want to use it because it has its own bloom filter implementation and we wouldn't need to store everything in memory on a server if we were using Redis. But given that we're actually storing everything in memory here, and given that rebuilding the filter is probably pretty quick, I don't see much benefit to using Redis. At best, it saves us a small amount of time at startup, but only if there hasn't been a change in the number of active wallets we're tracking.

Instead, let's get rid of Redis entirely. We can use an `atomic.Pointer[bloomFilter]` to track the active bloom filter, and when we update the bloom filter, use the atomic pointer's `Store` function to point to the newly created filter. In your processing loop, you'd just want to load the current reference every time you're going to use the bloom filter. There's some overhead, but using `atomic` is generally much more efficient than using locks, and I think for our purposes it'll be okay. Something like:

```
// check if the wallet is in the bloom filter
chainAddress, err := persist.NewL1ChainAddress(persist.Address(win.Payload.ToAccount.Address.String()), win.Payload.Item.NFTID.Chain).MarshalJSON()
if err != nil {
	logger.For(ctx).Error(err)
	continue
}
bf := myAtomicPointer.Load()
if !bf.Test(chainAddress) {
	continue
}
```

I think you'll also want to put the atomic.Pointer in the global scope to make it easier to ensure that all methods are using the same reference to the same atomic pointer.

db/queries/core/query.sql
```diff
@@ -1936,3 +1936,10 @@ select communities.* from communities
         and token_community_memberships.token_definition_id = @token_definition_id
         and not communities.deleted
         and not token_community_memberships.deleted;
+
+
+-- name: CountActiveWallets :one
+select count(w.*) from users u join wallets w on w.id = any(u.wallets) where not u.deleted and not w.deleted and not u.universal;
```
radazen, line 1942: One edge case here: if one wallet gets added and a different wallet gets deleted, the count stays the same but our active wallets have changed. We could either:
- skip this check for now and just update the filter every hour, or
- use a query that checks whether any active wallets have been added since `<timestamp>`, and check that every hour

I think either is fine; if updating the bloom filter is fast, updating it hourly sounds okay to me.

radazen (commented):
Looking good! The email digest stuff looks A-OK to me. For OpenSea streaming, I left a few comments about how we store and update the bloom filter, but everything else looks great!

emails/send.go
```diff
@@ -357,7 +359,7 @@ func sendDigestEmailsToAllUsers(c context.Context, v DigestValues, queries *core
 	})
 }
 
-func sendDigestEmailToUser(c context.Context, u coredb.PiiUserView, emailRecipient persist.Email, digestValues DigestValues, s *sendgrid.Client, sendRealEmail bool) (*rest.Response, error) {
```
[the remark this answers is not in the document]
benny-conn: I realized it was unnecessary because on anything but prod we only send emails to Admin verified email accounts.

opensea-streamer/main.go
```diff
@@ -0,0 +1,362 @@
...
+		for {
+			_, message, err := conn.ReadMessage()
+			if err != nil {
+				logger.For(ctx).Error(err)
+				break
+			}
+
+			var oe openseaEvent
+			err = json.Unmarshal(message, &oe)
+			if err != nil {
```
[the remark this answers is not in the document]
benny-conn: This one is an interesting one because with the JS SDK we are able to filter down to specific events but because we are doing this in go we get all incoming events from every type of event. Because of this, most of them error out here simply because they are the wrong event type and don't match the JSON schema we defined.

opensea-streamer/main.go
benny-conn: Just used for debugging for now but I imagine can be used when a user signs up if we want to add that functionality later?

PR #1387, opened 2024-02-06
custom metadata eth provider

benny-conn:
Changes:

- **Added** a new `CustomMetadataFetcher` interface for determining if there are any special handlings of a certain token. Attached this to the `EthereumProvider` and made the logic changes to the fetch metadata multichain function

At first I tried to migrate the logic from the indexer custom metadata handling to the multichain package but it caused a lot of cyclical imports and required a lot of extra imports for eth clients and ipfs clients and such. Since the indexer already has this dandy metadata handling implemented, the new fetcher interface just uses the indexer provider we already have and calls the indexer to get the metadata for the custom handled contracts like punks and autoglyphs.


tokenprocessing/media.go
```diff
@@ -975,6 +979,9 @@ func cacheObjectsFromURL(pCtx context.Context, tids persist.TokenIdentifiers, me
 			return result, nil
 		}
 		logger.For(pCtx).Infof("cached animation for %s in %s", tids, time.Since(timeBeforeCache))
+		if tids.ContractAddress == eth.PunkAddress {
+			return append(result[1:], obj...), nil
```
jarrel-b, line 983: Wondering why we skip the first result, isn't `result[0]` the original SVG that would get used for the detail page?

tokenprocessing/media.go
```diff
@@ -681,14 +681,18 @@ func rasterizeAndCacheSVGMedia(ctx context.Context, svgURL string, tids persist.
 		return nil, fmt.Errorf("could not decode base64 data: %s", err)
 	}
 
+	pngObjectType := objectTypeThumbnail
+	if tids.ContractAddress == eth.PunkAddress {
+		pngObjectType = objectTypeImage
```
jarrel-b, line 686: What media should be written at the end of the pipeline, is it: an `-svg` object from the original URL  and a `-image` object from the rasterizer that is used as the thumbnail? Or is it a single `-image` object that is used for the thumbnail and the detail page? Wondering if changing this to `objectTypeImage` would overwrite the object written by `cacheRawMedia` (and if that's what we actually want to happen)

jarrel-b (approved):

PR #1395, opened 2024-02-11
farcaster bulk api for autosocial

benny-conn:
Changes:

- **Updated** the autosocial service to use the farcaster bulk v2 API instead of the farcaster verification check in v1.

Because Neynar complained, I am going to pause the current autosocial cron job until we deploy this.


cmd/main.go
jarrel-b, line 1: This looks like a scratch script, dunno if you wanted this checked in or not!

service/farcaster/neynar.go
```diff
@@ -145,6 +148,45 @@ func (n *NeynarAPI) UserByAddress(ctx context.Context, address persist.Address)
...
+
+	if resp.StatusCode != http.StatusOK {
+		bs, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, err
+		}
+		return nil, fmt.Errorf("neynar returned status %d (%s)", resp.StatusCode, bs)
+	}
+
+	defer resp.Body.Close()
```
jarrel-b, line 177: Can we move to after the error check so that response body is always closed?

autosocial/handler.go
```diff
@@ -161,23 +188,36 @@ func addLensProfileToUser(ctx context.Context, l *lens.LensAPI, address []persis
...
+	for address, fusers := range users {
+		if len(fusers) == 0 {
 			continue
 		}
-		u, err := n.UserByAddress(ctx, a.Address())
-		if err != nil {
-			logrus.Error(err)
-			continue
+		// we only store one farcaster profile per user
+		u := fusers[0]
```
jarrel-b, line 201: Do we still want to use `fusers[0]` assuming it might not meet the criteria for `Fid` and `ActiveStatus`?

autosocial/handler.go
```diff
@@ -161,23 +188,36 @@ func addLensProfileToUser(ctx context.Context, l *lens.LensAPI, address []persis
...
+	inner:
```
jarrel-b, line 202: Don't think you need `inner` here!

jarrel-b (approved):

autosocial/handler.go
benny-conn: Will add a check to ensure it is valid after the search for loop, good catch

autosocial/handler.go
benny-conn: Appreciate this, it was a superstition addition but if you think we don't need it we are good lol

PR #1396, opened 2024-02-11
fetch social following for farcaster

benny-conn:
Changes:

- **Added** a case in the social connections to get a user's following farcaster accounts

There was an unused whole database that I removed the only reference to in an old backfill script we did for fetching farcaster following. We can probably drop this database.


publicapi/socials.go
```diff
@@ -60,6 +63,8 @@ func (s SocialAPI) NewLensAuthenticator(userID persist.DBID, address persist.Add
 	}
 }
 
+const farcasterRefetchTime = time.Hour * 24 * 3
```
jarrel-b, line 66: Is this still needed?

jarrel-b (approved):
lgtm!

publicapi/socials.go
benny-conn: Nope! You are right good catch

PR #1397, opened 2024-02-11
use generic name when token name not found

benny-conn:
Changes:

- **Added** a check and retry for empty names of tokens received when creating user facing notification data for it. If all else fails, it will use a generic "token" or "tokens"


jarrel-b (approved):
👍

PR #1407, opened 2024-02-18
add asm to email

benny-conn:
Changes:

- **Added** asm to every email sent with the unsubscribe group ID for that email

To verify that this does in fact add the unsub group to the template as expected, someone needs to attempt to unsubscribe this week :(


radazen (approved):
👍 👍

PR #1408, opened 2024-02-18
get sendgrid unsubscriptions to sync

benny-conn:
Changes:

- **Added** a new get handler for the email service that will query sendgrid for what a user it is currently unsubscribed to. It will also sync what we have stored in the database for that user, using SendGrid as the source of truth. This is better than using us as the source of truth because unsubscribe links in emails do not connect to gallery. If a user clicks that link it will unsubscribe them on sendgrid but not in our database. By syncing with sendgrid first we ensure accuracy.

Note:

- It was in my TODO that there needs to be a gql mutation for updating subscription statuses. That already exists and has for a while: 
```
updateEmailNotificationSettings(
    input: UpdateEmailNotificationSettingsInput!
  ): UpdateEmailNotificationSettingsPayloadOrError @authRequired
  ```

### DIFF
--- a/graphql/schema/schema.graphql
+++ b/graphql/schema/schema.graphql
@@ -864,19 +864,19 @@ enum EmailUnsubscriptionType {
 type UserEmail {
   email: Email
   verificationStatus: EmailVerificationStatus
-  emailNotificationSettings: EmailNotificationSettings
+  emailNotificationSettings: EmailNotificationSettings @goField(forceResolver: true)
 }
 
 type EmailNotificationSettings {
   unsubscribedFromAll: Boolean!
   unsubscribedFromNotifications: Boolean!
-  unsubscribedFromDigest: Boolean # TODO make this non-nullable
+  unsubscribedFromDigest: Boolean!
 }
 
 input UpdateEmailNotificationSettingsInput {
   unsubscribedFromAll: Boolean!
   unsubscribedFromNotifications: Boolean!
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: graphql/schema/schema.graphql [hunk 1]
```diff
@@ -864,19 +864,19 @@ enum EmailUnsubscriptionType {
...
   unsubscribedFromNotifications: Boolean!
-  unsubscribedFromDigest: Boolean # TODO make this non-nullable
+  unsubscribedFromDigest: Boolean!
 }
 
 input UpdateEmailNotificationSettingsInput {
   unsubscribedFromAll: Boolean!
   unsubscribedFromNotifications: Boolean!
-  unsubscribedFromDigest: Boolean # TODO make this non-nullable
+  unsubscribedFromDigest: Boolean!
```
radazen, line 879: @Robinnnnn It looks like this will be a breaking change, though it also looks like we originally made it non-breaking to buy time to update clients before making it breaking. Are web/mobile okay with this field becoming required?

radazen (approved):
Looks good to me!

graphql/schema/schema.graphql
Robinnnnn: this will indeed be a breaking change, and we never ended up updating the frontend.

all good though, just need to make a small change to web!

Robinnnnn (approved):

PR #1409, opened 2024-02-18
skip view gallery notifs

benny-conn:
Changes:

- **Added** a check to skip view gallery notifications in the email for notifications

Considerations:

- All of the other notification types are by default supported now that we have consolidated how we turn notifications into user facing data. We just have had this email type disabled since that change because it also switched from using pubsub to cloud scheduler to prevent duplicate emails being sent. Going to test on dev then re-enable the task on prod


radazen (approved):
👍 👍

Robinnnnn (approved):

PR #1430, opened 2024-03-01
order queries for user notifications by last updated

benny-conn:
Changes:

- **Changed** queries to order by last updated for retrieving user notifications


radazen (changes requested):
I think you also need to update the parameters being passed to these queries. Otherwise, the cursors that use the queries will continue to pass in `CreatedAt` values, and the queries will compare those to `last_updated`.

jarrel-b:
Closing because addressed here: https://github.com/gallery-so/go-gallery/pull/1481

PR #1478, opened 2024-04-09
Also submit existing tokens that don't have media when syncing

jarrel-b:
**Changes**
Before, only new tokens were sent to tokenprocessing. This PR changes the behavior to include existing tokens that lack media. This PR also changes the context passed throughout optimistic syncing to the context with added logging fields.


jarrel-b:
Syncing tests failed but I ran them locally to confirm that tests are still passing.

